Issue #7, opened 2016-12-16
Feature request - Make it easier to add inset dividers to LinearLayout and RecyclerView


tadfisher:
It would be lovely to support inset dividers on LinearLayout and RecyclerView.

For both, fine-grained control would be nice; for example, to inset only the middle dividers in a LinearLayout, or to provide dividers only for specific items in a RecyclerView.

I currently use a couple of modifications to LinearLayout and a custom item decoration to achieve this, which I will submit.

Supporting documentation: https://material.io/guidelines/components/dividers.html#dividers-types-of-dividers

ldjcmu:
We have no plans to implement this.

PR #8, opened 2016-12-17
[Layouts] Add an InsetDividerLinearLayout component

tadfisher:
Ref #7

Adds a standalone modification to LinearLayoutCompat which allows for
control over divider insets. Users may specify "start" and "end" insets,
which affect the appropriate dimension in both horizontal and vertical
orientations, as well as LTR and RTL layout directions. Users may also
specify which dividers to inset, similar to the `showDividers`
attribute.

To make it easier to adhere to the Material Design baseline grid,
dividers are drawn over child views instead of offsetting them. This
also allows for backgrounds to be set directly on children without
leaving gaps where they would normally be offset.

* https://material.io/guidelines/layout/metrics-keylines.html#metrics-keylines-baseline-grids

Additionally, start/end dividers are drawn outside of the layout's
padding by ignoring child dimensions and simply drawing them along the
top/bottom (in vertical orientation) or left/right (in horizontal
orientation) edges of the layout. This makes it easier to apply the
recommended top/bottom padding to the layout per the Lists spec.

* https://material.io/guidelines/components/lists.html#lists-specs

Tests were added to verify some basic behavior as well as covering all
permutations of inset mode, orientation, and layout direction. These
pass on emulators from platform versions 2.3.7 to 7.1.1.

### DIFF
--- a/lib/src/android/support/design/widget/InsetDividerLinearLayout.java
+++ b/lib/src/android/support/design/widget/InsetDividerLinearLayout.java
@@ -1,0 +1,393 @@
+/*
+ * Copyright (C) 2016 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package android.support.design.widget;
+
+import static android.support.annotation.RestrictTo.Scope.LIBRARY_GROUP;
+
+import android.content.Context;
+import android.content.res.TypedArray;
+import android.graphics.Canvas;
+import android.graphics.drawable.Drawable;
+import android.support.annotation.IntDef;
+import android.support.annotation.RestrictTo;
+import android.support.design.R;
+import android.support.v4.view.ViewCompat;
+import android.support.v7.widget.LinearLayoutCompat;
+import android.util.AttributeSet;
+import android.view.View;
+import java.lang.annotation.Retention;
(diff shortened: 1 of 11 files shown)


lib/src/android/support/design/widget/InsetDividerLinearLayout.java
```diff
@@ -0,0 +1,393 @@
...
+
+  private boolean hasMiddleDividerInsets() {
+    return (mApplyInsets & INSET_DIVIDER_MIDDLE) != 0;
+  }
+
+  private boolean hasEndDividerInsets() {
+    return (mApplyInsets & INSET_DIVIDER_END) != 0;
+  }
+
+  int getFirstVisibleChildIndex() {
```
JakeWharton: `@VisibleForTesting`

lib/src/android/support/design/widget/InsetDividerLinearLayout.java
```diff
@@ -0,0 +1,393 @@
...
+    final int count = getChildCount();
+    for (int i = 0; i < count; i++) {
+      if (getChildAt(i).getVisibility() != GONE) {
+        return i;
+      }
+    }
+    return Integer.MAX_VALUE;
+  }
+
+  int getLastVisibleChildIndex() {
```
JakeWharton: `@VisibleForTesting`

lib/src/android/support/design/widget/InsetDividerLinearLayout.java
```diff
@@ -0,0 +1,393 @@
...
+
+    if (hasEndDivider()) {
+      final boolean hasInsets = hasEndDividerInsets();
+      drawHorizontalDivider(canvas, getHeight() - mDividerHeight,
+          hasInsets ? insetStart: 0,
+          hasInsets ? insetEnd : 0);
+    }
+  }
+
+  void drawDividersHorizontal(Canvas canvas) {
```
JakeWharton: `private`? +the next two methods

jestelle:
Thanks for the PR!  We're catching up from holidays and new year planning.  But we'll be giving this more attention soon!

cketcham:
Since this pr is quite old and there were no pings we're going to close it for now. If it is still a desirable addition to the library please ping on this pr. We'll reopen it and take a look.

PR #19, opened 2016-12-22
[BottomNavigationView] Add state save/restore to keep selected tab

### LINKED ISSUE #17: Feature Request - Bottom Navigation View doesn't keep selected tab state
### Overview
Current version of the Bottom Navigation View doesn't keep it's view state when a configuration change takes place, instead, it defaults to the first tab. I think the expected behavior would be to keep the current selected tab state without triggering a new **OnNavigationItemSelectedListener::onNavigationItemSelected()**

nico-gonzalez:
Closes #17 

Adds save/restore state to BottomNavigationView by keeping track of the selected tab id, and resetting it's state to checked when is being restored. Note this does not trigger a call to **OnNavigationItemSelectedListener::onNavigationItemSelected()** which I think it should be the right behavior

### DIFF
--- a/lib/src/android/support/design/internal/BottomNavigationPresenter.java
+++ b/lib/src/android/support/design/internal/BottomNavigationPresenter.java
@@ -19,6 +19,8 @@ package android.support.design.internal;
 import static android.support.annotation.RestrictTo.Scope.LIBRARY_GROUP;
 
 import android.content.Context;
+import android.os.Build;
+import android.os.Bundle;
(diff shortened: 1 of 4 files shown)


jestelle:
Thanks for the PR!  We're catching up from holidays and new year planning.  But we'll be giving this more attention soon!

tec27:
Implemented in 9e2c201185365271677f052b31c8681e38d49526. Thank you for the contribution, sorry we didn't make use of it!

PR #29, opened 2017-02-06
[CodeHealth] Add a travis configuration for this project.

miguelandres:
Currently it builds and runs only with Android emulator for API-level 19, will add support for multiple API levels later down the road.

Modifying the build.gradle file is necessary to speed up builds, pre-dexing is really expensive on clean builds (and CI builds are always clean), so this adds support to disable pre-dexing from the command line: pre-dexing is still enabled by default for faster dev time, but is disabled by command line on CI builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,39 @@
+language: android
+jdk: oraclejdk8
+
+env:
+  - ANDROID_EMULATOR_API_LEVEL=19
+
+android:
+  components:
+    - tools # to get the new `repository-11.xml`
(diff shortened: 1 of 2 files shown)


miguelandres:
@tec27 This seems to be working fine now on Travis CI, will try to get Circle if it's just minor changes, but otherwise we can just stick with good ol' Travis CI.

This CI run is still not done but preemptively sending it off for review since it seems to be in a good path already.

miguelandres:
Closing because this was integrated internally, will come in to github soon.

tec27:
Merged as a562f0608e378af0b44bb248630f3b9fa91360ac

PR #31, opened 2017-02-21
[CoordinatorLayout] Fix saved state restoration on API<13 devices

cbeyls:
On API<13 devices, we always get a `null` ClassLoader in the `CoordinatorLayout.SavedState` constructor. Provide a default one in that case so the Behavior states can be properly restored instead of crashing with a "`ClassNotFoundException when unmarshalling ...`" error.

This is basically the same behavior as other components of the support library like `ViewPager` and `RecyclerView`.

- Related issues on the Android bug tracker: [#196430](https://code.google.com/p/android/issues/detail?id=196430#c14), [#232471](https://code.google.com/p/android/issues/detail?id=232471)


miguelandres:
Along with Support Library, we do not plan to support devices with API < 14, so I'm closing this PR. The current version of the code already sets minSdk to 14.

Sorry for the delay

PR #33, opened 2017-03-06
[BottomNavigationView] Add public setter to BottomNavigation shifting mode

### LINKED ISSUE #22: Feature Request - Bottom Navigation View to provide api to set the shifting mode
The mShiftingMode field controls the shifting behaviour of the Bottom Navigation View. The Design Support Library does not provide at this time an API to set this field as the behaviour is hard coded to be enabled when the count of menu items exceeds the number 3. I understand that this is according to the material design specs. Nevertheless I think the library should not force this on the developers. I found one possible solution to this but it feels more like a hack. http://stackoverflow.com/questions/40176244/how-to-disable-bottomnavigationview-shift-mode

jotaemepereira:
Resolves #22 

I have to add some tests, i need to add a getter of the current status of the shifting mode.

### DIFF
--- a/lib/src/android/support/design/internal/BottomNavigationMenuView.java
+++ b/lib/src/android/support/design/internal/BottomNavigationMenuView.java
@@ -242,6 +242,10 @@ public class BottomNavigationMenuView extends ViewGroup implements MenuView {
     return mItemBackgroundRes;
   }
 
+  public void setShiftingMode(boolean shiftingMode) {
+    mShiftingMode = shiftingMode;
+  }
+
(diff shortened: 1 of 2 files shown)


SUPERCILEX:
It looks like this was fixed in 8493677be92782be3f151047c4affa9b751bc2ee.

Benjiko99:
This PR should be closed, it was indeed done in 8493677 and there has been no activity for a pretty much several years...how is this PR still open?

afohrman:
Closing this PR since this API now exists in the library.

PR #34, opened 2017-03-19
[BottomNavigationView] Added bottom navigation behavior and scrims fix.

alexfacciorusso:
**NOTE** work in progress!

Added API for show/hide BottomNavigationView in reaction to nested scrolling (with default Behavior) and support to translucent navigation bar.

The last thing remaining is to fix the Snackbar and make it dodge the navigation view!
Attached video with features + bug. I think the FAB shouldn't fall under navigation bar, but maybe is a fault of my demo app. 

https://youtu.be/2CCYDU6R9Ik

Here is the layout used in the demo:
```xml
<android.support.design.widget.CoordinatorLayout ...>
...
<android.support.design.widget.BottomNavigationView
        android:id="@+id/bottom_navigation"
        android:layout_gravity="bottom"
        android:background="@color/colorPrimary"
        app:elevation="8dp"
        app:itemBackground="@color/colorPrimary"
        app:itemIconTint="@android:color/white"
        app:itemTextColor="@android:color/white"
        app:menu="@menu/navigation" />
</android.support.design.widget.CoordinatorLayout>
```

### DIFF
--- a/lib/res/anim/design_bottomnav_in.xml
+++ b/lib/res/anim/design_bottomnav_in.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+  ~ Copyright (C) 2015 The Android Open Source Project
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~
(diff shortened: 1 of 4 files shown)


ymarian:
I think BottomNav is meant to be fixed. The spec doesn't cover this so if added It should be done outside the library. BottomAppBar does support this behavior.

PR #35, opened 2017-05-09
Added badges

appsforartists:
See also https://github.com/material-components/material-components-web/pull/601 and https://github.com/material-components/material-components-ios/pull/1405


miguelandres:
@tec27 Sorry, i merged this one thinking this was the remixer notification, how do we bring this into the Google internal repo?

PR #45, opened 2017-09-25
Add configuration for aar release

miguelandres:
./gradlew uploadArchives generates a maven repo witn everything we need in /tmp/myRepo

### DIFF
--- a/lib/src/android/support/design/animation/build.gradle
+++ b/lib/src/android/support/design/animation/build.gradle
@@ -2,8 +2,7 @@ apply plugin: 'com.android.library'
 apply plugin: 'com.github.dcendents.android-maven'
 
 archivesBaseName = 'design-animation'
-version = '1.0'
-
+version = rootProject.ext.mdcLibraryVersion
 dependencies {
   compile "com.android.support:support-annotations:${project.rootProject.ext.supportVersion}"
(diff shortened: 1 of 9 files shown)


lib/src/android/support/design/animation/build.gradle
```diff
@@ -1,6 +1,8 @@
 apply plugin: 'com.android.library'
+apply plugin: 'com.github.dcendents.android-maven'
 
-archivesBaseName = 'design'
+archivesBaseName = 'design-animation'
```
pingpongboss: just an idea, but this string can probably be autogen'd

lib/src/android/support/design/backlayer/build.gradle
```diff
@@ -1,6 +1,8 @@
 apply plugin: 'com.android.library'
+apply plugin: 'com.github.dcendents.android-maven'
 
-archivesBaseName = 'design'
+archivesBaseName = 'design-backlayer'
+version = '1.0'
```
pingpongboss: do we want this version to be consistent? Maybe a similar technique to ${project.rootProject.ext.supportVersion} can be used.

miguelandres:
This is incomplete, I haven't sent it for review yet (assigned any
reviewer), in fact I'm creating this pr in order to be able to import the
change internally when it's done and review with internal tools

PR #54, opened 2018-01-19
Kind of fix Travis build

SUPERCILEX:
This PR gets licenses and all of that working, but there are a bunch of compile errors because this lib is completely broken when it comes to the Gradle build. 😕

PR #55, opened 2018-01-19
Save bottom sheet state on config change

SUPERCILEX:
See https://android-review.googlesource.com/c/platform/frameworks/support/+/593733


yaraki:
@tec27 Hi, Travis. The CI seems to be failing due to some configuration error. Can you please take a look?

SUPERCILEX:
@yaraki thanks for pointing me in the right direction! @tec27 see #54, it looks like the Gradle modules aren't setup correctly so all the packages are wrong and nothing compiles.

SUPERCILEX:
@dsn5ft Can you take a look?

dsn5ft:
Hi @SUPERCILEX, we have submitted this change internally and it will be included in an upcoming release, as well as synced here to GitHub. Thank you!

SUPERCILEX:
SGTM

PR #14, opened 2016-12-20
[TextInputLayout] Add baseline alignment support

hkurokawa:
Closes #13 

Though #13 is not accepted yet, I am preparing this PR to describe the details of the resolution I proposed in #13. The PR should be closed if #13 is rejected.

If you think I am jumping the gun, feel free to close this PR.

Thank you.

PR #56, opened 2018-01-20
[TextInputLayout] Add baseline alignment support

### LINKED ISSUE #13: Feature Request - Add baseline alignment support to TextInputLayout
#### Context, problem or idea, solution or next step
Since 24.2.0, there is no way to align the baseline of the EditText surrounded by a TextInputLayout with the baseline of another component. I believe there are many situations where baseline alignment is required, such as showing an icon on the left of the input field, showing a unit label on the right of the field, and so on.

I am wondering if TextInputLayout could return the baseline position of its child with the `getBaseline()` method, the behaviour would be consistent and it would be helpful to many developers.

Thanks.

#### Screenshots/Video (if possible)
N/A

#### Stack trace (if possible)
N/A

### Reproduction steps
In the below layout, `@id/my_edit_text` and `@id/my_unit_label` are not aligned on their baseline.

```xml
<LinearLayout
    android:orientation="horizontal">
    <android.support.design.widget.TextInputLayout
        android:layout_width="0dp"
        android:layout_weight="1">

        <android.support.design.widget.TextInputEditText
            android:id="@+id/my_edit_text"
            android:inputType="text" />

    </android.support.design.widget.TextInputLayout>

    <TextView
        android:id="@+id/my_unit_label"
        android:text="@string/unit_label"/>
</LinearLayout>
```

### Version number
25.1.0

### Operating system and device
All Android OS and Android devices

### Related issues
https://code.google.com/p/android/issues/detail?id=224553

SUPERCILEX:
Fixes #14, fixes #13

Simpler version of those above issues.


SUPERCILEX:
@dsn5ft Mind taking a look at this one?

SUPERCILEX:
@cketcham bump

lib/java/com/google/android/material/textfield/TextInputLayout.java
```diff
@@ -795,6 +795,12 @@ public EditText getEditText() {
     return editText;
   }
 
+  @Override
+  public int getBaseline() {
+    EditText text = getEditText();
+    return text == null ? super.getBaseline() : text.getPaddingTop() + text.getBaseline();
```
afohrman: Have you verified that this works? getPaddingTop() alone won't provide the correct baseline, since TextInputLayout handles top spacing differently based in its boxBackgroundMode. calculateLabelMarginTop() is probably closer to what you want, although I don't think that will be enough to provide the correct baseline, either. You should also be aware that TextInputLayout adds spacing to the bottom of the box background, (boxBottomOffsetPx) which would throw off the calculation unless it's accounted for.

You should be able to get the correct baseline by offsetting the EditText's baseline by calculateLabelMarginTop(), getPaddingTop(), and the bottom offset.

SUPERCILEX: Are you sure all that is necessary? We've been using it for several months now in [FirebaseUI](https://github.com/firebase/FirebaseUI-Android/blob/4645d8c0567f15e4c455a2212a37e0ef81e27f03/auth/src/main/java/com/firebase/ui/auth/util/ui/BaselineTextInputLayout.java#L23-L27) without any problems. Also, [both](https://stackoverflow.com/q/39974041/4548500) of [these](https://stackoverflow.com/q/32710660/4548500) Stack Overflow questions found the same resolution.

If you think it won't work, would you mind explaining what `boxBackgroundMode` is and how I could activate it?

Example screenshot:
screenshot_20180904-105605

SUPERCILEX:
@afohrman Left a comment [here](https://github.com/material-components/material-components-android/pull/56#discussion_r215012110).

lib/java/com/google/android/material/textfield/TextInputLayout.java
afohrman: Yep, all that is necessary. This solution would help with aligning legacy text fields with other legacy text fields, because legacy text fields only add top padding to make room for the label. The text field boxes do other calculations in order to align properly, so these calculations need to be taken into account.

The currently proposed getBaseline() produces this effect when put in a LinearLayout with baselineAligned set to "true":

image

And this is the effect of a getBaseline() method that takes those factors into account:

image

The getBaseline() method in the screenshot above uses the following code:
```
  @Override
  public int getBaseline() {
    return editText != null
        ? editText.getBaseline() + calculateLabelMarginTop() + getPaddingTop() +
        : super.getBaseline();
  }
```

boxBottomOffsetPx should only be added if boxBackgroundMode != BOX_BACKGROUND_NONE.

SUPERCILEX: Ohhh, I see. I haven't used the Material theme yet so I didn't notice that. Thanks for the great catch!

SUPERCILEX:
@afohrman Addressed all your comments 👍

lib/java/com/google/android/material/textfield/TextInputLayout.java
```diff
@@ -798,7 +798,7 @@ public EditText getEditText() {
   @Override
   public int getBaseline() {
     if (editText != null) {
-      return editText.getPaddingTop() + editText.getBaseline();
+      return editText.getBaseline() + getPaddingTop() + calculateLabelMarginTop();
```
afohrman: Looking good, but boxBottomOffsetPx still needs to be accounted for. It should be added to the calculation only if boxBackgroundMode != BOX_BACKGROUND_NONE. You may want to split that logic out into a separate method.

SUPERCILEX: #### I'm not sure that's right. Here's a screenshot _without_ the extra padding:
screenshot_20180904-153231

#### And here is _with_ the extra padding:
screenshot_20180904-153407


BTW, here's my code:
```java
int padding = getPaddingTop();
if (boxBackgroundMode != BOX_BACKGROUND_NONE) {
  padding += boxCollapsedPaddingTopPx;
}
```

afohrman: The extra offset to account for is boxBottomOffsetPx, not boxCollapsedPaddingTopPx.

SUPERCILEX: Nope, that will move the legacy widget down. The first screenshot was fine... what is there to fix?

afohrman: It shouldn't move the legacy widget down if the offset is only being added when boxBackgroundMode != BOX_BACKGROUND_NONE.

From what I can see in the first screenshot, the legacy text field is a little off. Can you add a screenshot that includes a TextView, EditText, and the TextInputLayout legacy and updated styles side by side?

SUPERCILEX: Would you mind giving me the XML code you used in your screenshot?

Sidenote: what would the _bottom_ offset have to do with anything? The baseline only cares about what's above it AFAIK.

SUPERCILEX: #### @afohrman This is with current code, I really don't see what's wrong in terms of baseline:
screenshot_20180906-195442

afohrman: Cool, this looks great. I'll merge the change.

Thanks for contributing!

afohrman: Oh never mind, that screenshot was using an earlier version of the code. I've verified that it works and I'll merge the changes.

Thank you!

SUPERCILEX: Phew! 😌 Can't wait to see this finally go through!!! 😁👏🎉

SUPERCILEX:
@afohrman anything holding you back from merging?

dsn5ft:
@SUPERCILEX this has been submitted internally and will be synced out to GitHub soon.

SUPERCILEX:
@dsn5ft Awesome! ❤️

SUPERCILEX:
@dsn5ft would you mind also closing https://issuetracker.google.com/issues/113439293? Thanks!

Issue #61, opened 2018-02-21
MenuView performs rebinding/layout/inflation on every click event


AttwellBrian:
### Overview
Clicking on a menu entry causes the menu's `notifyDataSetChanged()` to be invoked internally. This causes text appearance to be rebound and a few views to be re-inflated as well.  This adds 30-40ms of delay to opening a menu entry on a high end phone. See screenshot of trace. Or open the attached trace file in chrome:tracing.

Note: this trace is based on an internal fork of android that inserts a couple arm instructions in every method call. The tracing overhead is only a few percentage points. So this trace file is trustworthy. But in case you have reasonable doubts, the systrace screenshot shows similar issues & performance overhead.

### Reproduction steps

1. Create app with a MenuView and 10ish menu entries. The Uber app, for example :)
2. Start your profiler of choice
3. Click on a menu entry

### Version number

Tested on Android 6P with android 7.1.2, with design library version 27.0.2.

### Attached

To view the full trace file, download the trace from https://issuetracker.google.com/issues/73723207 where I originally posted this issue. Then open it in chrome:tracing.

vmtrace screenshot
<img width="928" alt="vmtracing_screenshot_menu_click" src="https://user-images.githubusercontent.com/549900/36507616-21706a6c-170f-11e8-8685-0d0de6e6e8d8.png">

systrace screenshot
<img width="358" alt="systrace_screenshot_menu_click" src="https://user-images.githubusercontent.com/549900/36507650-3cae7ab2-170f-11e8-90f2-6a9a1dabf95c.png">

AttwellBrian:
@wcshi thanks for merging!

PR #62, opened 2018-02-21
[NavigationMenu] Don't rebind/relayout/reinflate any views after clicking a menu item

AttwellBrian:
If none of the check state has changed let's not update the menu. While larger improvements may make sense, this diff is designed to mitigate the most serious issue in a way that doesn't create risk.

Background for this diff is at https://github.com/material-components/material-components-android/issues/61. This diff should close the issue.

Test: used these modified classes inside the Android Uber app and ran a profiler. The problematic 50ms delay caused by this bug was fixed. You can see the before/after traces in https://issuetracker.google.com/issues/73723207.

### DIFF
--- a/lib/java/android/support/design/internal/NavigationMenuPresenter.java
+++ b/lib/java/android/support/design/internal/NavigationMenuPresenter.java
@@ -323,13 +323,16 @@ public class NavigationMenuPresenter implements MenuPresenter {
           setUpdateSuspended(true);
           MenuItemImpl item = itemView.getItemData();
           boolean result = menu.performItemAction(item, NavigationMenuPresenter.this, 0);
+          boolean checkStateChanged = false;
           if (item != null && item.isCheckable() && result) {
             adapter.setCheckedItem(item);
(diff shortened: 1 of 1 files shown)


AttwellBrian:
Travis is going to fail given that master has been red for a while. This doesn't mean anything is wrong with the diff.

lib/java/android/support/design/internal/NavigationMenuPresenter.java
```diff
@@ -323,13 +323,16 @@ public void onClick(View v) {
...
             adapter.setCheckedItem(item);
+            checkStateChanged = true;
           }
           setUpdateSuspended(false);
-          updateMenuView(false);
+          if (checkStateChanged) {
+            updateMenuView(false);
+          }
         }
-      };
```
JakeWharton: Revert

AttwellBrian: Well. I feel silly.

Thanks for spotting that.

PR #96, opened 2018-05-09
[Snackbar] Match BaseTransientBottomBar duration docs

Jawnnypoo:
If you look at `BaseTransientBottomBar.setDuration()` you are able to set the duration to a custom millisecond value. This is not reflected in the `Snackbar.make` methods. This updates the docs for these.


cketcham:
The documentation has been updated

Issue #112, opened 2018-05-18
Attribute to customize Bounding box stroke


RowlandOti:
I can see that the bounding box for the `TextInputLayout` has been explicitly declared here:

```java
 defaultStrokeColor =
        ContextCompat.getColor(context, R.color.mtrl_textinput_default_box_stroke_color);
    disabledColor = ContextCompat.getColor(context, R.color.mtrl_textinput_disabled_color);
    hoveredStrokeColor =
        ContextCompat.getColor(context, R.color.mtrl_textinput_hovered_box_stroke_color);
```

i.e : 

https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/textfield/TextInputLayout.java#L314

Are there plans to provide an attribute that allows developers, to for example override these? For example, use color states?

snotyak:
I should mention that while this default looks alright on when using light backgrounds, it looks terrible with dark ones.

petrkubac:
Hello guys, you can override the color in your colors.xml
` <color name="mtrl_textinput_default_box_stroke_color" tools:override="true">@color/white_40</color>`

RowlandOti:
Bad approach, @petrkubac  we may want to use themes. Attributes are a better alternative.

petrkubac:
@RowlandOti I have updated my post to make it more understandable what I was talking about. You can still use themes, but if you need to change the default box stroke colour now (until the issue is resolved), you can override it like I did.

RowlandOti:
Cool. Thanks for clarifying, it's currently the solution I was using, as it seems to be the only alternative at the moment. It is quite unsatisfactory, hence I look forward to a fix from the team.

asmaa-adel:
@afohrman any updates for this issue as i have more than one theme and in each one i want deferent default color for the box stroke

mrZizik:
What is status of this issue?

afohrman:
We are aware of this issue and we plan to add a defaultBoxStrokeColor (or similarly named) attribute so that this value can be parameterized. We plan to do this after we finish the color system, but we'd be happy to accept a pull request for this if anyone would like to contribute.

selena775:
~petrkubac , what if we want to have two different TextInputLayouts depending on a theme. How can we use mtrl_textinput_default_box_stroke_color to set two different colours. TextInputLayout should be the very foundation of UI controls. Please provide a solution.

RowlandOti:
[...]

PR #181, opened 2018-07-31
[TextInputLayout] Creates attr defaultBoxStrokeColor for TextInputLayout

BugsBunnyBR:
This PR closes https://github.com/material-components/material-components-android/issues/112
by adding `defaultBoxStrokeColor` in TextInputLayout attr.

### DIFF
--- a/lib/java/com/google/android/material/textfield/TextInputLayout.java
+++ b/lib/java/com/google/android/material/textfield/TextInputLayout.java
@@ -313,8 +313,13 @@ public class TextInputLayout extends LinearLayout {
       defaultHintTextColor =
           focusedTextColor = a.getColorStateList(R.styleable.TextInputLayout_android_textColorHint);
     }
-    defaultStrokeColor =
-        ContextCompat.getColor(context, R.color.mtrl_textinput_default_box_stroke_color);
+    if (a.hasValue(R.styleable.TextInputLayout_defaultBoxStrokeColor)) {
+      defaultStrokeColor =
+          a.getColor(R.styleable.TextInputLayout_defaultBoxStrokeColor, Color.TRANSPARENT);
+    } else {
+      defaultStrokeColor =
+          ContextCompat.getColor(context, R.color.mtrl_textinput_default_box_stroke_color);
+    }
     disabledColor = ContextCompat.getColor(context, R.color.mtrl_textinput_disabled_color);
     hoveredStrokeColor =
         ContextCompat.getColor(context, R.color.mtrl_textinput_hovered_box_stroke_color);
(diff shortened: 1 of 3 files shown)


lib/java/com/google/android/material/textfield/TextInputLayout.java
```diff
@@ -313,8 +313,13 @@ public TextInputLayout(Context context, AttributeSet attrs, int defStyleAttr) {
...
     }
-    defaultStrokeColor =
-        ContextCompat.getColor(context, R.color.mtrl_textinput_default_box_stroke_color);
+    if (a.hasValue(R.styleable.TextInputLayout_defaultBoxStrokeColor)) {
+      defaultStrokeColor =
+          a.getColor(R.styleable.TextInputLayout_defaultBoxStrokeColor, Color.TRANSPARENT);
+    } else {
+      defaultStrokeColor =
+          ContextCompat.getColor(context, R.color.mtrl_textinput_default_box_stroke_color);
+    }
```
dsn5ft: Any reason you do the hasValue() check instead of using the default value parameter?

    defaultStrokeColor =
        a.getColor(
            R.styleable.TextInputLayout_defaultBoxStrokeColor,
            ContextCompat.getColor(context, R.color.mtrl_textinput_default_box_stroke_color));

BugsBunnyBR: This approach avoids the computation of the getColor, but I can remove it since your approach is easier to maintain.

MGaetan89:
It would be nice to update the corresponding documentation to mention this new attribute.

lib/java/com/google/android/material/textfield/res-public/values/public.xml
```diff
@@ -35,6 +35,7 @@
   <public name="boxPaddingBottom" type="attr"/>
   <public name="boxStrokeColor" type="attr"/>
   <public name="boxStrokeWidth" type="attr"/>
+  <public name="defaultBoxStrokeColor" type="attr"/>
```
afohrman: "Default" is a little confusing in an attribute, since it's not clear what the attribute change - does it change the behavior of the default? The color of the box stroke that's used unless boxStrokeColor is set?

This was sort of ok as an internal name, but ideally we use something other than default for the public facing attribute. It would be nice to use boxStrokeColor, but that already exists as the focusedBoxStrokeColor. Rather than overhauling all of the box stroke attributes, how about naming this something like "unfocusedBoxStrokeColor" or "unfocusedBoxStrokeColor", or a better name that implies that it's the color that's used for the non-focused state?

BugsBunnyBR: I was with some free time so I tried the change...placed it in another branch while waiting a feedback from this pr... https://github.com/BugsBunnyBR/material-components-android/tree/color-state-list

douglasiacovelli: Any news about this PR, guys? It seems reasonable to me that the attribute `boxStrokeColor` could receive a ColorStateList

BugsBunnyBR:
I am closing this PR as the maintainers stoped looking at this PR.

PR #187, opened 2018-08-22
[FloatingActionButton] Public setter for 'maxImageSize'

mateuswsouza:
Make it is possible to change the maxImageSize programatically along with customSize and other FAB custom attributes.


mateuswsouza:
Signed it.

ymarian:
Shouldn't this be in the FloatingActionButton instead of ShadowDelegateImpl? 
Also I think the first commit updating the field. Is the right thing since that seems to be used to calculate padding?

mateuswsouza:
I didn't see the reason of updating the field in FloatingActionButton.java, as the Impl seems to be handling some custom Material Design properties such as "HoveredFocusedTranslationZ" and "ShowMotion", but didn't study the whole code enough to do the right changes, it didn't compile. Will try to up fix it untill the next week, but would love contributions @ymarian.

PR #211, opened 2018-10-07
[MaterialButton] Use Layout#getWidth instead of Paint#measureText

alexjlockwood:
Using `Paint#measureText()` to measure the text's visual width is definitely wrong here, because that method will return the total length of the text regardless of how it is laid out (Paint doesn't know anything about how the text ends up using line breaks and stuff like that).

Using the text view's Layout object is necessary in order to obtain a reliable measurement of the text's visual width.

This PR fixes the bug filed here: https://issuetracker.google.com/issues/117365856

Here is a before vs. after screenshot:

material-button-icon-bug

#### Don't forget:

### DIFF
--- a/lib/java/com/google/android/material/button/MaterialButton.java
+++ b/lib/java/com/google/android/material/button/MaterialButton.java
@@ -333,7 +333,8 @@ public class MaterialButton extends AppCompatButton {
     }
 
     Paint textPaint = getPaint();
-    int textWidth = getLayout().getWidth();
+    int textWidth = 
+        Math.min((int) textPaint.measureText(getText().toString()), getLayout().getWidth());
     int localIconSize = iconSize == 0 ? icon.getIntrinsicWidth() : iconSize;
     int newIconLeft =
         (getMeasuredWidth()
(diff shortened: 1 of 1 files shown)


ymarian:
Thanks a lot for the detailed bug and the fix!! 
I think we actually need both solutions and take the min between the two values. Do you think that covers everything?

The paint measure is needed for the case where the button is much bigger than the text. 

To repro you can have your button as in the layout you provided. Set the width to say 200dp and have a short text on it.

lib/java/com/google/android/material/button/MaterialButton.java
```diff
@@ -333,7 +333,7 @@ protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
     }
 
     Paint textPaint = getPaint();
-    int textWidth = (int) textPaint.measureText(getText().toString());
+    int textWidth = getLayout().getWidth();
```
ymarian: int textWidth = Math.min((int) textPaint.measureText(getText().toString()), getLayout().getWidth());

alexjlockwood: Good point, updated.

(Although I still wonder if there is a better way to do this w/o using `Paint#measureText()`...)

PR #214, opened 2018-10-10
[MaterialButton] Possible typo? Using `colorOnSurface` instead of `colorSurface` as the button's background

alexjlockwood:
My understanding of these attributes is that the "on" attributes (i.e. `colorOnPrimary`, `colorOnSurface`, etc.) are supposed to be used to describe text/icons when displayed on top of a background... and that the others (i.e. `colorPrimary`, `colorSurface`, etc.) are supposed to be used to describe the actual backgrounds.

If this is true, is it possible that there is a typo here and that you meant to use `colorSurface` instead?

#### Don't forget:


alexjlockwood:
Actually never mind, ignore me... I guess the button is the thing being displayed on top of the surface, so that's why it makes sense here I guess. :)

PR #228, opened 2018-11-10
[BottomSheetBehavior] Fix wrong position and state when touches settling view

Dong-wan:
**Issues when user touches the settling bottom sheet**
**https://issuetracker.google.com/issues/119289861**

&nbsp;

### 1. When touches normal child view in settling bottom sheet
User starts dragging the sheet and the settling animation is stopped immediately,
but the state of the sheet goes wrong to STATE_EXPANDED or STATE_COLLAPSED while dragging.

STATE_SETTLING > STATE_DRAGGING > STATE_EXPANDED or STATE_COLLAPSED (by `SettleRunnable`)

`SettleRunnable`'s `targetState` should be ignored when settling is aborted.
And the state should be remained STATE_DRAGGING.

- Videos
**[Before]** https://drive.google.com/open?id=1vvlI7_AtiAze1PcBfElRolC_LQvAVmQt
**[After(Fixed)]** https://drive.google.com/open?id=1G9387bRw4EzI7XOMLF-Hm2ro1627hsll

&nbsp;

### 2. When touches nested scroll child view in settling bottom sheet
User starts dragging the sheet, but the settling animation is not stopped,
and after the animation finished, the sheet's position is suddenly changed.
And also, nested scroll view could be scrolled while settling.

In settling state, the child view should be captured by `ViewDragHelper`, not the nested scroll child view.

- Videos
**[Before]** https://drive.google.com/open?id=1Q80Cuam0_5kjntzWySJKZoHBS1lq9OHb
**[After(Fixed)]** https://drive.google.com/open?id=1AORJELc322sg-daapgfpZyIevo5V1kEy

### DIFF
--- a/lib/java/com/google/android/material/bottomsheet/BottomSheetBehavior.java
+++ b/lib/java/com/google/android/material/bottomsheet/BottomSheetBehavior.java
@@ -288,10 +288,12 @@ public class BottomSheetBehavior<V extends View> extends CoordinatorLayout.Behav
       case MotionEvent.ACTION_DOWN:
         int initialX = (int) event.getX();
         initialY = (int) event.getY();
-        View scroll = nestedScrollingChildRef != null ? nestedScrollingChildRef.get() : null;
-        if (scroll != null && parent.isPointInChildBounds(scroll, initialX, initialY)) {
-          activePointerId = event.getPointerId(event.getActionIndex());
-          touchingScrollingChild = true;
+        if (state != STATE_SETTLING) {
+          View scroll = nestedScrollingChildRef != null ? nestedScrollingChildRef.get() : null;
+          if (scroll != null && parent.isPointInChildBounds(scroll, initialX, initialY)) {
+            activePointerId = event.getPointerId(event.getActionIndex());
+            touchingScrollingChild = true;
+          }
         }
         ignoreEvents =
             activePointerId == MotionEvent.INVALID_POINTER_ID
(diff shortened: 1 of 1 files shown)


Dong-wan:
@cketcham, Could you please review this commit? Thank you!

lib/java/com/google/android/material/bottomsheet/BottomSheetBehavior.java
```diff
@@ -288,10 +288,12 @@ public boolean onInterceptTouchEvent(CoordinatorLayout parent, V child, MotionEv
       case MotionEvent.ACTION_DOWN:
         int initialX = (int) event.getX();
         initialY = (int) event.getY();
-        View scroll = nestedScrollingChildRef != null ? nestedScrollingChildRef.get() : null;
-        if (scroll != null && parent.isPointInChildBounds(scroll, initialX, initialY)) {
-          activePointerId = event.getPointerId(event.getActionIndex());
-          touchingScrollingChild = true;
+        if (state != STATE_SETTLING) {
```
cketcham: Does this mean we won't intercept touch events while the view is settling? What if the user wants to stop the current settle animation to move the BottomSheet?

Dong-wan: It does not mean we won't intercept touch event.
This code block is **for scrolling of nested child**, not for dragging the bottom sheet.
Dragging events will be processed by `viewDragHelper`, not by here.

As you mentioned, when user wanted to stop the current settle animation, it was not worked if the nested scroll child is existed, and the nested scroll child was scrolled instead.
So I fixed it with this CL.

When `touchingScrollingChild` is set `true`, `viewDragHelper` will **not** capture the view that should be dragged, and then, nested scrolling will be performed.
So, we have to leave `touchingScrollingChild` as `false` to disable nested scrolling in settling state.

cketcham: Thanks for the explanation. That makes sense 👍

ymarian:
committed here: https://github.com/material-components/material-components-android/commit/e49cd32a4367e13b53a13c9eca22564c297f5317

PR #253, opened 2018-12-22
[HideBottomViewOnScrollBehavior]Height increase the bottomMargin of the view

Bakumon:
Because the Behavior is to slide down to disappear, the more reasonable sliding distance should be the height of the view plus the bottomMargin of the view.


Bakumon:
I signed it!

PR #254, opened 2018-12-26
[IndicatorViewController] Fix a caption view animations issue when you set two error subsequently.

shuhart:
You can reproduce that if you set different errors simultaneously. Animation is canceled but the caption view is stuck right below the box of TextInputLayout without any appropriate padding.

image

### DIFF
--- a/lib/java/com/google/android/material/textfield/IndicatorViewController.java
+++ b/lib/java/com/google/android/material/textfield/IndicatorViewController.java
@@ -238,8 +238,8 @@ final class IndicatorViewController {
                 }
               }
               if (captionViewToShow != null) {
-                captionViewToShow.setTranslationY(0);
-                captionViewToShow.setAlpha(1);
+                captionViewToShow.setTranslationY(0f);
+                captionViewToShow.setAlpha(1f);
               }
             }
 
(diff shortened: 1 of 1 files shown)


darkyuhoo01:
Comment on pull request

lib/java/com/google/android/material/textfield/IndicatorViewController.java
```diff
@@ -237,6 +237,10 @@ public void onAnimationEnd(Animator animator) {
                   errorView.setText(null);
                 }
               }
+              if (captionViewToShow != null) {
+                captionViewToShow.setTranslationY(0);
```
afohrman: `setTranslationY()` and `setAlpha()` take in floats, so the arguments passed in should be changed from floats to ints.

```suggestion
                captionViewToShow.setTranslationY(0f);
```

lib/java/com/google/android/material/textfield/IndicatorViewController.java
```diff
@@ -237,6 +237,10 @@ public void onAnimationEnd(Animator animator) {
...
+                captionViewToShow.setAlpha(1);
```
afohrman:
```suggestion
                captionViewToShow.setAlpha(1f);
```

PR #273, opened 2019-01-23
[build.gradle, cat_transformation_fragment.xml] fixed bug with Vector Drawable in Android SDK < 21

a23sokolov:
* for appropriate work vector on Android SDK < 21 in XML use app:srcCompat="@drawable/example"

### DIFF
--- a/catalog/build.gradle
+++ b/catalog/build.gradle
@@ -51,6 +51,7 @@ def srcDirs = [
 
 android {
   defaultConfig {
+    vectorDrawables.useSupportLibrary = true
     manifestPlaceholders = [
       application_name             : 'CatalogApplication',
       application_theme            : 'Catalog',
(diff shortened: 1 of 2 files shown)


catalog/build.gradle
```diff
@@ -51,6 +51,7 @@ def srcDirs = [
 
 android {
   defaultConfig {
+    vectorDrawables.useSupportLibrary = true
```
ymarian: This isn't necessary. I think the default might be true now?

catalog/java/io/material/catalog/transformation/res/layout/cat_transformation_fragment.xml
```diff
@@ -78,7 +78,7 @@
           android:background="?selectableItemBackgroundBorderless"
           android:contentDescription="@string/close_sheet"
           android:scaleType="fitCenter"
-          android:src="@drawable/ic_close_vd_theme_24px"
+          app:srcCompat="@drawable/ic_close_vd_theme_24px"
           android:tint="?attr/colorOnSurface"/>
```
ymarian: nit: Swap these two attributes so the android namespace attributes are before the app ones.

catalog/build.gradle
a23sokolov: I'm not sure! cause if you don't set this flag, will be generated png files for vectors if your minSdk lower 5.0

PR #299, opened 2019-03-11
[AppBar] AppBarLayout + WindowInset handling fixes

chrisbanes:
When AppBarLayout is used with a CollapsingToolbarLayout child, it currently does not handle WindowInsets very well. This commit fixes that by properly handling
insets.

The primary issue is when an AppBarLayout is used with a single Toolbar (very common use case), where the Toolbar will be displayed behind the status bar.

Another issue is with HeaderScrollingViewBehavior. To workaround CoordinatorLayout using its 'compatible measuring' for insets, it manually sets the scrolling view to fit system windows. This has other issues though because by default that view will get padded in. Fixed by manually adding the insets back into the measured dimensions.


kfaraj:
I submitted an issue https://github.com/material-components/material-components-android/issues/2120 and a pull request https://github.com/material-components/material-components-android/pull/2121 to complete this fix

PR #345, opened 2019-05-06
Set local night mode when toggle theme

G00fY2:
Set [local night mode](https://developer.android.com/reference/android/support/v7/app/AppCompatDelegate#setlocalnightmode) when toggle night theme to immediately apply the changes.


G00fY2:
I signed it!

dsn5ft:
Hello! Thanks for the PR. The AppCompat team recently did some work to improve AppCompatDelegate#setDefaultNightMode, so it now handles recreating your activity and applying DayNight. Very soon I will update our AppCompat version and this shouldn't be necessary.

PR #348, opened 2019-05-07
[BottomSheet] Remove rounded bottom corners of bottomsheets

hrach:
This PR removes bottom corners of bottomsheet. They shouldn't be shaped at all, as [specified in material docs](https://material.io/design/components/sheets-bottom.html#theming).

> *Bottom sheets can only be shaped on the top left and top right corners.

The current bug can be reproduced in catalog app. Maybe this styling was intentional purpose of this [reverted commit](https://github.com/material-components/material-components-android/commit/2fda77a6ab0f4619992cedeace08f6fc67afd133) that actually removed shaping on the top and not the bottom.

### DIFF
--- a/lib/java/com/google/android/material/bottomsheet/res/values/styles.xml
+++ b/lib/java/com/google/android/material/bottomsheet/res/values/styles.xml
@@ -44,6 +44,7 @@
     <item name="enforceMaterialTheme">true</item>
     <item name="android:background">@null</item>
     <item name="shapeAppearance">?attr/shapeAppearanceLargeComponent</item>
+    <item name="shapeAppearanceOverlay">@style/ShapeAppearanceOverlay.MaterialComponents.BottomSheet</item>
     <item name="backgroundTint">?attr/colorSurface</item>
     <item name="android:elevation" tools:ignore="NewApi">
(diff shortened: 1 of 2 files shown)


hrach:
I signed it!

lib/java/com/google/android/material/bottomsheet/res/values/styles.xml
```diff
@@ -44,6 +44,7 @@
     <item name="enforceMaterialTheme">true</item>
     <item name="android:background">@null</item>
     <item name="shapeAppearance">?attr/shapeAppearanceLargeComponent</item>
```
cketcham: We should be able to just set `<item name="shapeAppearance">@null</item>` to remove any rounded corners. Does that work?

hrach: Well, shouldn't bottomsheet have large corners, which are by default 0? Eg. active shaping should add them on the top.

cketcham: Yeah that sounds good. Thanks! We'll merge internally and sync out soon. We'll let you know if we have any other questions.

hrach:
Thanks for merging :)

Issue #216, opened 2018-10-12
NPE in TextInputLayout with hint and empty text on Meizu Phones


DementorAl:
java.lang.NullPointerException: Attempt to invoke virtual method 'int android.text.Layout.getLineForOffset(int)' on a null object reference
        at android.widget.Editor.updateCursorPositionMz(Editor.java:7598)
        at android.widget.Editor.updateCursorsPositions(Editor.java:1932)
        at android.widget.TextView.getUpdatedHighlightPath(TextView.java:5742)
        at android.widget.TextView.onDraw(TextView.java:5928)
        at android.view.View.draw(View.java:17301)
        at android.view.View.updateDisplayListIfDirty(View.java:16283)
        at android.view.ViewGroup.recreateChildDisplayList(ViewGroup.java:3711)
        at android.view.ViewGroup.dispatchGetDisplayList(ViewGroup.java:3691)
        at android.view.View.updateDisplayListIfDirty(View.java:16246)
        at android.view.View.draw(View.java:17067)
        at android.view.ViewGroup.drawChild(ViewGroup.java:3727)
        at android.view.ViewGroup.dispatchDraw(ViewGroup.java:3513)
        at android.view.View.draw(View.java:17304)
        at com.google.android.material.textfield.TextInputLayout.draw(TextInputLayout.java:1577)
        at android.view.View.updateDisplayListIfDirty(View.java:16283)
        at android.view.ViewGroup.recreateChildDisplayList(ViewGroup.java:3711)
        at android.view.ViewGroup.dispatchGetDisplayList(ViewGroup.java:3691)
        at android.view.View.updateDisplayListIfDirty(View.java:16246)
        at android.view.ViewGroup.recreateChildDisplayList(ViewGroup.java:3711)
        at android.view.ViewGroup.dispatchGetDisplayList(ViewGroup.java:3691)
        at android.view.View.updateDisplayListIfDirty(View.java:16246)
        at android.view.ViewGroup.recreateChildDisplayList(ViewGroup.java:3711)
        at android.view.ViewGroup.dispatchGetDisplayList(ViewGroup.java:3691)
        at android.view.View.updateDisplayListIfDirty(View.java:16246)
        at android.view.ViewGroup.recreateChildDisplayList(ViewGroup.java:3711)
        at android.view.ViewGroup.dispatchGetDisplayList(ViewGroup.java:3691)
        at android.view.View.updateDisplayListIfDirty(View.java:16246)
[...]

PR #358, opened 2019-05-13
[TextInputLayout] fix crash when focusing TextInputEditText on Meizu devices

CmoaToto:
Don't return parent hint from TextInputEditText.getHint() if the manufacturer is Meizu as their modifications in Textview leads to a crash

solve https://issuetracker.google.com/issues/112105087
(closed because duplicated Github issue: #216 )

### DIFF
--- a/lib/java/com/google/android/material/textfield/TextInputEditText.java
+++ b/lib/java/com/google/android/material/textfield/TextInputEditText.java
@@ -21,6 +21,7 @@ import com.google.android.material.R;
 import android.content.Context;
 import androidx.annotation.Nullable;
 import androidx.appcompat.widget.AppCompatEditText;
+import android.os.Build;
 import android.util.AttributeSet;
 import android.view.View;
 import android.view.ViewParent;
@@ -51,9 +52,10 @@ public class TextInputEditText extends AppCompatEditText {
   @Override
   public CharSequence getHint() {
     // Certain test frameworks expect the actionable element to expose its hint as a label. When
-    // TextInputLayout is providing our hint, retrieve it from the parent layout.
+    // TextInputLayout is providing our hint, retrieve it from the parent layout. Excepting for
+    // Meizu devices which doesn't handle this behaviour correctly and crash when getting focus.
     TextInputLayout layout = getTextInputLayout();
-    if ((layout != null) && layout.isProvidingHint()) {
+    if (layout != null && layout.isProvidingHint() && !Build.MANUFACTURER.equals("Meizu")) {
       return layout.getHint();
     }
     return super.getHint();
--- a/lib/java/com/google/android/material/textfield/TextInputEditText.java
+++ b/lib/java/com/google/android/material/textfield/TextInputEditText.java
@@ -21,6 +21,7 @@ import com.google.android.material.R;
 import android.content.Context;
 import androidx.annotation.Nullable;
 import androidx.appcompat.widget.AppCompatEditText;
+import android.graphics.Canvas;
 import android.os.Build;
 import android.util.AttributeSet;
 import android.view.View;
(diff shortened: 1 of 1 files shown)


CmoaToto:
@cketcham I explained my investigation and my tests in the Google issue here https://issuetracker.google.com/issues/112105087

> What happens on Meizu devices now that we aren't calling getHint() on the TextInputLayout if the hint was only specified on the TextInputLayout?
The layout works correctly, same behaviour as if we had put the hint on the TextInputLayout wich is the same behaviour as on other devices

> Can you debug, step through this function, and step into layout.getHint() and see if that is called correctly?
After further investigation, the crashing Meizu proprietary Editor.updateCursorPositionMz() method (see Google issue link) calls getHint(), but more important, it calls the final method TextView.getHintLayout() which is null as there is no Hint in the TextView.
So there is 2 solutions:
- Meizu doesn't check if the mHintLayout is null as it thinks it is not as getHint() returns a value (the TextInputLayout's one. So the getHint() method is wrong and in this case we must return null (or super.getHint()) for Meizu devices (done by the above commit)
- Excluding a constructor is wrong so I'll try to find a way to make the TextView create a blank HintLayout in order to let the Meizu device work as expected, but this workaround will create an unneeded Layout on all other devices

> Maybe the problem is that super.getHint() isn't called? What happens if we always call super.getHint() for Meizu devices but then return the value from layout.getHint()
Not working

cketcham:
Thanks for the info. This definitely looks like a crash because the device is expecting `getHintLayout()` to not be null even though it could be.

If we can't find anything better, I your current approach could be ok.

One other thought I had is to call something like:

    TextInputLayout layout = getTextInputLayout();
    if ((layout != null) && layout.isProvidingHint() && super.getHint() == null && Build.MANUFACTURER.equals("Meizu")) {
      setHint("");
    }

in `onDraw()` or `onAttachedToWindow()` or some where early enough to create the hint layout for these devices. Does that cause the hint layout to exist?

CmoaToto:
Indeed it is a very good balance between solutions 1 and 2. I tested it and it worked perfectly on several devices. I updated the PR accordingly.

CmoaToto:
Great! Thank you :)

cketcham:
@CmoaToto We're considering moving the call to `onAttachedToWindow()` instead of `onDraw()` can you verify the fix still works if we do that? thanks.

CmoaToto:
@cketcham You're right, this way the fix will be called only once per View creation on not at each onDraw.
I tested and it works perfectly.
I updated the PR accordingly

awenger:
thanks so much @CmoaToto and @cketcham ! This will fix 25% of all our crashes we saw in the last 7 days. Do you already know when this will be released?

cketcham:
The next release should happen around the end of May

adriancretu:
Crash still occurs on some Meizu devices, with all the fixes in place. Looking at the fix vs the crash logs, this happens because `Build.MANUFACTURER.equals("Meizu")` doesn't pass the test when the manufacturer is "meizu". This happens on Meizu M5c unrooted devices.

CmoaToto:
@adriancretu According to several dumps of FlymOS (Meizu) found on Github (https://github.com/search?p=4&q=%22ro.build.display.id%3DFlyme%22&type=Code) you're right, there is like 1 Meizu device on 20 that uses ro.product.manufacturer=meizu instead of ro.product.manufacturer=Meizu. I'll do a PR to fix this. 

FYI I searched but didn't find other ro.product.manufacturer from Meizu.

CmoaToto:
@adriancretu Here is the PR https://github.com/material-components/material-components-android/pull/411

@cketcham Please review it to completely fix this problem :)

robertlevonyan:
I have this crash on material:1.2.0-alpha02

PR #360, opened 2019-05-14
[Snackbar] Add support for changing text & background tint color at runtime

ricknout:
This adds support to `BaseTransientBottomBar` and `Snackbar` to allow for changing of the message text and background overlay color at runtime. The implementations and method names attempt to stay as close as possible to the existing `Snackbar#setActionTextColor` method.

No related GitHub issue, but please see: https://issuetracker.google.com/issues/132576777

#### Don't forget:

### DIFF
--- a/lib/java/com/google/android/material/snackbar/BaseTransientBottomBar.java
+++ b/lib/java/com/google/android/material/snackbar/BaseTransientBottomBar.java
@@ -40,6 +40,8 @@ import android.os.Bundle;
 import android.os.Handler;
 import android.os.Looper;
 import android.os.Message;
+
+import androidx.annotation.ColorInt;
 import androidx.annotation.IdRes;
 import androidx.annotation.IntDef;
 import androidx.annotation.IntRange;
@@ -362,6 +364,21 @@ public abstract class BaseTransientBottomBar<B extends BaseTransientBottomBar<B>
     background.setColor(backgroundColor);
     background.setCornerRadius(cornerRadius);
     return background;
+  }
+
+  void updateThemedBackgroundOverlayColor(@ColorInt int overlayColor) {
+    int backgroundColor = MaterialColors.getColor(view, R.attr.colorSurface);
+    updateThemedBackgroundColors(backgroundColor, overlayColor);
+  }
+
+  private void updateThemedBackgroundColors(@ColorInt int backgroundColor, @ColorInt int overlayColor) {
+    if (!(view.getBackground() instanceof GradientDrawable)) return;
+    int color =
+        MaterialColors.layer(
+            backgroundColor,
+            overlayColor,
+            view.getBackgroundOverlayColorAlpha());
+    ((GradientDrawable) view.getBackground()).setColor(color);
   }
 
   private void updateBottomMargin() {
--- a/lib/java/com/google/android/material/snackbar/BaseTransientBottomBar.java
+++ b/lib/java/com/google/android/material/snackbar/BaseTransientBottomBar.java
@@ -40,8 +40,6 @@ import android.os.Bundle;
 import android.os.Handler;
 import android.os.Looper;
 import android.os.Message;
-
-import androidx.annotation.ColorInt;
 import androidx.annotation.IdRes;
 import androidx.annotation.IntDef;
 import androidx.annotation.IntRange;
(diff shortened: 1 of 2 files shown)


lib/java/com/google/android/material/snackbar/BaseTransientBottomBar.java
```diff
@@ -364,6 +366,21 @@ private Drawable createThemedBackground() {
...
   }
 
+  void updateThemedBackgroundOverlayColor(@ColorInt int overlayColor) {
+    int backgroundColor = MaterialColors.getColor(view, R.attr.colorSurface);
+    updateThemedBackgroundColors(backgroundColor, overlayColor);
+  }
+
+  private void updateThemedBackgroundColors(@ColorInt int backgroundColor, @ColorInt int overlayColor) {
+    if (!(view.getBackground() instanceof GradientDrawable)) return;
+    int color =
```
pavlospt: @ricknout Will they layering of `MaterialColors` work properly with this solution? I mean if for example one uses a blue color as a `colorSurface` and applies a red color as `backgroundColor` would it still be a desirable result?

ricknout: Hmm not so sure. I copied what was done in `BaseTransientBottomBar#createThemedBackground` for consistency, so I assume it would happen either way

pavlospt: @ricknout Okay thanks for explaining! I was just wondering if the alternative provided on the issue you linked, would be better, since this is a direct user action and the color they would like to apply is the one they passed to that method. That's just a thought I was having, but I a still learning the Material internals, so there might be a specific reasoning behind that color layering that I am not aware of :) Thanks again for jumping on the issue and providing a solution so quickly!

ricknout: @pavlospt No problem :-) The alternative in the issue might indeed be better, interested in what the MDC team thinks

dsn5ft: Thanks for the PR! I think the alternative (a getBackground() or setBackgroundTint() method) is preferable because the layering of `colorSurface` and `colorOnSurface` is just meant to provide the default behavior according to the Material spec. If the client wants to override that behavior I think they would probably want to pass in a color that will be used for the background without any layering, and if they do want some layering they could do it before passing in the color.

ricknout: @dsn5ft okay great, thanks for the feedback. I'll close this PR and open a new one with these changes, so that the commit history is clean

dsn5ft: Np! No need to close the PR if you don't want to. We'll submit this change internally anyway and then this PR will get automatically closed when the commit is synced to GitHub. So the commit history from this PR won't make it to master.

ricknout: Ah okay! Will just submit an update then

ricknout: Okay @dsn5ft I've updated the PR. I had to add in a few workarounds to ensure the tint works on all API levels (see the comment in the method)

dsn5ft: looks good!

lib/java/com/google/android/material/snackbar/Snackbar.java
```diff
@@ -355,6 +383,23 @@ public Snackbar setActionTextColor(@ColorInt int color) {
...
 
+  /**
+   * Sets the tint color of the background Drawable.
+   */
+  @NonNull
+  public Snackbar setBackgroundTintColor(@ColorInt int color) {
+    if (view.getBackground() != null) {
+      // Drawable doesn't implement setTint in API 21 and Snackbar does not yet use
+      // MaterialShapeDrawable as its background (i.e. TintAwareDrawable)
+      if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP_MR1) {
```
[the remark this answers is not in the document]
ricknout: Apparently`Drawable#setTint` is not implemented in 21, only from 22 onwards. I got the hint from elsewhere in the MDC-Android codebase, see: https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/tabs/TabLayout.java#L2912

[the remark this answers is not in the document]
dsn5ft: interesting! sounds good to me. submitting this PR internally.

PR #409, opened 2019-06-21
[Bottom Navigation] Add tooltip fallback to Bottom Navigation item titles

ataulm:
Bottom navigation tab titles can get truncated/clipped if they're too long or if the user's text size is set as large:

image

One way to mitigate this is to use tooltips, like AppBar's `ActionMenuItemView`. `BottomNavigationItemView` already sets the tooltip if one is available, but the tooltip property needs to be explicitly set by the client.

This PR adds similar behaviour to the `ActionMenuItemView`: fall back to the title if the tooltip is not present.

Before | After
---|---
tooltips-before | tooltips-after

### DIFF
--- a/lib/java/com/google/android/material/bottomnavigation/BottomNavigationItemView.java
+++ b/lib/java/com/google/android/material/bottomnavigation/BottomNavigationItemView.java
@@ -126,7 +126,11 @@ public class BottomNavigationItemView extends FrameLayout implements MenuView.It
     if (!TextUtils.isEmpty(itemData.getContentDescription())) {
       setContentDescription(itemData.getContentDescription());
     }
-    TooltipCompat.setTooltipText(this, itemData.getTooltipText());
+    if (!TextUtils.isEmpty(itemData.getTooltipText())) {
+      TooltipCompat.setTooltipText(this, itemData.getTooltipText());
+    } else {
+      TooltipCompat.setTooltipText(this, itemData.getTitle());
+    }
     setVisibility(itemData.isVisible() ? View.VISIBLE : View.GONE);
   }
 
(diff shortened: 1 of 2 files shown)


lib/java/com/google/android/material/bottomnavigation/BottomNavigationItemView.java
```diff
@@ -126,7 +126,11 @@ public void initialize(MenuItemImpl itemData, int menuType) {
     if (!TextUtils.isEmpty(itemData.getContentDescription())) {
       setContentDescription(itemData.getContentDescription());
     }
-    TooltipCompat.setTooltipText(this, itemData.getTooltipText());
```
ataulm: I'm unsure how `itemData.getTooltipText()` works on pre-API 26, because [it was only added in that version](https://developer.android.com/reference/android/view/MenuItem#getTooltipText()).

I had to set Robolectric config to 26 to avoid a `NoSuchMethodException`.

lib/javatests/com/google/android/material/bottomnavigation/BottomNavigationItemViewTest.java
```diff
@@ -0,0 +1,95 @@
...
+    itemView.initialize(menuItem, MENU_TYPE);
+
+    CharSequence updatedTitle = "menu item title updated";
+    itemView.setTitle(updatedTitle);
+
+    assertThat(itemView.getTooltipText()).isEqualTo(updatedTitle);
+  }
+
+  private MenuItemImpl createMenuItemImpl(CharSequence title, CharSequence tooltip) {
+    MenuItemImpl menuItem = Mockito.mock(MenuItemImpl.class);
```
ataulm: The constructor for `MenuItemImpl` is package protected.

ataulm:
The failing CATs _look_ unrelated. I tested this on an API 25 emulator and an API 28 device.

PR #415, opened 2019-07-03
[BottomSheet] Updated BottomSheetDialog and Fragment dismiss animation

froriz5:
Updated `BottomSheetDialog` and `BottomSheetDialogFragment` to use the dismiss transition animation from the `BottomSheetBehavior.STATE_HIDDEN` (Swipe Down animation).

Issue: https://issuetracker.google.com/issues/136638478


lib/java/com/google/android/material/bottomsheet/BottomSheetDialog.java
```diff
@@ -112,6 +115,49 @@ protected void onStart() {
...
+  @Override
+  public void cancel() {
+    // DialogSheet was swiped down to STATE_HIDDEN, so safe to defer to parent cancel().
+    if (getBehavior().getState() == BottomSheetBehavior.STATE_HIDDEN) {
+      super.cancel();
+    } else {
+      // DialogSheet cancelled through another method.
+      WeakReference<BottomSheetDialog> dialogWeakReference = new WeakReference<>(this);
+
+      BottomSheetBehavior behavior = getBehavior();
```
cketcham: The default BototmSheetCallback that is set on the dialog will call `cancel()`. Does this still work if you just set the state to hidden, and don't override the callback at all?

froriz5: I think that would work.
However, my only concern is if the default behavior can be overridden.

I know the user can set their own `BottomSheetCallback` in the `BottomSheetBehavior`. So if they override the default callback, than we'd lose the `default implementation which calls `cancel()` on `STATE_HIDDEN` right?

lib/java/com/google/android/material/bottomsheet/BottomSheetDialogFragment.java
```diff
@@ -31,4 +38,38 @@
   public Dialog onCreateDialog(Bundle savedInstanceState) {
     return new BottomSheetDialog(getContext(), getTheme());
   }
+
+  @Override
+  public void dismiss() {
```
cketcham: I think both `dismissAllowingStateLoss()` and `dismiss()` will need to have the same logic which prevents the fragment from being dismissed immediately.

Could this method do something similar to cancel in `BottomSheetDialog`? If the bottom sheet is hidden, call through to super. Otherwise just set the state of the BottomSheet to hidden. That should call through to cancel the dialog, which should call `onCancel(DialogInterface dialog)` in `BottomSheetDialogFragment` which can be modified to call `dismiss()`

froriz5: Yep, I can do that 👍

lib/java/com/google/android/material/bottomsheet/BottomSheetDialog.java
cketcham: Maybe we should set an alternative way to listen? For example, `BottomSheetDialog` can have a method to add a `BottomSheetCallback` and it can delegate to that in it's own callback. That might be the clearest solution.

Or we can check if the callback that is currently set is the one from `BottomSheetDialog` and if it's not, we can wrap it then.

It's kind of broken if someone overrides the callback now though right? Because if it's not using it's callback, when the bottomsheet is hidden, the dialog won't be dismissed?

froriz5: Yeah, I think we'd need to do one of those.
I tested this out with the Catalog app in this project, and noticed that the example there is overriding the behavior itself, so the default callback is completely ignored.

BottomSheetCallbackReferences

I'm gonna add a check to see if the set callback is the default, then to call `super.cancel()`.
If not, I'll instantiate a new one like I have currently.

froriz5: I went with option 2, of checking the currently set `BottomSheetCallback`, to see if it's the default one. If it is, I defer the logic to that callback and simply set the state to `STATE_HIDDEN`.

If it's different, then I instantiate a new callback to override the existing one to handle closing the dialog.

lib/java/com/google/android/material/bottomsheet/BottomSheetDialogFragment.java
froriz5: Done 👍

cketcham:
There's also some window animation stuff that is set up with `mtrl_bottom_sheet_slide_in` and `mtrl_bottom_sheet_slide_out`. If were now just using the animation from the BottomSheet itself, can we remove those?

froriz5:
It looks like the animation for slide_down won't be needed.
However, the enter animation to slide up to reveal is still needed.

I can remove the `android:windowExitAnimation` from the appropriate styles (`Animation.Design.BottomSheetDialog`), if we don't want to delete the actual animation files at this time.

e.g.
```
<style name="Animation.Design.BottomSheetDialog" parent="Animation.AppCompat.Dialog">
    <item name="android:windowEnterAnimation">@anim/design_bottom_sheet_slide_in</item>

    <!--  Delete this one -->
    <!--<item name="android:windowExitAnimation">@anim/design_bottom_sheet_slide_out</item>-->
</style>
```

Or should I delete all references to enter animations?

froriz5:
@cketcham Should I go ahead and delete those animation values I defined above?
Is there also any other feedback for this PR?

lib/java/com/google/android/material/bottomsheet/BottomSheetDialog.java
```diff
@@ -112,6 +115,54 @@ protected void onStart() {
...
+
+      // If the default callback was not overridden, then set the state to STATE_HIDDEN and defer
+      // to the default callback behavior.
+      if (behavior.getBottomSheetCallback() == bottomSheetCallback) {
+        behavior.setState(BottomSheetBehavior.STATE_HIDDEN);
+      } else {
+        WeakReference<BottomSheetDialog> dialogWeakReference = new WeakReference<>(this);
+
+        behavior.setBottomSheetCallback(new BottomSheetBehavior.BottomSheetCallback() {
```
cketcham: Would it work to just set the callback back to `bottomSheetCallback`? It does the same thing right?

froriz5: Yep that work.
Just pushed a commit to address that 👍

cketcham:
@froriz5 Hey sorry for the delay! I think removing `android:windowExitAnimation` and the actual animations defined for those cases is good.

froriz5:
@cketcham Pushed a commit removing those unneeded animation values 👍

cketcham:
Thanks! From what I can tell it seems ok, can you rebase your changes on top of master? Then I can pull it in and run some more tests to make sure things don't break.

froriz5:
@cketcham I rebased off the latest master and pushed the changes 👍

lib/java/com/google/android/material/bottomsheet/BottomSheetDialogFragment.java
```diff
@@ -31,4 +39,87 @@
...
+
+      behavior.setState(BottomSheetBehavior.STATE_HIDDEN);
+    }
+  }
+
+  private BottomSheetBehavior.BottomSheetCallback getDismissBehaviorCallback(
+      final boolean allowingStateLoss) {
+
+    final WeakReference<BottomSheetDialogFragment> dialogFragmentWeakReference =
+        new WeakReference<>(this);
```
froriz5: @cketcham After the rebase, I had to fix two compilation issues

- Had to make the `boolean allowingStateLoss` param in this function `final`, as it was accessed inside the callback.
- Had to change the generic for `WeakReference`
  - `WeakReference<AppCompatDialogFragment>` to `WeakReference<BottomSheetDialogFragment>`

cketcham:
Thanks, I did notice a few tests failed including https://github.com/material-components/material-components-android/blob/master/tests/javatests/com/google/android/material/bottomsheet/BottomSheetDialogTest.java#L150

froriz5:
@cketcham I can't seem to run the test locally for some reason.
When I tried running the `./gradlew connectedAndroidTest`, but only got one test that failed, `AppBarHorizontalScrollingTest#testScrollAndClick`:
Screen Shot 2019-07-30 at 9 46 23 AM



From a quick look at the failing test you linked above, I did notice something.

Looking at the test below that one, `testHideBottomSheet `, which tests Hiding the BottomSheet, it invokes a `SystemClock.sleep(31)` after setting the `STATE_HIDDEN` on the behavior.
Source: https://github.com/material-components/material-components-android/blob/master/tests/javatests/com/google/android/material/bottomsheet/BottomSheetDialogTest.java#L168

I'm thinking the reason this linked test is failing is, as you theorized, we are not performing a similar  `SystemClock.sleep(31)` before canceling the BottomSheet. Instead, what happens now is the Assertion of checking if the dialog is visible occurs immediately after `cancel` is invoked.

I think adding a similar delay mechanic will be needed for this test as well.

cketcham:
I'm starting to think we might actually need to change our approach (because of the number of tests that are failing across all of google's apps). It's unfortunate, but it seems like we're breaking an implicit contract that calling dismiss will immediately dismiss the dialog. If we commit this change, we'll likely break bottom sheet tests for a lot of users.

I think we need to make this behavior opt in. One solution will be to just add a method like `dismissWithAnimation()` or something that will just get the `BottomSheetBehavior` and call hide on it, and then just document this method so people know they can use it.

Probably a better option is to set a flag or call a method on the BottomSheet to `enableHideWithAnimation()`. When this is called it would set a boolean that would let a client opt into the new behavior when `dismiss()` is called. This would also make it easier to eventually migrate to this new behavior by default since we'd be able to set `enableHideWithAnimation()` to true by default in another major release.

froriz5:
@cketcham I agree with your deduction.
I've added a flag, `dismissWithAnimation`, to the `BottomSheetBehavior` that can be turned on to enable this new dismiss behavior.

I modified the `BottomSheetMainDemoFragment` in the Catalog app to enable this behavior, so that it can be observed in the demo app.

cketcham:
Thanks!

cketcham:
submitted this internally. It should be synced out to github soon. thanks!

lib/java/com/google/android/material/bottomsheet/BottomSheetDialogFragment.java
```diff
@@ -31,4 +39,94 @@
...
   }
+
+  @Override
+  public void dismiss() {
+    Dialog dialog = getDialog();
+    if (dialog != null) {
+      View bottomSheet = dialog.findViewById(R.id.design_bottom_sheet);
+      if (bottomSheet != null) {
+        BottomSheetBehavior<View> behavior = BottomSheetBehavior.from(bottomSheet);
+        if (behavior.dismissWithAnimation) {
```
froriz5: @cketcham Commenting here for visibility.
I was testing some stuff locally, and it looks like if the behavior has
`isHidable = false` and we use the `BottomSheetBehavior.STATE_HIDDEN`, an exception is thrown here:

https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/bottomsheet/BottomSheetBehavior.java#L1127

I think we should add one my check in this `if` statement:

```
if (behavior.isHidable() && behavior.dismissWithAnimation) { 
  dismissWithAnimation(behavior)
} else {
  super.dismiss()
}

PR #453, opened 2019-07-19
[MaterialButtonToggleGroup] Fix documentation for styling instructions.

StephaneBg:
Documentation stated to use materialButtonOutlinedStyle attribute for child buttons. But this attribute is private so a lint warning showed up. Instead public style should be advised.
This PR corrects the documentation to use Widget.MaterialComponents.Button.OutlinedButton style.


StephaneBg:
I signed it!

ymarian:
Can you make the attribute public instead?

dsn5ft:
Thanks for opening this request, we went with adding the `materialButtonOutlinedStyle` attr to the public API (52c1a4d).

PR #524, opened 2019-08-18
[TextInputLayout] Public method for borderStroke to be able to change colors dynamically

krossovochkin:
Currently there are two ways to change borderStroke colors;
- via xml (initial object creation)
- by calling setBorderStrokeColor

The issue with first one is that it is not possible to change color programmatically.
The issue with second is that it changes only *focusedColor* (despite API method name and some doc).

Proposed is the public method which updates all stroke-colors (similar logic to what is in inflater).

Basically, proposed method should've been primary method of the public API (instead of setBoxStrokeColor).
But to keep backward compatibility, let's at least have additional method to override any state of the stroke color.

Thank you

### DIFF
--- a/lib/java/com/google/android/material/textfield/TextInputLayout.java
+++ b/lib/java/com/google/android/material/textfield/TextInputLayout.java
@@ -854,7 +854,7 @@ public class TextInputLayout extends LinearLayout {
   /**
    * Set the box's stroke color state list.
    *
-   * @param boStrokeColorStateList the color state list to use for the box's stroke
+   * @param boxStrokeColorStateList the color state list to use for the box's stroke
    */
   public void setBoxStrokeColorStateList(ColorStateList boxStrokeColorStateList) {
     if (boxStrokeColorStateList != null && boxStrokeColorStateList.isStateful()) {
--- a/lib/java/com/google/android/material/textfield/TextInputLayout.java
+++ b/lib/java/com/google/android/material/textfield/TextInputLayout.java
@@ -374,7 +374,7 @@ public class TextInputLayout extends LinearLayout {
   @ColorInt private final int disabledFilledBackgroundColor;
   @ColorInt private final int hoveredFilledBackgroundColor;
 
-  @ColorInt private final int disabledColor;
+  @ColorInt private int disabledColor;
 
   // Only used for testing
   private boolean hintExpanded;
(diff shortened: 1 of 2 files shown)


krossovochkin:
@googlebot I signed it!

krossovochkin:
I guess build failed not because of the changes but maybe because of the infrastructure issues?
```
The command "yes | sdkmanager --update" failed and exited with 1 during .
```

Should I create separately issue for this to get more attention?

Thank you!

lib/java/com/google/android/material/textfield/TextInputLayout.java
```diff
@@ -851,6 +851,31 @@ public int getBoxStrokeColor() {
     return focusedStrokeColor;
   }
 
+  /**
+   * Set the box's stroke color state list.
+   *
+   * @param boStrokeColorStateList the color state list to use for the box's stroke
```
rachitmishra: `boxStrokeColorStateList`

krossovochkin: Thank you, fixed!

ldjcmu:
Hi Krossovochkin,

First off thanks for the PR and sorry for the very delayed response. I wanted to clarify some process items and talk about the PR.

- Because of how we store our code internal to Google, we don't merge PRs directly in github (hence the 0 successful merges except rollbacks). Instead, we create an internal copy of the PR referencing the external author, issue number, and pull number which then gets sync'd out. (Commit 11ec97a846fba69b6b2c9e89dee42e2b5a66b358 is an example of this). So while it doesn't look like PRs are merged, we do watch this and "accept" PRs with a slightly indirect process.

- We fell behind on our PR management process which is why this is two weeks delayed in response. Since you submitted this, we have fixed our CI builds (as you noted) and should be in a much better place to monitor and accept future PRs. I'm sorry for the delay in this occurrence and hope it won't dissuade you from continuing to provide feedback and PRs for the library.

- With regards to this PR, having looked at our TextInputLayout color implementation, there's a couple things we want to change (as you noted, it's confusing). We've slated these larger color changes for the next month and are tracking them with issue #593. I'm going to close this PR as we want to make a larger set of changes that address this issue in a different way.

krossovochkin:
Hi @ldjcmu 
Thank you for your response.
Yes it seems good idea to revisit stroke colors, looking forward to see updates in library.

Regarding processes of PRs – now it's more clear. But probably would be good to see this in contributing doc so everyone aware beforehand.

Thank you!

ldjcmu:
Thanks for the feedback. We're updating the contributing guidance to explain the process.

PR #532, opened 2019-08-23
[BottomNavigationView] Fixed method call in the documentation

svantepolk:
I noticed that the method referenced in the documentation doesn't actually exist. It seems like it should be getOrCreateBadge instead of showBadge.


svantepolk:
Just realised that the method name changed in later versions.

PaulWoitaschek:
Why did you close this?

PR #574, opened 2019-09-07
[MaterialCardView] Updated documentation with the reference of the Androidx CardView 

gabrielemariotti:
The components extends the Androidx-Cardview instead of Support Library. Updated the links to the right class.


ldjcmu:
This should sync in today. Thanks for the helpful fix and sorry for the delay.

PR #579, opened 2019-09-09
Update AndroidX and Kotlin dependencies

MGaetan89:
This PR updates the following dependencies:

|Dependency|Previous version|New version|
|-----|-----|-----|
|`Kotlin`|1.3.21|1.3.50|
|`annotation`|1.0.1|1.1.0|
|`appCompat`|1.1.0-rc01|1.1.0|
|`core`|1.1.0-rc03|1.1.0|
|`fragment`|1.0.0|1.1.0|
|`lifecycle`|2.0.0|2.1.0|
|`transition`|1.0.1|1.1.0|
|`vectorDrawable`|1.1.0-rc01|1.1.0|
|`viewpager2`|1.0.0-beta03|1.0.0-beta04|

Now only two dependencies are used in a non-stable version: `coordinatorlayout` (1.1.0-beta01) and `viewpager2` (1.0.0-beta04).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 buildscript {
-  ext.kotlinVersion = '1.3.21'
+  ext.kotlinVersion = '1.3.50'
   repositories {
     google()
     jcenter()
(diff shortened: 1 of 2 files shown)


ldjcmu:
Hi MGaetan89,

I checked internally and we have a policy of being on the earliest stable version that works (assuming we're not using APIs that aren't present in later versions). The idea is that we don't want to force version updates if we don't have to on dependent projects, and the dependent projects can still use later versions if they so choose. As such, I'm about to sync out a change that updates just the non-stable versions, but leaves the others unchanged.

PR #644, opened 2019-10-01
fix incorrect imports

johan12345:
The build is failing on the current `master`. I needed to replace these `android.support.v7` imports with the corresponding `androidx.appcompat` versions to get it to work.


ldjcmu:
I think I pushed a fix for this about an hour ago. It's buried in https://github.com/material-components/material-components-android/commit/b99709200e42366bec214c1e862f5f69ba22d516

PR #649, opened 2019-10-03
[AppBarLayout] Added format to appBarLayoutStyle attribute

seanghay:
Specify the format to `appBarLayoutStyle` attribute, so Android Studio knows what to suggest.


seanghay:
> @googlebot I signed it!

Issue #662, opened 2019-10-08
[NavigationView] Separators should inherit size of ?android:listDivider


tadfisher:
**Is your feature request related to a problem? Please describe.**

The separator view in `@layout/design_navigation_item_separator` is hardcoded to a height of 1dp. Our app uses a different size for the `?android:listDivider` drawable, so our navigation drawer is inconsistent with the rest of our app.

**Describe the solution you'd like**

I would like the separators in `NavigationView` to be consistent with other usages of separators in our app; for example, RecyclerView's `DividerItemDecoration` uses the intrinsic size of the `?android:listDivider` drawable by default.

**Describe alternatives you've considered**

I've considered forking Material Components for Android and changing the height of the separator view to `wrap_content`, as `NavigationMenuPresenter` is hidden and restricted.

ymarian:
see https://github.com/material-components/material-components-android/pull/663
For the suggested alternative

PR #663, opened 2019-10-08
[NavigationView] Set separator height to wrap_content

tadfisher:
Fixes: #662 

As far as I'm aware, the divider drawables set by the platform themes have always specified both a width and height, so this should be safe unless the user has set `?android:listDivider` to a drawable with no intrinsic height.

### DIFF
--- a/lib/java/com/google/android/material/internal/res/layout/design_navigation_item_separator.xml
+++ b/lib/java/com/google/android/material/internal/res/layout/design_navigation_item_separator.xml
@@ -20,7 +20,7 @@
 
   <View
       android:layout_width="match_parent"
-      android:layout_height="1dp"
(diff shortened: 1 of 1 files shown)


ymarian:
Does this solve https://github.com/material-components/material-components-android/issues/662?

ymarian:
We discussed this and we think just having a dimen resource that you can override seems a better approach here. So define:

`<dimen name="design_navigation_item_separator_height>1dp</dimen> `

Then your app can redefine this resource as:

`<dimen name="design_navigation_item_separator_height>-2</dimen> `

or whatever suits your usecase better.

ymarian:
Since there isn't activity here I'm just closing this.

The suggested workaround would also work by redeclaring the resource as

Issue #708, opened 2019-10-24
[MaterialAlertDialog] Text LineHeight isn't applied


kingargyle:
**Description:**

When trying to set the text line height to a MaterialAlertDialog it appears to be ignored.  This appears to be because it is just extending the AlertDialog, and not using the the new MaterialTextView that correctly applies the lineheight.

**Expected behavior:** 
The specified line height from the text appearance style should be applied.

**Source code:** 
`
    <style name="TextAppearance.MyTheme.Body2" parent="TextAppearance.MaterialComponents.Body2">
        <item name="fontFamily">@font/tradegothic_lt_std_family</item>
        <item name="android:textSize">14sp</item>
        <item name="android:letterSpacing">0.013</item>
        <item name="android:lineHeight">22dp</item>
        <item name="lineHeight">22dp</item>
    </style>
`

When the above is assigned to ?textAppearanceBody2 attribute, the lineheight should be applied, currently it is not on APIs lower than 28.

**Android API version:** Android API version here

API 26

**Material Library version:** Material Android Library version you are using here (e.g., 1.1.0-alpha07)

1.1.0-Beta01

**Device:** Device on which the bug was encountered here

Google Pixel running Oreo 8.1

forntoh:
I am facing the same issue

Issue #684, opened 2019-10-16
[Component CardView Sample] LineHeight isn't adjusting


kingargyle:
**Description:** 
When making a modification to the line height, it isn't being reflected in the Card components texts.   This is happening because the TextViews in the component_cardview.xml are using the AppCompatTextView instead of the MaterialTextView.   Once I change the component_cardview.xml to use the MaterialTextView everything starts to display as expected.

**Expected behavior:** 
LineHeight adjustments should reflect in any of the textviews.

**Android API version:**
28 (or any of the supported versions)

**Material Library version:** 
1.1.0-beta01

**Device:** 
Samsung S9

If I get some time I'll try and get a pull request to fix the issue.

ikim24:
Fixed in https://github.com/material-components/material-components-android/commit/ce83a9e4dd91c85f334036808cccbb411e9ca38d

PR #709, opened 2019-10-24
[MaterialAlertDialog][MaterialThemeBuilder][Libs] Update from TextView to MaterialTextView

kingargyle:
This pull request addresses issues #708 and #684.

This addresses a number of components that are just making use of TextView as well.

There are several places throughout the project that are still using the
TextView or the AppCompatTextView.  This causes issues with lineHeight not
being respected by multiple components in the project on devices that are
lower than API 28.

The changes here update all the components that are referencing the TextView or
the AppCompatTextView to use the MaterialTextView allow the lineheight to be
adjusted on all API versions supported by the library.

#### Don't forget:

### DIFF
--- a/lib/java/com/google/android/material/bottomnavigation/res/layout/design_bottom_navigation_item.xml
+++ b/lib/java/com/google/android/material/bottomnavigation/res/layout/design_bottom_navigation_item.xml
@@ -32,7 +32,7 @@
       android:clipChildren="false"
       android:clipToPadding="false"
       android:duplicateParentState="true">
-    <com.google.android.material.textview.MaterialTextView
+    <TextView
         android:id="@+id/smallLabel"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
@@ -40,7 +40,7 @@
         android:ellipsize="end"
         android:maxLines="1"
         android:textSize="@dimen/design_bottom_navigation_text_size"/>
-    <com.google.android.material.textview.MaterialTextView
+    <TextView
         android:id="@+id/largeLabel"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
--- a/lib/java/com/google/android/material/bottomnavigation/res/layout/design_bottom_navigation_item.xml
+++ b/lib/java/com/google/android/material/bottomnavigation/res/layout/design_bottom_navigation_item.xml
@@ -32,7 +32,7 @@
       android:clipChildren="false"
       android:clipToPadding="false"
       android:duplicateParentState="true">
-    <TextView
+    <com.google.android.material.textview.MaterialTextView
         android:id="@+id/smallLabel"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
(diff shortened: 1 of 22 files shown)


dsn5ft:
This actually shouldn't be needed because we automatically inflate `<TextView`s into `<com.google.android.material.textview.MaterialTextView`s:

https://github.com/material-components/material-components-android/blob/master/docs/components/MaterialTextView.md#usage

https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/theme/MaterialComponentsViewInflater.java#L109-L113

If the lineHeights really aren't being applied in MaterialThemeBuilder, we should figure why the auto-inflation or something else about MaterialTextView isn't working.

kingargyle:
> If the lineHeights really aren't being applied in MaterialThemeBuilder, we should figure why the auto-inflation or something else about MaterialTextView isn't working.

The line heights are definitely not adjusting.  The easiest way to see at least with 1.1.0-beta01 is to apply style like the following to a textview:

```
    <style name="TextAppearance.AbercrombieAppTheme.Body2" parent="TextAppearance.MaterialComponents.Body2">
        <item name="fontFamily">@font/source_sans_pro_regular_family</item>
        <item name="android:textSize">14sp</item>
        <item name="android:letterSpacing">0.013</item>
        <item name="android:lineHeight">50dp</item>
        <item name="lineHeight">50dp</item>
    </style>
```

View it on a device with below API 28, and one with 28 or higher.   The one on the lower device, will not have the large lineheight applied to it.  The one with API 28 will.

dsn5ft:
Gotcha, there may be an issue with how we're reading `android:lineHeight` vs `lineHeight` then.

As far as the behavior you're seeing, does updating the MaterialThemeBuilder's `<TextViews` to `<com.google.android.material.textview.MaterialTextView` actually fix that issue for pre-28?

dsn5ft:
Follow-up question: if you just set `lineHeight` instead of also setting `android: lineHeight` does that fix the issue?

kingargyle:
> Gotcha, there may be an issue with how we're reading `android:lineHeight` vs `lineHeight` then.
> 
> As far as the behavior you're seeing, does updating the MaterialThemeBuilder's `<TextViews` to `<com.google.android.material.textview.MaterialTextView` actually fix that issue for pre-28?

Yes, if we change to MaterialTextView it fixes the issue on pre-28 devices, as it gets the app:lineHeight correctly.

I'll try and test with remove the android:lineHeight and just leaving lineHeight defined and see how it works.  Probably won't get to it until tomorrow.

dsn5ft:
Actually, taking a closer look (thanks @hunterstich!), you should be able to revert all the changes under `lib/` and for `material-theme-builder/` you can change the `<androidx.appcompat.widget.AppCompatTextView`s to `<TextView`. That should fix the auto-inflation to `MaterialTextView`s, and it's the preferred way to use the component.

kingargyle:
Thanks I'll submit an updated pr tomorrow

kingargyle:
@dsn5ft sorry for the delay, I've reverted the other changes, and update the layouts that were using AppCompatTextView directly to use TextView.

dsn5ft:
No worries! This looks good so we'll merge it in soon.

PR #724, opened 2019-10-28
[TextInputLayout] Add support for focused background color

### LINKED ISSUE #710: [TextInputLayout] Support boxBackgroundColor focused state
## Problem Description
You cannot set a different filled `boxBackgroundColor` for when the `TextInputLayout` or the child `TextInputEditText` has focus.

## Proposed Solution Description
We should be able to set a color resource in a Color State List that will change the background color of the `TextInputLayout` when it or its child `TextInputEditText` has focus.  For example:

```xml
<selector xmlns:android="http://schemas.android.com/apk/res/android">

    <!-- This is what I would like to do, but it currently does nothing: -->
    <item android:alpha="0.25" android:color="?colorOnSurface" android:state_focused="true" />

    <!-- The following three items are the only states that TextInputLayout respects for boxBackgroundColor: -->
    <item android:alpha="0.16" android:color="?colorOnSurface" android:state_hovered="true" />
    <item android:alpha="0.04" android:color="?colorOnSurface" android:state_enabled="false" />
    <item android:alpha="0.12" android:color="?colorOnSurface" />
</selector>
```

## Possible Alternative Solution
A programmatic approach could add a focus changed listener to each `TextInputLayout` and override the `boxBackgroundColor` when the view or its child gets focus.  This seems really hacky though.

## Additional Context
The default Material Text Field component example (https://material.io/components/text-fields/) shows a darker background color than the hovered color when the text field is focused.  It seems like this feature is simply missing from the Android implementation.

nkotula:
Closes #710 

This enables developers to define a color resource to be used when a
TextInputLayout or its child TextInputEditText have focus via a Color
State List.

### DIFF
--- a/lib/java/com/google/android/material/textfield/res/color/mtrl_filled_background_color.xml
+++ b/lib/java/com/google/android/material/textfield/res/color/mtrl_filled_background_color.xml
@@ -16,6 +16,7 @@
 -->
 <selector xmlns:android="http://schemas.android.com/apk/res/android">
   <item android:alpha="0.16" android:color="?attr/colorOnSurface" android:state_hovered="true"/>
+  <item android:alpha="0.12" android:color="?attr/colorOnSurface" android:state_focused="true"/>
   <item android:alpha="0.04" android:color="?attr/colorOnSurface" android:state_enabled="false"/>
   <item android:alpha="0.12" android:color="?attr/colorOnSurface"/>
 </selector>
--- a/build.gradle
+++ b/build.gradle
@@ -56,7 +56,7 @@ ext {
     ? project.property('mavenRepoUrl') : 'file://localhost/tmp/myRepo/')
 
   // Current version of the library (could be in-development/unreleased).
-  mdcLibraryVersion = '1.1.0-alpha10'
+  mdcLibraryVersion = '1.2.0-alpha01'
   mdcLibraryPackage = "com.google.android.material"
   mdcLibraryDir = "com/google/android/material"
 }
--- a/lib/java/com/google/android/material/badge/BadgeDrawable.java
+++ b/lib/java/com/google/android/material/badge/BadgeDrawable.java
@@ -35,6 +35,7 @@ import android.os.Parcel;
 import android.os.Parcelable;
 import androidx.annotation.AttrRes;
 import androidx.annotation.ColorInt;
+import androidx.annotation.Dimension;
 import androidx.annotation.IntDef;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
@@ -187,6 +188,12 @@ public class BadgeDrawable extends Drawable implements TextDrawableDelegate {
     @PluralsRes private int contentDescriptionQuantityStrings;
     @BadgeGravity private int badgeGravity;
 
+    @Dimension(unit = Dimension.PX)
+    private int horizontalOffset;
+
+    @Dimension(unit = Dimension.PX)
+    private int verticalOffset;
+
     public SavedState(@NonNull Context context) {
       // If the badge text color attribute was not explicitly set, use the text color specified in
       // the TextAppearance.
@@ -207,6 +214,8 @@ public class BadgeDrawable extends Drawable implements TextDrawableDelegate {
       contentDescriptionNumberless = in.readString();
       contentDescriptionQuantityStrings = in.readInt();
       badgeGravity = in.readInt();
+      horizontalOffset = in.readInt();
+      verticalOffset = in.readInt();
     }
 
     public static final Creator<SavedState> CREATOR =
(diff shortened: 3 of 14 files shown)


nkotula:
Apparently I jumped the gun on opening the PR.  I didn't realize that Corporate CLAs had to be individually reviewed, so the CLA won't be verified for a bit.

lib/java/com/google/android/material/textfield/res/color/mtrl_filled_background_color.xml
```diff
@@ -16,6 +16,7 @@
 -->
 <selector xmlns:android="http://schemas.android.com/apk/res/android">
   <item android:alpha="0.16" android:color="?attr/colorOnSurface" android:state_hovered="true"/>
+  <item android:alpha="0.12" android:color="?attr/colorOnSurface" android:state_focused="true"/>
```
nkotula: I left the opacity the same as the default case just to keep the visuals consistent with current source.  Note that this bug (https://issuetracker.google.com/issues/68029303) means that I had to put this above the `state_enabled="false"` item.

nkotula:
@googlebot I signed it!

nkotula:
> focusFilledBackgroundColor should be set to defaultFilledBackgroundColor in line 524

Can I get some clarification on this?  Does that mean this:

```java
focusFilledBackgroundColor =
    filledBackgroundColorStateList.getColorForState(
        new int[] {android.R.attr.state_focused}, -1);
```

should instead be:

```java
focusFilledBackgroundColor = defaultFilledBackgroundColor
```
?

leticiarossi:
exactly

nkotula:
Hrm... Doesn't that defeat the purpose of checking for a user-provided `state_focused` Color State in something like a `materialThemeOverlay`?  The `focusedFilledBackgroundColor` doesn't get set anywhere else except the constructor, so if we always set it to the default color then it won't get the correct BG color when focused.

leticiarossi:
My bad, I got the wrong line number. I meant to say where it's set in the else block after the `if (filledBackgroundColorStateList.isStateful()) {`

lib/java/com/google/android/material/textfield/res/color/mtrl_filled_background_color.xml
hurenakg: Is this change already released?

nkotula: If you check the full changelists under Releases, it looks like `1.1.0-beta02` didn't get it, but `1.2.0-alpha01` has it.

hurenakg: Thanks. It would be great if we can set another state colors, like custom states or error state.

hurenakg: I mean, for filledBox background color.

GregorySokoloff:
It would be very helpful if you could provide an example of how to use this great new feature. Specifically, how to override the default focused background color with an entry in a custom app theme. Or how to set the color programmatically. Thanks for any help.

nkotula:
@GregorySokoloff This PR just allows you to include an `android:state_focused="true"` to your `TextInputLayout` color selector XML.  Here's an example:

```xml
<!-- my_text_input_layout_background_color.xml -->

<selector xmlns:android="http://schemas.android.com/apk/res/android">

    <!-- This is the feature that was added with this PR.  You can customize the color and alpha as much as you want: -->
    <item android:alpha="0.25" android:color="?colorOnSurface" android:state_focused="true" />
    
    <!-- The following three items were the only states that TextInputLayout respected for boxBackgroundColor before this PR went through: -->
    <item android:alpha="0.16" android:color="?colorOnSurface" android:state_hovered="true" />
    <item android:alpha="0.04" android:color="?colorOnSurface" android:state_enabled="false" />
    <item android:alpha="0.12" android:color="?colorOnSurface" />
</selector>
```

Then, whenever a `TextInputLayout` or its child `TextInputEditText` get focus, it will use the focused color defined in your selector.

GregorySokoloff:
@nkotula  Thanks very much for kindly providing the example.

I guess because there's no default theme attribute for this selector, there's no way to override it by using a theme.

GregorySokoloff:
@nkotula Another obtuse question: is this feature in any of the releases? (Obviously, I'm no expert, but I could not find it in any of the release notes since you committed it.)

nkotula:
@GregorySokoloff  It is not included in any of the `1.1.0-Beta` or `1.1.0-RC` releases, no.  I'm not sure why, but that's not up to me.

GregorySokoloff:
@nkotula Thanks again for responding. I'll wait to see if it makes it into the next alpha build. Cheers, Greg

PR #737, opened 2019-11-06
[WIP][TextInputLayout] helperText disappear after calling setError(null)

### LINKED ISSUE #525: [TextInputLayout] Hiding already hidden error causes helper text to be hidden
**Description:** Calling `TextInputLayout.setError(null)` without checking `TextInputLayout.getError()` for **null** will cause the helper text to show only for a brief amount of time (< 1s) before disappearing again.

**Expected behavior:** Calling `TextInputLayout.setError(null)` consecutively is expected to only hide the error only and not the helper text as well.

**Source code:**
```java
final TextInputLayout inputLayout = findViewById(R.id.input_email);
inputLayout.setHelperText("Email");
inputLayout.setError("Email required");
inputLayout.setError(null);
inputLayout.setError(null);
```

**Android API version:** 28

**Material Library version:** 1.1.0-alpha09

**Device:** Asus Zenfone Max Pro (M1)

matpag:
This should fix the `setError(null)` bug where the helperText disappears.

The PR is in WIP because even if it solves the main problem maybe does not cover all the cases.

For example if I do something like this: 
```
inputLayout.setHelperTextEnabled(false);
inputLayout.setError("byebye");
```
The `errorView` is not showing the error text (but showing the error icon drawable) only the first time the above methods are called, if I call them again it works fine. 
I need to check if this is another separate bug or is related to my changes.

Closes #525

### DIFF
--- a/lib/java/com/google/android/material/textfield/IndicatorViewController.java
+++ b/lib/java/com/google/android/material/textfield/IndicatorViewController.java
@@ -141,6 +141,10 @@ final class IndicatorViewController {
     if (captionDisplayed == CAPTION_STATE_HELPER_TEXT) {
       captionToShow = CAPTION_STATE_NONE;
     }
+    if (captionToShow == captionDisplayed) {
+      // Do not proceed switching caption texts because there is no need
+      return;
+    }
     updateCaptionViewsVisibility(
         captionDisplayed, captionToShow, shouldAnimateCaptionView(helperTextView, null));
   }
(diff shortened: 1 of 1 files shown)


leticiarossi:
Thanks for the PR! I submitted what I already had in the works.

emrsn-gabrieltrifa:
I still have the same issue..

matpag:
@emrsn-gabrieltrifa Are you using the latest alpha version of the material library?

emrsn-gabrieltrifa:
I'm using implementation 'com.google.android.material:material:1.0.0'

matpag:
This is not fixed in 1.0.0 branch, if you want this fixed you should use the alpha version (or stay with the workaround waiting for the 1.2.0 release)

emrsn-gabrieltrifa:
Interesting, I thought 1.0.0 was the latest one. Android Studio does not show me any update available, and they normally highlight it in the gradle file when there are new version.

Thanks.

PR #753, opened 2019-11-10
[dialog] Fix multichoice and singlechoice alert dialog in rtl mode of old api

sadeghpro:
Hi
Single and multi choice alert dialog have some issue with rtl language like persian and arabic on api 16,17,18 (I test this three version)
api 16 not support rtl then drawable always in left but if I use persian work this happend:
screenshot-2019-11-10_21 35 10 649
api 18 support rtl then drawable on change to right when set locale to "fa" but if use persian and English word together this happend:
screenshot-2019-11-10_21 30 33 637

 on api 17 there is a strange bug. the same bug like above exist and there are two drawable in both right and left when locale set to "fa":
screenshot-2019-11-10_21 41 54 908

then I change xml layout of single and multi choice alert dialog to fix this problem, also separate layout of api 17 and above from default layout
after this commit ui on api 17 :
screenshot-2019-11-10_21 44 46 97

I also check this code on api 16,17,18,23 on both ltr and rtl language and every thing is normal

### DIFF
--- a/lib/java/com/google/android/material/dialog/res/layout/mtrl_alert_select_dialog_multichoice.xml
+++ b/lib/java/com/google/android/material/dialog/res/layout/mtrl_alert_select_dialog_multichoice.xml
@@ -15,17 +15,31 @@
      limitations under the License.
 -->
 
+<!--
+    This layout file is used by the AlertDialog when displaying a multi select list of items.
+
+    A left android:gravity will handle text alignemnt on API 16 and below while
+    android:textAlignemnt will be used on 17+. A start android:gravity is not added
+    to avoid text being aligned right while a drawable is aligned left in RTL configurations
+    on API 16 and below.
+
+    app:drawableLeftCompat (and app:drawableStartCompat) used instead of android:drawableLeft
+    (and android:drawableStart) because of bug on api 17 that aligned drawable both left and
+    right when in RTL configurations
+-->
 <CheckedTextView xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
     android:id="@android:id/text1"
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
     android:minHeight="?attr/listPreferredItemHeightSmall"
-    android:gravity="center_vertical"
+    android:gravity="left|center_vertical"
+    android:textAlignment="viewStart"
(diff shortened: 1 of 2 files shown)


sadeghpro:
@googlebot I signed it!

lib/java/com/google/android/material/dialog/res/layout/mtrl_alert_select_dialog_multichoice.xml
```diff
@@ -20,12 +20,9 @@
...
-    android:gravity="center_vertical"
+    android:gravity="left|center_vertical"
     android:paddingLeft="@dimen/abc_select_dialog_padding_start_material"
     android:paddingRight="?attr/dialogPreferredPadding"
-    android:paddingStart="@dimen/abc_select_dialog_padding_start_material"
-    android:paddingEnd="?attr/dialogPreferredPadding"
     android:drawableLeft="?android:attr/listChoiceIndicatorMultiple"
-    android:drawableStart="?android:attr/listChoiceIndicatorMultiple"
     android:drawablePadding="20dp"
     android:ellipsize="marquee" />
```
hunterstich: I've been looking into this to see if there's a way to keep one, non-qualified, resource file instead of splitting out a v17 specific layout. It looks like you can fix the issue this PR addresses by changing the original multichoice and singlechoice layouts to:

```
  <!--  A left android:gravity will handle text alignemnt on API 16 and below while-->
  <!--  android:textAlignemnt will be used on 17+. A start android:gravity is not added -->
  <!--  to avoid text being aligned right while a drawable is aligned left in RTL configurations -->
  <!--  on API 16 and below.-->

  <CheckedTextView
     ...
     android:gravity="left|center_vertical" 
     android:textAlignment="viewStart"
     ...
     app:drawableLeftCompat="?android:attr/listChoiceIndicator(Single|Multiple)"
     app:drawableStartCompat="?android:attr/listChoiceIndicator(Single|Multiple)"
     .../>
```

As far as I've tested, this produces the same result as the changes you've provided for API 16, 17 and 17+. It would be great to avoid splitting out the layouts if possible!

sadeghpro:
oh your code is so better than mine. I test it on api 16, 17, 18, 19 , 23, 26, 28 all it's ok.
I resend new request

sadeghpro:
@hunterstich I add some comment.

hunterstich:
Perfect. Thanks! I'll work on merging this in

PR #783, opened 2019-11-20
[FloatingActionButton] Add support for enabled/disabled state

ricknout:
This adds missing support for enabled/disabled state to `FloatingActionButton`:
- Adds the `android:enabled` attribute to the FAB styleable and parses it in the View. This allows FABs to be enabled/disabled in XML. Previously, this was only possible with the `setEnabled(boolean)` method (although the tint colors were wrong).
- Adds color selectors for FAB background and icon tints that have the same state values as that of `ExtendedFloatingActionButton`.
- References the aforementioned selectors in the `Widget.MaterialComponents.FloatingActionButton` style.

Sample comparison of using `android:enabled` or `setEnabled(boolean)` set to false using the MDC-Android FAB catalog:

| Before | After |
| ----- | ----- |
| Screenshot_20191120-230205 | Screenshot_20191120-230407 |


lib/java/com/google/android/material/floatingactionbutton/FloatingActionButton.java
```diff
@@ -246,6 +246,8 @@ public FloatingActionButton(
     boolean ensureMinTouchTargetSize = a
         .getBoolean(R.styleable.FloatingActionButton_ensureMinTouchTargetSize, false);
 
+    setEnabled(a.getBoolean(R.styleable.FloatingActionButton_android_enabled, true));
```
cketcham: Is this required? I think the super class would handle setting the enabled state based on the `android:enabled` attribute.

lib/java/com/google/android/material/floatingactionbutton/res/color/mtrl_fab_bg_color_selector.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
```
cketcham: Since these are the same as EFAB and I expect them to behave the same way, maybe both classes should use the same selectors?

ricknout: I was thinking the same. For the naming - would it be fine if the color resources used the `mtrl_fab` prefix for both FAB and EFAB?

lib/java/com/google/android/material/floatingactionbutton/FloatingActionButton.java
ricknout: It doesn't seem to. Tracing it back: `FloatingActionButton` <- `VisibilityAwareImageButton` <- `ImageButton` <- `ImageView` <- `View` - the source suggests none of these classes parse/handle the `android:enabled` attribute

lib/java/com/google/android/material/floatingactionbutton/res/color/mtrl_fab_bg_color_selector.xml
ricknout: Done @cketcham ✅

lib/java/com/google/android/material/floatingactionbutton/FloatingActionButton.java
cketcham: Yep that's right!

PR #790, opened 2019-11-26
[CardView] Fix potential Proguard/Dexguard/R8(?) optimization problem

rossbacher:
If using Proguard/Dexguard optimization and enabling all optimizations (e.g. not using default `getDefaultProguardFile('proguard-android-optimize.txt')` it would optimize `isParentCardViewDoneInitializing` away as it is final and thus can be inlined (as `true` because that i what it is set to in the initializer).
However this code assumes that it is `false` during the `super(...)` call and thus using Progurad/Dexguard optimization on this will cause an NPE in `setBackgroundDrawable(...)`.
One could argue that this is a bug in Proguard/Dexguard but it is also not a good design to use a final field and rely on it having different values during runtime.

I did not test this with R8 but did so with Proguard and Dexguard

#### Don't forget:


rossbacher:
@googlebot I signed it!

PR #796, opened 2019-11-27
[MaterialButtonToggleGroup] Clean up

zhaonian:
What I did:
 - minor IDE warnings
 - Java doc typo
 - renamed `MaterialButtonToggleGroupUnitTest` to `MaterialButtonToggleGroupTest` to be consistent with other tests, and merged two `@Before` into one and added doc.

TESTED=builds and runs as expected & unit test passed

### DIFF
--- a/lib/javatests/com/google/android/material/button/MaterialButtonToggleGroupTest.java
+++ b/lib/javatests/com/google/android/material/button/MaterialButtonToggleGroupTest.java
@@ -35,12 +35,10 @@ import org.junit.runner.RunWith;
 import org.robolectric.RobolectricTestRunner;
 import org.robolectric.annotation.Config;
 
-/**
- * Tests for {@link com.google.android.material.button.MaterialButtonToggleGroup}.
- */
+/** Tests for {@link com.google.android.material.button.MaterialButtonToggleGroup}. */
 @RunWith(RobolectricTestRunner.class)
 @Config(sdk = 21)
-public class MaterialButtonToggleGroupUnitTest {
+public class MaterialButtonToggleGroupTest {
 
   private static final float CORNER_SIZE = 10f;
   private final Context context = ApplicationProvider.getApplicationContext();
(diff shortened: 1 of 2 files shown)


catalog/java/io/material/catalog/button/res/layout/cat_buttons_toggle_group_fragment.xml
```diff
@@ -101,35 +101,35 @@
         android:id="@+id/icon_only_group"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content">
-      <Button
```
wcshi: Why explicitly hardcode these? These should autoinflate to MaterialButton via MaterialComponentsViewInflater?

zhaonian: oh okk. Rreverted.

lib/javatests/com/google/android/material/button/MaterialButtonToggleGroupTest.java
```diff
@@ -35,31 +35,26 @@
...
+public class MaterialButtonToggleGroupTest {
 
   private static final float CORNER_SIZE = 10f;
   private final Context context = ApplicationProvider.getApplicationContext();
 
   private MaterialButtonToggleGroup toggleGroup;
 
   @Before
-  public void themeApplicationContext() {
+  public void setup() {
```
wcshi: Going to reject this because our readability guidelines suggest splitting the work up as appropriate and giving the methods descriptive names rather than just setUp or tearDown, otherwise pulling in your PR. Thanks for contributing

zhaonian: oh interesting thx for the review! :)
I thought there's no way to guarantee the order of execution of multiple `@Before`s, and here don't we want to make sure `context`'s theme has been set before the buttons created?

wcshi: Ohh, good point, will annotate themeApplicationContext with @BeforeClass to guarantee it is called before the other method.

zhaonian: It looks like `@BeforeClass` requires the method to be `static`, however the `context` in our `themeApplicationContext()` is non-static.  :/

wcshi: You are right, I changed themeApplicationContext to be a private method called by the other @Before method.

zhaonian: yayy thanks!

Issue #81, opened 2018-05-01
Add new `List` component guidance


sfdexter:
https://material.io/go/design-lists

Nemant:
How's the progress on this going?

ldjcmu:
The Material Android team is starting to plan for this component for work to being in Q1. As lists are a fairly broad components we'd love for people to share (in this issue) specific use cases (design mocks are great) of how you hope to use the component. The use cases will also help us prioritize pieces of the spec.

patrickpilch:
For our project, we would leverage gestures, subheaders, and list controls.

osherifo:
you can dooo ittttt!!!!

wcshi:
@ethanhsuhsu  thanks for stopping by ADS and offering to contribute towards this feature request. Do I have permission to post the feedback you submitted here so we can start a discussion on this issue?

ethanhsuhsu:
@wcshi You do indeed have my permission... 

To start off the conversation I will share the current issue

Here is a round-up of everything you need to make yourself for a [simple list](https://material.io/components/lists/#types):

* A card item xml layout that takes up 100 lines (using ConstraintLayout) describing what each list item looks like
* A RecyclerView adapter where you rewrite all of the ViewHolder recycling logic (also 100 line file)
* And the list goes on and on for data changes, swipe animations, and item dragging

The material spec should be easily implementable and for those who do not prioritize UI, adding a list should not be hard. 
Here is what our list component could look like (only an example for thought) :

A method that initializes a recyclerview and returns a View object

MaterialList(context, Cards.one_line_fourth) { cardholder -> cardholder.apply {
picture = mydata[0].picture
title = mydata[0].name
image = checkimage}

Params: the activity/fragment context, an enum specifying which card layout to use (xml layouts implemented in component library code) 
Lambda: Binds data to each list item

Notice: Potentially we can reduce a 200 line boilerplate implementation to a few lines of code!

I would be interested in hearing other ideas and I think with the combined power of type, shape and color theming we can make Material Lists... easier. Thanks for listening to my request and I would be excited to help out as much as possible. I have experience in this area (aka. creating 200+ line RecyclerViews).

cketcham:
I'm not sure that we need to totally abstract `RecyclerView` away, maybe we could do that eventually if it ends up reducing a lot of boiler plate, but I do see how not having to write ViewHolders and Adapters could be nice. What if we create those for use in a RecyclerView to start?

[...]

PR #798, opened 2019-11-28
[Lists] Added adapters to existing Material List ViewHolders

ethanhsuhsu:
> For anyone new, Material Lists is a new component being developed based on requests in #81. Read #81 for more specifics and how this PR helps the general project. Thank you @cketcham and @wcshi for promoting the creation of this new component! 😊 

I added adapters to handle the existing ViewHolders ([Single](https://github.com/ethanhsuhsu/material-components-android/blob/list-component/lib/java/com/google/android/material/lists/oneline/SingleLineItemViewHolder.java), [Two](https://github.com/ethanhsuhsu/material-components-android/blob/list-component/lib/java/com/google/android/material/lists/twoline/TwoLineItemViewHolder.java) and [Three](https://github.com/ethanhsuhsu/material-components-android/blob/list-component/lib/java/com/google/android/material/lists/threeline/ThreeLineItemViewHolder.java) ListItemViewHolder) and use them in RecyclerViews. 

**Notable code patterns 📝 used in the adapters:**
- The three adapters extend [MaterialListAdapter](https://github.com/ethanhsuhsu/material-components-android/blob/list-component/lib/java/com/google/android/material/lists/MaterialListAdapter.java), which helps define behavior they need to implement. Right now it is small but will be more useful in the future when we want the adapters to do more stuff like handle data changes. 
- The three adapters use an interface callback ([MaterialListBinder](https://github.com/ethanhsuhsu/material-components-android/blob/list-component/lib/java/com/google/android/material/lists/MaterialListBinder.java)) to do binding work outside of the adapter itself. 

**Other general improvements :wrench: I made specific to adapters:**
- I removed the inheritance structure between the ViewHolders because as we add more list card layouts and holders, it will be hard to change fields and layouts if other holders depend on them. It is also hard to see  🔍 all of the actual fields of a ViewHolder, they should be easily visible 👀. 

_Currently the ThreeLineItemViewHolder extends and has the view fields of the TwoLineItemViewHolder and so on with the Two and Single LineItemViewHolders_
- To show an example of my adapters in action, I added a [demo in the catalog app 📱](https://bit.ly/material-list-adapter-demo) that allows you to switch between different card holders and adapters.


ethanhsuhsu:
@googlebot I fixed it.

ethanhsuhsu:
@cketcham I submitted a pull request (#798) that adds adapters (as promised). I would like you (or the team) to review my contribution when convenient, so that I can address comments in a timely manner. This project is soo exciting.

lib/java/com/google/android/material/lists/MaterialListAdapter.java
```diff
@@ -0,0 +1,22 @@
+package com.google.android.material.lists;
+
+import androidx.annotation.NonNull;
+import androidx.recyclerview.widget.RecyclerView;
+import androidx.recyclerview.widget.RecyclerView.ViewHolder;
+
+/** The parent adapter that defines behavior for specific adapters to define */
+public abstract class MaterialListAdapter extends RecyclerView.Adapter<ViewHolder> {
```
bohsen: I know the work on `MaterialListAdapter` is scheduled for later, but will `androidx.recyclerview.widget.ListAdapter` and `androidx.paging.PagedListAdapter` fit into this somehow? You don't need to answer this elaborate. A yes or no will suffice.

lib/java/com/google/android/material/lists/MaterialListBinder.java
```diff
@@ -0,0 +1,8 @@
+package com.google.android.material.lists;
+
+import androidx.recyclerview.widget.RecyclerView.ViewHolder;
+
+/** Binder similar to onBindViewHolder() but in interface callback format */
+public abstract class MaterialListBinder<T extends ViewHolder> {
+  public abstract void onBind(T viewHolder, int position);
+}
```
bohsen: Any reason for not including the type of data the `MaterialListBinder` binds to? Like this:

```java
public abstract class MaterialListBinder<T, VH extends RecyclerView.ViewHolder> {
  public abstract void onBind(@NonNull VH viewHolder, @NonNull T item);
}

Otherwise LGTM
```

lib/java/com/google/android/material/lists/oneline/SingleLineItemViewHolder.java
```diff
@@ -14,20 +14,21 @@
...
+
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.ImageView;
 import android.widget.TextView;
 
 /** A simple single line list item. */
-public class SingleLineItemViewHolder extends ViewHolder {
+public class SingleLineItemViewHolder extends RecyclerView.ViewHolder {
```
bohsen: 👍

lib/java/com/google/android/material/lists/threeline/ThreeLineItemAdapter.java
```diff
@@ -0,0 +1,32 @@
...
+
+import androidx.annotation.NonNull;
+import androidx.recyclerview.widget.RecyclerView.ViewHolder;
+
+import com.google.android.material.lists.MaterialListAdapter;
+import com.google.android.material.lists.MaterialListBinder;
+import com.google.android.material.lists.threeline.ThreeLineItemViewHolder;
+
+/** An Adapter that shows three line list items */
+public class ThreeLineItemAdapter extends MaterialListAdapter {
```
bohsen: 👍

lib/java/com/google/android/material/lists/threeline/ThreeLineItemViewHolder.java
```diff
@@ -14,24 +14,33 @@
...
+
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
+import android.widget.ImageView;
 import android.widget.TextView;
 
 /** A simple three line list item. */
-public class ThreeLineItemViewHolder extends TwoLineItemViewHolder {
+public class ThreeLineItemViewHolder extends RecyclerView.ViewHolder {
```
bohsen: 👍

lib/java/com/google/android/material/lists/twoline/TwoLineItemViewHolder.java
```diff
@@ -14,24 +14,32 @@
...
+
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
+import android.widget.ImageView;
 import android.widget.TextView;
 
 /** A simple two line list item. */
-public class TwoLineItemViewHolder extends SingleLineItemViewHolder {
+public class TwoLineItemViewHolder extends RecyclerView.ViewHolder {
```
bohsen: 👍

EdricChan03:
@bohsen Please consider merging your comments into one review.

lib/java/com/google/android/material/lists/oneline/SingleLineItemAdapter.java
```diff
@@ -0,0 +1,31 @@
...
+import android.view.ViewGroup;
+
+import androidx.annotation.NonNull;
+import androidx.recyclerview.widget.RecyclerView.ViewHolder;
+
+import com.google.android.material.lists.MaterialListAdapter;
+import com.google.android.material.lists.MaterialListBinder;
+
+/** An Adapter that shows single line list items */
+public class SingleLineItemAdapter extends MaterialListAdapter {
```
cketcham: What do you think about making `MaterialListAdapter` generic where you pass in the type of the view holder, rather than creating a new adapter for each type.

cketcham: Maybe this isn't necessary.. It would make `onCreateViewHolder()` more complicated.. you'd have to pass in some kind of factory or something to create instances of the correct type. This is probably just as simple.

lib/java/com/google/android/material/lists/MaterialListAdapter.java
ethanhsuhsu: We may support paged data types later on, but the answer is no. If you want to customize the data source/ adapter you currently have two options:
- Customize how you use MaterialListBinder
- Use the ViewHolders in your own adapters

lib/java/com/google/android/material/lists/MaterialListBinder.java
ethanhsuhsu: Glad you asked @bohsen ! The reason is because there is no standard data structure for lists (could be a list, linkedklist, array, map) and we want to let the user customize how they send data to the adapter. Even if we allowed iterable data structures the user would have little control.

lib/java/com/google/android/material/lists/MaterialListAdapter.java
```diff
@@ -0,0 +1,22 @@
...
+  public MaterialListAdapter(int itemCount) {
+    this.itemCount = itemCount;
+  }
+
+  @Override
+  public int getItemCount() {
+    return itemCount;
+  }
+
+  public abstract void onBindViewHolder(@NonNull ViewHolder holder, int position);
```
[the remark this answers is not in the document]
ethanhsuhsu: @EdricChan03 MaterialListAdapter is an abstract class that serves the purpose of making sure all features are implemented in child adapters. If the child adapters did not extend MaterialListAdapter, they would not have to define onBindViewHolder and could be incomplete. We want all MaterialListAdapters to have the same functionality and structure.

lib/java/com/google/android/material/lists/MaterialListAdapter.java
```diff
@@ -0,0 +1,22 @@
...
+  int itemCount;
+
+  public MaterialListAdapter(int itemCount) {
+    this.itemCount = itemCount;
+  }
+
+  @Override
+  public int getItemCount() {
+    return itemCount;
+  }
```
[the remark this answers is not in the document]
ethanhsuhsu: @EdricChan03 getItemCount is meant for the RecyclerView to call, and tells it how many items to eventually allocate for. I assume you are talking about supporting dynamic data, we currently do not support that and this PR is not meant for that.

catalog/java/io/material/catalog/lists/MaterialListAdapterDemoFragment.java
```diff
@@ -0,0 +1,117 @@
...
+      }),
+      new ThreeLineItemAdapter(1000, new MaterialListBinder<ThreeLineItemViewHolder>() {
+        @Override
+        public void onBind(ThreeLineItemViewHolder vh, int position) {
+          vh.primaryText.setText(R.string.mtrl_list_item_one_line);
+          vh.secondaryText.setText(R.string.mtrl_list_item_secondary_text);
+          vh.tertiaryText.setText(R.string.mtrl_list_item_tertiary_text);
+          vh.icon.setImageResource(R.drawable.logo_avatar_anonymous_40dp);
+        }
+      })};
```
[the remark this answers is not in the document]
ethanhsuhsu: The dataset is irrelevant, this demo is meant to show that the relationship between MaterialListBinder and MaterialListAdapter works

ethanhsuhsu:
@bohsen @cketcham @EdricChan03 I believe I addressed all of your issues in my comments. If you have any further comments be sure to comment back with `@ethan627hsu`. 

@EdricChan03 please email me at ethanhsuhsu@gmail.com so we can setup instant messaging to avoid cluttering the PR threads. 

I am happy to address "missing functionality" in new PRs but this is only meant to be a start and I would like to only fix critical issues and bugs. After this gets merged, you can raid ⚔️ #81 with all of your recommendations!

ethanhsuhsu:
@bohsen @cketcham @wcshi As I understand, all of the comments were largely concerns and clarifications over change requests. If this is incorrect and you have specific changes in mind not addressed feel free to respond below and I will be happy to address your changes in my code.

Otherwise... Ready for merging?

lib/java/com/google/android/material/lists/MaterialListBinder.java
cketcham: I think those are two different questions. @bohsen's is `Can we add the type of the item to the bind method in this interface?`. Yours is `Can we make an interface for MaterialListAdapter that doesn't restrict how the user sends data to the adapter?`. Right?

In theory we could make an interface that has a method to return a size, and another to return an object of a particular type given the index. We'd be able to use that in the current implementation rather than passing the size directly. This would allow us to add a type, but keep the current flexibility.

I actually think an interable is the most common case.. And it's pretty easy to convert some other data type to be interable to pass in, so I'm not sure how much of an advantage there is to keep it more open.

catalog/java/io/material/catalog/lists/MaterialListAdapterDemoFragment.java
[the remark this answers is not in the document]
cketcham: I like this demo because it shows the different list types that exist, so I think it's good to keep, but maybe you could make another demo that shows how someone might actually use `MaterialListBinder` to bind data. A simple case, like binding an array of strings or an array of simple objects, would be useful.

lib/java/com/google/android/material/lists/MaterialListBinder.java
bohsen: @ethanhsuhsu @cketcham I think the API should be picky about what to accept as input in regards to data type to withstand changes later on. And I believe the `List` interface would be the right choice, as this is the case in almost all cases. Plus other `Collection` types are easily transformed into `List`s with `asList/toList` extensions/static utility methods. 

> In theory we could make an interface that has a method to return a size, and another to return an object of a particular type given the index. We'd be able to use that in the current implementation rather than passing the size directly. This would allow us to add a type, but keep the current flexibility.

If you look at how `androidx.recyclerview.widget.ListAdapter` is implemented, this is how they've done (using a class though - `AsyncListDiffer`). This also brings the benefit of being able to "install" a changelistener. It actually might be worth considering if `MaterialListAdapter` should reuse from `ListAdapter`/`AsyncListDiffer` if this ends up having a dependency on `androidx.recyclerview` anyway, as mentioned [here](https://github.com/material-components/material-components-android/issues/81#issuecomment-556043356). Also `androidx.recyclerview` version 1.1.0 was just released as stable.

> could be a `List`, `LinkedList`, `Array`, **`Map`**

IMO support for `Map`s is unnecessary and I also don't believe this would support it that well.  The main argument being that the `position` integer parameter would have to map to the value. In my experience integers are rarely used as keys in `Map`s. Please excuse me if I'm missing something in that regard.

But all of this is probably targeted at dynamic data support, which @ethanhsuhsu already mentioned [here](https://github.com/material-components/material-components-android/pull/798#discussion_r352803979) isn't the purpose of this PR. So maybe merge this  PR and continue with the next step in the implementation (Update the list item xml layouts to be more consistent with Material standards), if all other comments have been addressed? And then people interested in this can ponder on this until the next PR.

Btw. Sorry for not following best practice earlier in regard to my review. That was sloppy on my behalf.

cketcham: I also agree `List` is probably the most common, and it's pretty easy to convert to and from a list. I'd be ok with either `List` or `Iterable`.

I understand the purpose of this PR isn't targeted at dynamic data support.. but I don't know that the current api (passing only the size to the adapter) makes the most sense considering most often we'll want to support a `List` with objects with a consistent type. It might be better to take in a list in the constructor.. Or use `androidx.recyclerview.widget.ListAdapter` like you suggest to have the option of doing diffs easily.

ethanhsuhsu:
Based on the feedback I received (thanks @cketcham and @bohsen for the comments! ) I have added some ⚒ modifications to the core structures:

* The adapter is now standardized, no need for specific adapters for each view holder
* MaterialListBinder.onBind now provides specific data objects instead of an index 😄 
* MaterialListAdapter supports Lists, 🗒 Arrays and custom definable types.

And for the catalog app:
* Changed the demo implementation to use a String dataset
* Added a new picture/thumbnail 🖼 for Lists in the catalog card

In order to make these changes I did have to do some research on Java reflection and generics. @cketcham @bohsen Feel free to write reviews so we can discuss and improve the overall code quality!

lib/java/com/google/android/material/lists/viewholders/TwoLineItemViewHolder.java
```diff
@@ -14,29 +14,34 @@
...
-    this.secondary = itemView.findViewById(R.id.mtrl_list_item_secondary_text);
+    this.icon = itemView.findViewById(R.id.mtrl_list_item_icon);
+    this.primaryText = itemView.findViewById(R.id.mtrl_list_item_text);
+    this.secondaryText = itemView.findViewById(R.id.mtrl_list_item_secondary_text);
   }
-
-  @NonNull
-  public static TwoLineItemViewHolder create(@NonNull ViewGroup parent) {
-    return new TwoLineItemViewHolder(
+  public TwoLineItemViewHolder (@NonNull ViewGroup parent) {
```
EdricChan03: Extra space between the constructor declaration and the parameters?

```suggestion
  public TwoLineItemViewHolder(@NonNull ViewGroup parent) {
```

lib/java/com/google/android/material/lists/viewholders/SingleLineItemViewHolder.java
```diff
@@ -14,33 +14,37 @@
...
     this.icon = itemView.findViewById(R.id.mtrl_list_item_icon);
     this.text = itemView.findViewById(R.id.mtrl_list_item_text);
+
   }
 
-  @NonNull
-  public static SingleLineItemViewHolder create(@NonNull ViewGroup parent) {
-    return new SingleLineItemViewHolder(LayoutInflater.from(parent.getContext())
-        .inflate(R.layout.material_list_item_single_line, parent, false));
+  public SingleLineItemViewHolder(@NonNull ViewGroup parent){
```
EdricChan03: Space missing between the parameter declaration and the constructor body:

```suggestion
  public SingleLineItemViewHolder(@NonNull ViewGroup parent) {
```

lib/java/com/google/android/material/lists/viewholders/SingleLineItemViewHolder.java
```diff
@@ -14,33 +14,37 @@
...
-  public final ImageView icon;
-  public final TextView text;
+  public SingleLineItemViewHolder(@NonNull View itemView) {
+    super(itemView);
 
-  public SingleLineItemViewHolder(@NonNull View view) {
-    super(view);
     this.icon = itemView.findViewById(R.id.mtrl_list_item_icon);
     this.text = itemView.findViewById(R.id.mtrl_list_item_text);
+
```
EdricChan03: Remove the extra whitespace here.

lib/java/com/google/android/material/lists/listdatatypes/ListListData.java
```diff
@@ -0,0 +1,22 @@
+package com.google.android.material.lists.listdatatypes;
+
+import java.util.List;
+
+/** Implementation of {@link CustomListData} for {@link List}s */
+public class ListListData<T> implements CustomListData<T> {
```
EdricChan03: `ListListData`? Doesn't seem very conventional - I suggest we should rename the class to something like `ListData`?

lib/java/com/google/android/material/lists/listdatatypes/ArrayListData.java
```diff
@@ -0,0 +1,18 @@
+package com.google.android.material.lists.listdatatypes;
+
+/** Implementation of {@link CustomListData} for arrays*/
```
EdricChan03: Missing space:

```suggestion
/** Implementation of {@link CustomListData} for arrays */
```

lib/java/com/google/android/material/lists/MaterialListAdapter.java
```diff
@@ -0,0 +1,73 @@
...
+    } catch (InstantiationException e) {
+      e.printStackTrace();
+    } catch (InvocationTargetException e) {
+      e.printStackTrace();
+    } catch (NoSuchMethodException e) {
+      e.initCause(
+          new NoSuchMethodException(viewHolderClass.getName() +
+              " needs a constructor with only a ViewGroup parameter"));
+      e.printStackTrace();
+    }
```
EdricChan03: Could you explain why you're printing the stack trace and not actually handling the exceptions?

catalog/java/io/material/catalog/lists/MaterialListAdapterDemoFragment.java
```diff
@@ -0,0 +1,141 @@
...
+      "Eleventh Item",
+      "Twelfth Item",
+      "Thirteenth Item",
+      "Fourteenth Item",
+      "Fifteenth Item",
+      "Sixteenth Item",
+      "Seventeenth Item",
+      "Eighteenth Item",
+      "Nineteenth Item",
+      "Twentieth Item"};
```
EdricChan03: What about converting this to a resource string such that it can be translated in the future? You can then loop the single resource string 20 times and have a `%s` that is replaced with the number.

For e.g.:

```xml
<string name="cat_list_item_prefix">Item %d</string>
```

```java
private List<String> getListData() {
  List<String> listData = new ArrayList();
  for (int x = 1; x <= 20; x++) {
    listData.add(resources.getString(R.string.cat_list_item_prefix, x));
  }
  return listData;
}
```

lib/java/com/google/android/material/lists/listdatatypes/ListListData.java
ethanhsuhsu: Yes, I was trying to stick to the naming convention of (Data Type) + ListData. I renamed it to just ListData.

lib/java/com/google/android/material/lists/MaterialListBinder.java
ethanhsuhsu: Lists, arrays and custom types are now supported!

catalog/java/io/material/catalog/lists/MaterialListAdapterDemoFragment.java
ethanhsuhsu: The idea behind the demo is to show how to use Material Lists with a dataset. I define the dataset explicitly to prevent confusion and have readers understand the dataset. This later helps the reader understand the list code, rather than not knowing what the dataset looks like. 

I originally used Collections.nCopies() to automatically create items, but then I found the code harder to read and understand. This code is built for readability and not efficiency.

lib/java/com/google/android/material/lists/MaterialListAdapter.java
ethanhsuhsu: Other than printing why the error occurred what else can I do to "handle the exceptions" ?

Just for clarification, the IllegalAccessException, InstantiationException and InvocationTargetException cannot happen under normal circumstances. However, the NoSuchMethodException can happen so I added a message about why it happened.

ethanhsuhsu:
@cketcham Can I get a review of my new contributions? 

https://github.com/material-components/material-components-android/pull/798#issuecomment-565896414

lib/java/com/google/android/material/lists/MaterialListAdapter.java
```diff
@@ -0,0 +1,73 @@
...
+    this.materialListBinder = materialListBinder;
+    this.viewHolderClass = viewHolderClass;
+    customListData = new ListData<>(data);
+  }
+
+  public MaterialListAdapter(MaterialListBinder<T, E> materialListBinder, Class<E> viewHolderClass, CustomListData<T> data) {
+    this.materialListBinder = materialListBinder;
+    this.viewHolderClass = viewHolderClass;
+    customListData = data;
+  }
```
BeksOmega: I love that you added support for different collection types! But I still think this implementation could use a bit of polishing.

It feels to me like it would be better to have developers to convert the collection to a `CustomListData` outside of the constructor. This would cut down on duplicate code in your constructors. And if you defined extension functions like `toListData()` on some common collections (array, list, etc) creating a `ListAdapter` would still be really easy for developers.

But speaking of extensions, I'm having trouble seeing the advantage of defining the `CustomListData` class, rather than just using a `List` as bohsen and cketcham suggested. If you go with what I suggested above, it seems like you're basically duplicating the behavior of `List`. Is there a reason you decided to go with this implementation over the other?

lib/java/com/google/android/material/lists/MaterialViewHolder.java
```diff
@@ -0,0 +1,14 @@
+package com.google.android.material.lists;
+
+import android.view.View;
+
+import androidx.annotation.NonNull;
+import androidx.recyclerview.widget.RecyclerView;
+
+/** A standard view holder that all specific Material view holders extend */
+public abstract class MaterialViewHolder extends RecyclerView.ViewHolder {
```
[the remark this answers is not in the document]
ethanhsuhsu: I do agree with your point about ```e.printstacktrace()``` but here is the situation (excluding the NoSuchMethodException):
* These three errors are not meant to be thrown and there are no reasonable circumstances where they can be thrown. If you are able to call the constructor -> no error. If you cannot get the constructor -> NoSuchMethodException.
* These errors should not "bubble up" because in these cases the developer did not do anything wrong. 

What would be a better way to handle these errors @BeksOmega? Maybe we should bubble them up? How can we handle the errors more elegantly?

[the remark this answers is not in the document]
ethanhsuhsu: Regarding the ```MaterialViewHolder``` right now it is just for organization. In the future it will include real functionality when we have to stuff like reordering and selection.

Good question, it has previously been unclear.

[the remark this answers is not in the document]
BeksOmega: So I found the [AOSP guide for error handling](https://source.android.com/setup/contribute/code-style) (which I assume applies here?). From my reading, I believe if you don't expect the errors to be thrown the best thing to do is let them bubble up. But you're more familiar with this area so I think it's better if you give it a look.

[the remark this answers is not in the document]
ethanhsuhsu: Nice research, didn't even know that existed! After looking at the method definition for getConstructor() and newInstance() I learned that the three other exceptions are called if the class is configured incorrectly.

Therefore I am going to Log.e a description for each error when it happens. It is the equivalent of this option:

Screen Shot 2019-12-18 at 9 20 08 PM

- [X] Errors do occur (When class is configured incorrectly)
- [X] It should crash (The app has not setup their custom view holder correctly)

To clarify, these crashes only happen if the app defines a custom view holder. If the app uses our view holders it will be fine 😎

lib/java/com/google/android/material/lists/MaterialListAdapter.java
ethanhsuhsu: Good news, MaterialListBinder now integrates a data source into it. Check out the demo for implementation differences.

lib/java/com/google/android/material/lists/MaterialViewHolder.java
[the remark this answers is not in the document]
ethanhsuhsu: In the final implementation, I used RuntimeException instead of Log.e since it seems like the more recommended way to error crash.

catalog/java/io/material/catalog/tableofcontents/TocModule.java
```diff
@@ -66,7 +66,7 @@
       TabsFragment.Module.class,
       TextFieldFragment.Module.class,
       TopAppBarFragment.Module.class,
-      TransformationFragment.Module.class,
```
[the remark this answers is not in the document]
ethanhsuhsu: Good catch! No, it is not necessary. I will remove that.

lib/java/com/google/android/material/lists/MaterialListBinder.java
```diff
@@ -0,0 +1,25 @@
+package com.google.android.material.lists;
+
+import java.util.Arrays;
+import java.util.List;
+
+/** Interface that defines interactions between {@link MaterialViewHolder}s and data */
+public class MaterialListBinder<T, E extends MaterialViewHolder> {
+
+  protected List<T> data;
```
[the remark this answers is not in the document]
ethanhsuhsu: The main issue is that the way we handle data is scattered. Some of it would be in the adapter and other parts are in the Recyclerview itself. Having one class to serve the purpose of handling data would be simple for developers.

lib/java/com/google/android/material/lists/MaterialListAdapter.java
```diff
@@ -0,0 +1,54 @@
...
+    this.materialListBinder = materialListBinder;
+    this.viewHolderClass = viewHolderClass;
+  }
+
+  @NonNull
+  @Override
+  public E onCreateViewHolder(@NonNull ViewGroup parent, int viewType) {
+
+    try {
+      return viewHolderClass.getConstructor(ViewGroup.class).newInstance(parent);
```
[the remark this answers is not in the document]
ethanhsuhsu: The reflection implemention is safe, and it does create view holders appropriately. I am going to design the adapter so that you can override onCreate but with the ViewHolder already created (and onBind).

catalog/java/io/material/catalog/lists/res/layout/cat_material_list_adapter_popup_item.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
ethanhsuhsu: It ensures xml compatibility, but it is not necessary. I will remove it.

catalog/java/io/material/catalog/lists/ListsFragment.java
```diff
@@ -63,7 +76,7 @@ public Fragment createFragment() {
     @Provides
     @ActivityScope
     static FeatureDemo provideFeatureDemo() {
-      return new FeatureDemo(R.string.cat_lists_title, R.drawable.ic_lists_24px) {
```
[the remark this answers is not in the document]
ethanhsuhsu: When you use the catalog app, you can see that the component cards have a picture of the component included. I added a picture of a sample list for the "Lists" component card. Also... the vector does not work with dark mode

ethanhsuhsu:
As seen in this thread, shoving (the base of) a component into one PR is hard. To combat this, I am working closely with Cameron to discuss specifics 🖌 on creating components for Material in the near future.

Thank you 😸 for all your feedback on the "first draft" of Material lists over the last several weeks. Using your feedback, I am working with Cameron to establish a final architecture for this component (yay!) so that we can code efficiently and effectively together (also yay!) 🎆. 

💯 Stay tuned for more developments - Ethan Hsu

catalog/java/io/material/catalog/lists/ListsFragment.java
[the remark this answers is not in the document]
cketcham: ok this change sounds good, but could you pull it into a separate pr?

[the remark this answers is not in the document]
ethanhsuhsu: Sure thing, I will add the addition of a new thumbnail to a new PR

PR #821, opened 2019-12-11
Use conditional keep rule for MaterialComponentsViewInflater

### LINKED ISSUE #818: MaterialComponentViewInflater should not be marked @Keep
https://github.com/material-components/material-components-android/blob/fe160f33e65e25d6656f114a0712b112a4eadd61/lib/java/com/google/android/material/theme/MaterialComponentsViewInflater.java#L45-L46

This `@Keep` annotation will cause `MaterialComponentViewInflater` and thus `AppCompatViewInflater` to be kept even if they are not used. This in turn causes every custom subtype of the platform views inside AppCompat (and their resources) to be kept, even if unused.

The annotation should be removed and replaced with a conditional keep rule in the aar's consumer proguard rules that only keeps `MaterialComponentViewInflater` if `AppCompatViewInflater` is also kept.
```
-if class androidx.appcompat.app.AppCompatViewInflater
-keep class com.google.android.material.theme.MaterialComponentsViewInflater
```

This allows apps which are not using the view inflater to have the majority of these subtypes removed by shrinking and thus a bunch of their resources to be removed dramatically affecting the final APK size.

JakeWharton:
This allows it to be removed when `AppCompatViewInflater` is also unused, along with their many transitive class dependencies.

Closes #818

### DIFF
--- a/lib/proguard-inflater.pro
+++ b/lib/proguard-inflater.pro
@@ -12,5 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# MaterialComponentsViewInflater inflates Material Components rather than their AppCompat counterparts.
--keep class com.google.android.material.theme.MaterialComponentsViewInflater { *; }
+# AppCompatViewInflater reads the viewInflaterClass theme attribute which then reflectively
+# instantiates MaterialComponentsViewInflater using the no-argument constructor. We only need to
+# keep this constructor and the class name if AppCompatViewInflater is also being kept.
+-if class androidx.appcompat.app.AppCompatViewInflater
+-keep class com.google.android.material.theme.MaterialComponentsViewInflater {
+    <init>();
+}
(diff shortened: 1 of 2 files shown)


lib/proguard-inflater.pro
```diff
@@ -12,5 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# MaterialComponentsViewInflater inflates Material Components rather than their AppCompat counterparts.
--keep class com.google.android.material.theme.MaterialComponentsViewInflater { *; }
+# AppCompatViewInflater reads the viewInflaterClass theme attribute which then reflectively
```
dsn5ft: Thanks @JakeWharton! It looks like `viewInflaterClass` is technically read by `AppCompatDelegateImpl`:

https://cs.android.com/androidx/platform/frameworks/support/+/androidx-master-dev:appcompat/appcompat/src/main/java/androidx/appcompat/app/AppCompatDelegateImpl.java;l=1420;drc=9502ef7329a3a1b8e9dc4afbd51460ddf545050c

Should the condition be changed to be dependent on `AppCompatDelegateImpl` instead of `AppCompatViewInflater`?

JakeWharton: My hesitation there is that is a private type whose name is free to change at any time. Whereas `AppCompatViewInflater` is the public type that `MaterialComponentsViewInflater` extends and can never change.

dsn5ft: Sounds good, will submit the PR internally and sync the change back out here.

Issue #762, opened 2019-11-14
[MaterialDatePicker] Fullscreen setting paddedPickerWidth when docs describe 12dp


cammace:
**Description:** When using the MaterialDatePicker within my application with fullscreen mode enabled, I am getting an annoying padding behavior which along the sides of the calendar which I can only assume should not be there. 

**Expected behavior:** As detailed in the material design picker docs, the margins on the right and left side of the calendar should only be 12dp but I am seeing a calculated value. This appears to be because the width of `mtrl_calendar_frame` is adjusted using the [method found here](https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/datepicker/MaterialDatePicker.java#L398-L406). This logic makes sense for dialog hosted calendar but when it is in fullscreen, the width of the calendar should `match_parent`. 
![image](https://user-images.githubusercontent.com/5652865/68821438-59b4cb00-065c-11ea-9274-ed6cdf1f53ca.png)

**Actual behavior:**
<img src="https://user-images.githubusercontent.com/5652865/68821676-f5ded200-065c-11ea-8fc0-faef046b4426.png" width=360/>

**Source code:** My hunch is that this if statement for fullscreen adjusting the layout params for the container shouldn't be there. 

https://github.com/material-components/material-components-android/blob/c266251ee9f3124270e425173039a7067b0175d4/lib/java/com/google/android/material/datepicker/MaterialDatePicker.java#L183-L187

**Android API version:** Android API 28

**Material Library version:** `1.1.0-beta2`

**Device:** Pixel 2 (Android Studio Emulator)

matpag:
If you think the layout on widescreen, like a tablet, the `getPaddedPickerWidth` method prevent the picker to extend from this:

![Screenshot_1573774645](https://user-images.githubusercontent.com/4432213/68905078-3bc09680-0740-11ea-8439-47cee6475bec.png)

to this:

![Screenshot_1573774637](https://user-images.githubusercontent.com/4432213/68905085-411de100-0740-11ea-906a-2dfa853d6d8d.png)

Which to me seems the correct behaviour to apply.

This could be customizable with a custom style attribute but don't know if it is really worth. You need to wait for a material member answer on this request.

paulfthomas:
As explained in the previous comment (thanks for the screenshots @matpag!), this is the intended behavior and allows to limit the width of the picker.

PR #830, opened 2019-12-15
[MaterialDatePicker] Don't clip the background when there's a selected range of dates

rubensousa:
Partially addresses #762 

| Before | After |
| ------------- | ------------- |
| clipped  | unclipped  |

I'm not sure if this should be done for tablets or landscape mode, but according to the spec, the layout should look like this:

image


rubensousa:
@googlebot I signed it!

wcshi:
Design guidance for desktop / landscape advise drawing the range contained to the calendar region (existing behavior).

rubensousa:
@wcshi CanI close this then?

wcshi:
@rubensousa yes, please close, thank you for your interesting in contributing to the library.

Issue #803, opened 2019-12-02
[Slider] Implement OnSeekBarChangeListener support


Pitel:
https://developer.android.com/reference/android/widget/SeekBar.OnSeekBarChangeListener

**Is your feature request related to a problem? Please describe.**
* I want to perform some action when user stops dragging and finaly decides which value he wants.
* I don't want to fire any event, when the value is changed programatically (see #771)

None of it is easily doable with Slider, bit SeekBar can handle both.

**Describe the solution you'd like**
I'd like to be able to set `OnSeekBarChangeListener` on `Slider`.

**Describe alternatives you've considered**
Handling all the scanrious described above manually. With all those MotionEvents, it get's pretty messy, and buggy, and should be handled by Slider.

rubensousa:
I've opened a PR that fixes this: https://github.com/material-components/material-components-android/pull/831

cketcham:
fixed by cde09d4b4e8a2ad8a6d73e449f8fb09e404ea3ea

PR #831, opened 2019-12-16
[Slider] Add missing functionality from SeekBar regarding touch events and value changes

### LINKED ISSUE #771: [Slider] Allow to distinguish between user and programmatic change events
When using a `com.google.android.material.slider.Slider.OnChangeListener` it is not possibel to distinguish between user made changes or programmatic changes.

Similar to the app compat slider, it should be possible to be informed whether the change is user triggered or made by an API call.

rubensousa:
Fixes #771 and #803 

### Value changes

There's no way to know currently if the value is being changed due to a user triggered event.

To fix this, OnChangeListener was changed from:

```java
public interface OnChangeListener {
   void onValueChange(Slider slider, float value);
}
```

to:

```java
public interface OnChangeListener {
   void onValueChange(Slider slider, float value, boolean fromUser);
}
```

### Touch event tracking

There's currently no way to determine when the user is touching or releasing the Slider.
A new OnSliderTouchListener was added instead of adding the methods to OnChangeListener to track the touch events:

```java
public interface OnSliderTouchListener {
  void onStartTrackingTouch(Slider slider);

  void onStopTrackingTouch(Slider slider);
}
```

### DIFF
--- a/catalog/java/io/material/catalog/slider/SliderMainDemoFragment.java
+++ b/catalog/java/io/material/catalog/slider/SliderMainDemoFragment.java
@@ -25,6 +25,8 @@ import android.view.View;
 import android.view.ViewGroup;
 import android.widget.Button;
 import com.google.android.material.slider.Slider;
+import com.google.android.material.snackbar.Snackbar;
+
 import io.material.catalog.feature.DemoFragment;
 
 /** A fragment that displays the main Slider demo for the Catalog app. */
@@ -37,6 +39,19 @@ public class SliderMainDemoFragment extends DemoFragment {
         layoutInflater.inflate(R.layout.cat_slider_fragment, viewGroup, false /* attachToRoot */);
 
     Slider slider = view.findViewById(R.id.slider);
+    slider.setOnSliderTouchListener(new Slider.OnSliderTouchListener() {
+      @Override
+      public void onStartTrackingTouch(Slider slider) {
+        Snackbar.make(slider, getText(R.string.cat_slider_start_touch_description),
+            Snackbar.LENGTH_SHORT).show();
+      }
+
+      @Override
+      public void onStopTrackingTouch(Slider slider) {
+        Snackbar.make(slider, getText(R.string.cat_slider_stop_touch_description),
+            Snackbar.LENGTH_SHORT).show();
+      }
+    });
     Button button = view.findViewById(R.id.button);
 
     button.setOnClickListener(v -> slider.setValue(slider.getValueTo()));
--- a/catalog/java/io/material/catalog/slider/SliderMainDemoFragment.java
+++ b/catalog/java/io/material/catalog/slider/SliderMainDemoFragment.java
@@ -42,13 +42,13 @@ public class SliderMainDemoFragment extends DemoFragment {
     slider.setOnSliderTouchListener(new Slider.OnSliderTouchListener() {
       @Override
       public void onStartTrackingTouch(Slider slider) {
-        Snackbar.make(slider, getText(R.string.cat_slider_start_touch_description),
+        Snackbar.make(slider, R.string.cat_slider_start_touch_description,
             Snackbar.LENGTH_SHORT).show();
       }
 
       @Override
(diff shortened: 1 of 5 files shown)


lib/java/com/google/android/material/slider/Slider.java
```diff
@@ -976,8 +998,11 @@ public boolean onTouchEvent(@NonNull MotionEvent event) {
         thumbPosition = position;
         snapThumbPosition();
         ViewUtils.getContentViewOverlay(this).remove(label);
+        onStopTrackingTouch();
         invalidate();
         break;
+      case MotionEvent.ACTION_CANCEL:
```
cketcham: I think this can't happen because we call `requestDisallowInterceptTouchEvent(true)`. Do you see a case where this would be called? If so, do we have to do anything else here that happens in `ACTION_UP`?

lib/java/com/google/android/material/slider/Slider.java
```diff
@@ -572,6 +583,15 @@ public void setOnChangeListener(@Nullable OnChangeListener listener) {
     this.listener = listener;
   }
 
+  /**
+   * Registers a callback to be invoked when the slider touch event is being started or stopped
+   *
+   * @param listener The callback to run when the slider starts or stops being touched
+   */
+  public void setOnSliderTouchListener(@Nullable OnSliderTouchListener listener) {
```
cketcham: Could we keep a list with add and remove methods instead? I'm planning to make that same change to the `OnChangeListener`.

catalog/java/io/material/catalog/slider/SliderMainDemoFragment.java
```diff
@@ -37,6 +39,19 @@ public View onCreateDemoView(
         layoutInflater.inflate(R.layout.cat_slider_fragment, viewGroup, false /* attachToRoot */);
 
     Slider slider = view.findViewById(R.id.slider);
+    slider.setOnSliderTouchListener(new Slider.OnSliderTouchListener() {
```
cketcham: I think just adding this behavior to the main demo only is enough. Could you remove it from the other demos?

catalog/java/io/material/catalog/slider/SliderMainDemoFragment.java
```diff
@@ -37,6 +39,19 @@ public View onCreateDemoView(
...
+      @Override
+      public void onStartTrackingTouch(Slider slider) {
+        Snackbar.make(slider, getText(R.string.cat_slider_start_touch_description),
```
cketcham: you can just pass the resource id to this method

catalog/java/io/material/catalog/slider/SliderMainDemoFragment.java
rubensousa: Sure, I can do that :)

lib/java/com/google/android/material/slider/Slider.java
rubensousa: Yes, I can do this :)

catalog/java/io/material/catalog/slider/SliderMainDemoFragment.java
rubensousa: Done

catalog/java/io/material/catalog/slider/SliderMainDemoFragment.java
rubensousa: Done

lib/java/com/google/android/material/slider/Slider.java
rubensousa: You're right, I'll remove this section.

rubensousa: Done

lib/java/com/google/android/material/slider/Slider.java
rubensousa: Done. I've also done it for OnChangeListener since it's similar

rubensousa:
@cketcham I've addressed your suggestions. Please review once you have time again :)

PR #861, opened 2019-12-29
[HideBottomViewOnScrollBehavior] Add support for recycler view with reversed linear layout manager

Atomofiron:
BottomAppBar has uncoordinated behavior because of target RecyclerView's LinearLayoutManager has reversed layout.
```kotlin
linearLayoutManager.reverseLayout = true
```

Current behavior with non-reversed layout:

Current behavior with reversed layout:

Desired behavior with reversed layout:


Atomofiron:
@googlebot I signed it!

Issue #689, opened 2019-10-18
[TextInputLayout] ErrorText-Counter combination results in misplacement


LukasMitterhofer:
**Description:** When displaying an error and a character counter on small devices the character count is misplaced/cutoff
![image](https://user-images.githubusercontent.com/48203056/67081561-aa3d2380-f197-11e9-9644-173f273ee67a.png)

**Expected behavior:** The error text should have a linebreak instead

**Source code:** 
```
inputLayout.apply {
            isCounterEnabled = true
            counterMaxLength = limit
}
...
inputLayout.error = "A maximum of 140 characters is possible"
```
**Android API version:** 21

**Material Library version:** Material Android Library version 1.0.0

**Device:** Samsung Galaxy J1

PR #873, opened 2020-01-06
[BottomSheetDialog] Removed existing content during each setContentView

### LINKED ISSUE #698: [TextInputLayout] Features for prefix and suffix
Phone edit field with country code as prefix

Implement the possibility to add click listeners and "always visibility" option for prefix and suffix. In my case this needs for show country code picker(dialog).

**Describe alternatives you've considered**
In my implementation i use `SpannableString` with `ClickableSpan` for listening click action on country code.

Снимок

### LINKED ISSUE #869: [BottomSheetDialog] View overlapping when setContentView() called multiple times.
**Description:** When `setContentView(View view)` method is called a second time or every subsequent time on the same instance of BottomSheetDialog in order to set a new view (smaller height), the new view to set overlaps the previous view set on the BottomSheetDialog instead of the previous view being removed first. 
Also, the height of the BottomSheetDialog is not re-calculated when updating with the new view with a smaller height and instead remains the same as the previous view set on that instance of BottomSheetDialog. The height seems to be re-calculated only when updating with the new view with a greater height.

BottomSheetDialog overlapping bug

**Expected behavior:** Calling `setContentView(View view)` method on an instance of BottomSheetDialog should remove any existing view set and update it with the new view and also re-calculate the height of the BottomSheetDialog based on the new view.

**Source code:** 
```
        BottomSheetDialog dialog = new BottomSheetDialog(this);
        View view1 = LayoutInflater.from(this).inflate(R.layout.view1, null);
        dialog.setContentView(view1);
        dialog.show();

        View view2 = LayoutInflater.from(this).inflate(R.layout.view2, null);
        dialog.setContentView(view2);
```

**Android API version:** Android API 29

**Material Library version:** 1.2.0-alpha03

**Device:** Google Pixel 3A

ethanhsuhsu:
closes #869

Previously, calling setContentView added the parameter view to the layout which would create overlap between it and previously added views. Removing previous views before the new one is added prevents this issue.

setContentView is meant to *set* the view, not add it to others.

### DIFF
--- a/lib/java/com/google/android/material/bottomsheet/BottomSheetDialog.java
+++ b/lib/java/com/google/android/material/bottomsheet/BottomSheetDialog.java
@@ -204,6 +204,7 @@ public class BottomSheetDialog extends AppCompatDialog {
     }
 
     FrameLayout bottomSheet = (FrameLayout) container.findViewById(R.id.design_bottom_sheet);
+    bottomSheet.removeAllViews();
     if (params == null) {
       bottomSheet.addView(view);
     } else {


afohrman:
Thanks for the PR! It looks like it closes https://github.com/material-components/material-components-android/issues/869, not https://github.com/material-components/material-components-android/issues/698.

Could you please update the PR to reflect the correct issue it closes?

ethanhsuhsu:
Yep, it is updated now

EdricChan03:
@ethanhsuhsu Looks like you've still used the wrong issue to be closed: you edited it to #689 instead of #869.

ethanhsuhsu:
Wow, I just got very unlucky...

It now redirects @afohrman to 869

PR #874, opened 2020-01-06
[ExtendedFloatingActionButton] Allowed for width to adapt to text size after expand animation

### LINKED ISSUE #858: [ExtendedFloatingActionButton] ExtendedFAB size does not adjust when changed text
**Description:** 
ExtendedFAB size does not adjust when shrink/extended ExtendedFAB text is changed.

- Happens in ExtendedFAB once shrink/extended.
- Happens when with or without CollapsingToolbar.

**Expected behavior:** 

**Source code:** 
https://github.com/NUmeroAndDev/ExtendedFABIssueExample-android/tree/changed_text_issue

Please look `changed_text_issue` branch.  

**Android API version:** 29

**Material Library version:** 1.1.0-rc01

**Device:** Pixel 3 XL

ethanhsuhsu:
closes #858 

After the button has been expanded, the width is statically defined, meaning it stays at a single value. If the text is changed, the width will no longer grow and shrink with the size of the text, instead creating whitespace and cutting out words. To fix it, the width is set
to WRAP_CONTENT after the animation happens, which allows the width
to change with the text.

### DIFF
--- a/lib/java/com/google/android/material/floatingactionbutton/BaseMotionStrategy.java
+++ b/lib/java/com/google/android/material/floatingactionbutton/BaseMotionStrategy.java
@@ -25,6 +25,8 @@ import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.core.util.Preconditions;
 import android.view.View;
+import android.view.ViewGroup;
+
 import com.google.android.material.animation.AnimatorSetCompat;
 import com.google.android.material.animation.MotionSpec;
 import java.util.ArrayList;
@@ -99,6 +101,9 @@ abstract class BaseMotionStrategy implements MotionStrategy {
   @Override
   @CallSuper
   public void onAnimationEnd() {
+    ViewGroup.LayoutParams newLayoutParams =  fab.getLayoutParams();
+    newLayoutParams.width = ViewGroup.LayoutParams.WRAP_CONTENT;
+    fab.setLayoutParams(newLayoutParams);
     tracker.clear();
   }
 


ethanhsuhsu:
The code for the "extend" animation is here:
```java
extendStrategy = new ChangeSizeStrategy(
        changeSizeTracker,
        new Size() {
          @Override
          public int getWidth() {
            return getMeasuredWidth();
          }

          @Override
          public int getHeight() {
            return getMeasuredHeight();
          }
        },
        /* extending= */ true);  
```

As you see, the animation leaves the button with a defined value for the width `getMesauredWidth`. So the resulting width is correct, but WRAP_CONTENT still needs to be set. 

> Instead of adding a new listener, could we just always return WRAP_CONTENT as the width from extendStrategy?
By adding the listener during the strategy's declaration that is what we are doing. If you want me to move the code closer to the declaration I can do that.

cketcham:
Ok that make sense to me, what I think isn't totally clear from the code at first glance, is that it needs the width to be set during the animation to some value so it can interpolate between those values, but after the animation is over it should set the width to `WRAP_CONTENT`.

Maybe in `ChangeSizeStrategy#onAnimationEnd()` it should always be setting the width to `WRAP_CONTENT`?

ethanhsuhsu:
Yeah, I was trying to find the source for this kind of animation. That seems more optimal

ethanhsuhsu:
@cketcham I ended up applying the change in `BaseMotionStrategy` since that is where you can get a reference to the Ext FAB. Should be good now 😎

cketcham:
Actually found an issue with shrinking, but we're looking into this bug :)

ethanhsuhsu:
Sounds good @cketcham

ymarian:
This was fixed in https://github.com/material-components/material-components-android/commit/55200c3a8f18604a723c7ad6ab48fb9ec03391ba

ethanhsuhsu:
Yeet

PR #875, opened 2020-01-07
[Lists] Added new thumbnail for entry in the catalog

ethanhsuhsu:
Currently, a black vector drawable is what appears when you see the card for "lists" on the home screen of the catalog. To fit in with the other thumbnails of other components, this commit adopts a picture of an actual list.

Requested by @cketcham in pull request #798


ethanhsuhsu:
I also decided to delete the old vector asset, as it is unused in the repository after this commit.

PR #883, opened 2020-01-08
[ExtendedFloatingActionButton] Added option to hide full button on collapse

### LINKED ISSUE #880: How to get ExtendedFloatingActionButton to behave similar to FloatingActionButton when interacting with CoordinatorLayout and scrolling?
Trying to use the new `ExtendedFloatingActionButton`. It seems by default `ExtendedFloatingActionButton` acts differently from `FloatingActionButton` when interacting with `CoordinatorLayout`. See here the red one is the typical `FloatingActionButton` and the 2 black ones are the `ExtendedFloatingActionButton` with exactly the same setup (except for colors and text).
What I need is for the black ones to hide away just like the red one does.

FAB vs EXFAB small

Here's the layout:
```
<androidx.coordinatorlayout.widget.CoordinatorLayout xmlns:android="http://schemas.android.com/apk/res/android"
    android:fitsSystemWindows="true"
    tools:context=".ScrollingActivity">

    <com.google.android.material.appbar.AppBarLayout
        android:id="@+id/app_bar"
        android:layout_height="@dimen/app_bar_height"
        android:fitsSystemWindows="true"
        android:theme="@style/AppTheme.AppBarOverlay">

        <com.google.android.material.appbar.CollapsingToolbarLayout
            android:id="@+id/toolbar_layout"
            android:fitsSystemWindows="true"
            app:contentScrim="?attr/colorPrimary"
            app:layout_scrollFlags="scroll|exitUntilCollapsed"
            app:toolbarId="@+id/toolbar">

            <androidx.appcompat.widget.Toolbar
                android:id="@+id/toolbar"
                android:layout_height="?attr/actionBarSize"
                app:layout_collapseMode="pin"
                app:popupTheme="@style/AppTheme.PopupOverlay" />

        </com.google.android.material.appbar.CollapsingToolbarLayout>
    </com.google.android.material.appbar.AppBarLayout>

    <include layout="@layout/content_scrolling" />

    <com.google.android.material.floatingactionbutton.FloatingActionButton
        android:id="@+id/fab"
        android:layout_margin="@dimen/fab_margin"
        app:layout_anchor="@id/app_bar"
        app:layout_anchorGravity="bottom|end"
        app:srcCompat="@android:drawable/ic_dialog_email" />

    <com.google.android.material.floatingactionbutton.ExtendedFloatingActionButton
        android:theme="@style/Theme.MaterialComponents.NoActionBar"
        android:id="@+id/exfab1"
        android:layout_margin="@dimen/fab_margin"
        app:layout_anchor="@id/app_bar"
        app:layout_anchorGravity="bottom|start"
        app:strokeColor="#ffff00"
        app:strokeWidth="4dp"
        app:iconTint="#ffff00"
        android:text="txt"
        android:textColor="#ffff00"
        android:backgroundTint="#000000"
        app:icon="@android:drawable/ic_dialog_email" />

    <com.google.android.material.floatingactionbutton.ExtendedFloatingActionButton
        android:theme="@style/Theme.MaterialComponents.NoActionBar"
        android:id="@+id/exfab2"
        android:layout_margin="@dimen/fab_margin"
        app:layout_anchor="@id/app_bar"
        app:layout_anchorGravity="bottom|center"
        app:strokeColor="#ffff00"
        app:strokeWidth="4dp"
        app:iconTint="#ffff00"
        android:textColor="#ffff00"
        android:backgroundTint="#000000"
        app:icon="@android:drawable/ic_dialog_email" />

</androidx.coordinatorlayout.widget.CoordinatorLayout>
```

I see in source that there's a nested class `ExtendedFloatingActionButton.ExtendedFloatingActionButtonBehavior` and it seems to have maybe some properties that might be useful/related to this such as `autoHideEnabled` and `autoShrinkEnabled` but the class is protected and I can't figure out how it is meant to be accessed.
Please help.

ethanhsuhsu:
Until this commit, Extended FABs could only partially collapse and kept
the icon visible. Now, Extended FABs have an option to fully collapse,
similar to the collapse behavior of the FloatingActionButton

- [X] Identify the component the PR relates to in brackets in the title.
- [X] Link to GitHub issues it solves. closes #880
- [X] Sign the CLA bot. You can do this once the pull request is opened.

### DIFF
--- a/lib/java/com/google/android/material/floatingactionbutton/ExtendedFloatingActionButton.java
+++ b/lib/java/com/google/android/material/floatingactionbutton/ExtendedFloatingActionButton.java
@@ -36,11 +36,11 @@ import android.util.AttributeSet;
 import android.util.Property;
 import android.view.Gravity;
 import android.view.View;
-import android.view.ViewGroup;
 import android.view.ViewGroup.LayoutParams;
 import androidx.coordinatorlayout.widget.CoordinatorLayout;
 import androidx.coordinatorlayout.widget.CoordinatorLayout.AttachedBehavior;
 import androidx.coordinatorlayout.widget.CoordinatorLayout.Behavior;
+//import androidx.coordinatorlayout.widget.CoordinatorLayout.LayoutParams;
 import com.google.android.material.animation.MotionSpec;
 import com.google.android.material.appbar.AppBarLayout;
 import com.google.android.material.bottomsheet.BottomSheetBehavior;
@@ -199,6 +199,23 @@ public class ExtendedFloatingActionButton extends MaterialButton implements Atta
             context, attrs, defStyleAttr, DEF_STYLE_RES, ShapeAppearanceModel.PILL
         ).build();
     setShapeAppearanceModel(shapeAppearanceModel);
+
+    //Patches corner case where the width does not fit to the text: http://bit.ly/35vSUJe
+    addOnExtendAnimationListener(new AnimatorListener() {
+
+      @Override
+      public void onAnimationEnd(Animator animation) {
+        LayoutParams newLayoutParams =  ExtendedFloatingActionButton.this.getLayoutParams();
+        newLayoutParams.width = LayoutParams.WRAP_CONTENT;
+        ExtendedFloatingActionButton.this.setLayoutParams(newLayoutParams);
+      }
+      @Override
+      public void onAnimationStart(Animator animation) {}
+      @Override
+      public void onAnimationCancel(Animator animation) {}
+      @Override
+      public void onAnimationRepeat(Animator animation) {}
+    });
   }
 
   @Override
@@ -215,6 +232,21 @@ public class ExtendedFloatingActionButton extends MaterialButton implements Atta
   @Override
   public Behavior<ExtendedFloatingActionButton> getBehavior() {
     return behavior;
+  }
+
+  /**
+   * Determines how the extended button is hidden when it needs to be collapsed
+   * @param isHide If true, none of the button will be visible when collapsed. If false, the button
+   * will "shrink" and hide button text but keep the icon.
+   *
+   * This option is "false" and perform a shrink by default.
+   */
+  public void setHideBehavior(boolean isHide) {
+
+    ExtendedFloatingActionButtonBehavior<ExtendedFloatingActionButton> extendedBehavior = (ExtendedFloatingActionButtonBehavior<ExtendedFloatingActionButton>) behavior;
+    extendedBehavior.setAutoHideEnabled(isHide);
+    extendedBehavior.setAutoShrinkEnabled(!isHide);
+
   }
 
 
@@ -764,7 +796,7 @@ public class ExtendedFloatingActionButton extends MaterialButton implements Atta
     }
 
     private static boolean isBottomSheet(@NonNull View view) {
-      final ViewGroup.LayoutParams lp = view.getLayoutParams();
+      final LayoutParams lp = view.getLayoutParams();
       if (lp instanceof CoordinatorLayout.LayoutParams) {
         return ((CoordinatorLayout.LayoutParams) lp).getBehavior() instanceof BottomSheetBehavior;
       }
@@ -1092,7 +1124,7 @@ public class ExtendedFloatingActionButton extends MaterialButton implements Atta
       super.onAnimationEnd();
       animState = ANIM_STATE_NONE;
       if (!isCancelled) {
-        setVisibility(GONE);
+        setVisibility(INVISIBLE);
       }
     }
   }


lib/java/com/google/android/material/floatingactionbutton/ExtendedFloatingActionButton.java
```diff
@@ -217,6 +234,21 @@ protected void onAttachedToWindow() {
...
+  /**
+   * Determines how the extended button is hidden when it needs to be collapsed
+   * @param isHide If true, none of the button will be visible when collapsed. If false, the button
+   * will "shrink" and hide button text but keep the icon.
+   *
+   * This option is "false" and perform a shrink by default.
+   */
+  public void setHideBehavior(boolean isHide) {
+
+    ExtendedFloatingActionButtonBehavior<ExtendedFloatingActionButton> extendedBehavior = (ExtendedFloatingActionButtonBehavior<ExtendedFloatingActionButton>) behavior;
```
cketcham: We should probably just change the type of the behavior to be `ExtendedFloatingActionButtonBehavior`. I don't think it would be anything else.

lib/java/com/google/android/material/floatingactionbutton/ExtendedFloatingActionButton.java
```diff
@@ -1092,7 +1124,7 @@ public void onAnimationEnd() {
       super.onAnimationEnd();
       animState = ANIM_STATE_NONE;
       if (!isCancelled) {
-        setVisibility(GONE);
+        setVisibility(INVISIBLE);
```
cketcham: Is this required?

lib/java/com/google/android/material/floatingactionbutton/ExtendedFloatingActionButton.java
```diff
@@ -217,6 +234,21 @@ protected void onAttachedToWindow() {
...
   }
 
+  /**
+   * Determines how the extended button is hidden when it needs to be collapsed
+   * @param isHide If true, none of the button will be visible when collapsed. If false, the button
+   * will "shrink" and hide button text but keep the icon.
+   *
+   * This option is "false" and perform a shrink by default.
+   */
+  public void setHideBehavior(boolean isHide) {
```
cketcham: A client could call `getBehavior()` and then call these methods directly. Do you think there's value in adding this wrapper method? Or could we just add documentation?

cketcham: Because of the addition of this method, it looks like this api is a bit confusing. I think because if `autoShrink` is true it takes precedence over `autoHide` we could make this into an IntDef or something where you can only set the collapse behavior to one of them. For example, you could call `setCollapseBehavior(HIDE)` to perform the same thing as `setHideBehavior(true)`, or `setCollapseBehavior(SHRINK)` for `setHideBehavior(false)`. Then you'd also have any easy way to disable any of this by calling `setCollapseBehavior(NONE)`. What do you think about adding that method and deprecating the others?

lib/java/com/google/android/material/floatingactionbutton/ExtendedFloatingActionButton.java
ethanhsuhsu: Yes it is required. If you don't make it invisible then the ext fab loses its layout position. And then when it reanimates back in, it starts in the bottom right corner and quickly jumps to the center line. This happens because alpha is being animated, but technically the view is not marked as `VISIBLE` so the layout position is not calculated.

lib/java/com/google/android/material/floatingactionbutton/ExtendedFloatingActionButton.java
ethanhsuhsu: Okay, I will add that in...

lib/java/com/google/android/material/floatingactionbutton/ExtendedFloatingActionButton.java
ethanhsuhsu: That it what I invisioned, for an enumaration. Originally I just wanted to keep my changes minimal was to maintain the current API. 

Great, glad you are onboard with the changesm I will change the booleans to enums in the client method and throughout the behavior class

ethanhsuhsu: Actually... they can't. ExtendedFabBehavior is marked protected so the ExtendedFabBehavior methods cannot be called. This created a need for a client method to be exposed.
 I would not recommend changing the class to public since it is just this one case that needs to be addressed. It would be unnecessary to expose an entire separate class to call... well, one method

ethanhsuhsu: However, if you want that to happen I would be glad implementing that.

I am just a contributor and I want to make API design decisions that match your needs.

falsevapor: 'getBehavior' returns 'Behavior<ExtendedFloatingActionButton>', doesn't it? A client can't cast that to protected 'ExtendedFloatingActionButton.ExtendedFloatingActionButtonBehavior' in order to get to the properties without extending from 'ExtendedFloatingActionButton' or somthing. Or am I missing something?

ethanhsuhsu: You are missing nothing @falsevapor. That is the exact conversation we are having!

falsevapor: I don't know how realistic this is but is there absolutely no way to combine both? Does it have to be one OR the other? Why not? I see a possible need to first autoshrink the button when collapsing the app bar to the regular/thin toolbar and then autohiding altogether when the toolbar is also collapsed. It seems to me there's a scenario when you might want both.

falsevapor: @ethanhsuhsu, sorry, I missed your relies here. Somehow github just showed the message from @cketcham only... nevermind

ethanhsuhsu: @falsevapor You may be on to something, how about you post that in a separate GitHub issue? I will address that when this is complete 😉 

With this PR, we have to address some challenges integrating the ExtFABBehavior class. This PR does not to be more complicated than it already is.

ethanhsuhsu: @cketcham What do you think about this decision. Should we make the ExtFABBehavior class public, or keep it how it is and expose a client method?

falsevapor: right, for now just to allow the access to one or the other is enough to get my original question answered/handled.

ethanhsuhsu: Ok

falsevapor: Thank you. Just let me know when you are ready for me to test this. Seems like it's still in discussion.

cketcham: Yeah.. the fact that that class is protected is what I was missing. I guess it makes sense for it to be protected since it's very specific to Extended Fab..

cketcham: I think having the option to first auto shrink and then hide would be great! But it's also outside the scope of this PR. Maybe we could change the api that you've added to `ExtendedFloatingActionButton`? Instead of having the method take a boolean, you could have it take an intdef or enum for HIDE, SHRINK, and NONE.

In the future, this should allow us to add new enum values that let us create more complicated behavior such as the option to do both like you've described.

lib/java/com/google/android/material/floatingactionbutton/ExtendedFloatingActionButton.java
cketcham: the `performNow()` method can set it to GONE because it's not animating? Or should that also be INVISIBLE.

lib/java/com/google/android/material/floatingactionbutton/ExtendedFloatingActionButton.java
falsevapor: Agreed. I opened a new ticket about allowing both.

lib/java/com/google/android/material/floatingactionbutton/ExtendedFloatingActionButton.java
ethanhsuhsu: I would say keep it since issues do not arise from it currently. 

Because I am contributing, I think as a general philosophy I should probably only address issue (or PR) specific issues. Architectural improvements can be saved for another issue / PR combo. Otherwise you get PR #798 😉 

Sorry, just a drawback of working with a good ol' contributor...

cketcham: Yeah I don't want to increase the scope of this PR, but I do want to make sure we aren't missing a case and this bug is fully fixed. I think `performNow()` is called instead if the view isn't laid out yet, wouldn't that mean it would lose its layout position if that happens? Or is something else happening that fixes this?

alshamiri5 (approved):

alshamiri5 (approved):

Hanna15091 (approved):

lib/java/com/google/android/material/floatingactionbutton/ExtendedFloatingActionButton.java
Hanna15091: - []()

PR #946, opened 2020-01-24
[Date Picker] Created a date validator class to handle resctrictions from previous dates until the current date.

onats1:

PR #972, opened 2020-02-03
[MaterialDatePicker] Added a DateValidator for a bounded range.

### LINKED ISSUE #971: [MaterialDatePicker] Add a DateValidator for a Bounded Range
There is no pre-built `DateValidator` that allows for a bounded range of accepted dates.

A default `DateRangeValidator` that can accept a `lower bound` and `upper bound` range would be an ideal solution.

froriz5:
Creates a new `DateValidator` for a bounded range of valid dates.
Ex:
`1/1/2020 - 2/1/2020`

- [x] Link to GitHub issues it solves. closes #971 


ymarian:
I think it makes more sense to create a CompositeDateValidator so that you can combine the two existing ones. 

Would you mind updating the PR to support a validator that takes a list of validators?

froriz5:
@ymarian in regards to the your comment:
> ... create a CompositeDateValidator so that you can combine the two existing ones.

As far as I can see, there's only a single pre-built validator in the library [DateValidatorPointForward](https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/datepicker/DateValidatorPointForward.java). This one can handle a lower bound date range.

Is there a second pre-built validator that can handle an upper bound date range?

ymarian:
@froriz5 I thought the pull request right below this one, was merged already:
https://github.com/material-components/material-components-android/pull/946
We decided that one makes sense.

So I think providing forwards, backwards, and composite gives people enough flexibility to write their own for any other case.

froriz5:
@ymarian I pushed a commit removing the `DateRangeValidator` and replacing it with a `CompositeDateValidator` 👍

lib/java/com/google/android/material/datepicker/CompositeDateValidator.java
```diff
@@ -0,0 +1,110 @@
...
+   */
+  @Override
+  public boolean isValid(long date) {
+    boolean valid = true;
+    for (DateValidator validator: validators) {
+      if (validator != null) {
+        valid = valid && validator.isValid(date);
+      }
+    }
+    return valid;
```
ymarian: nit:

```java
  if (validator == null) {
	  continue;
	}
	if (!validator.isValid(date)) {
    return false;
  }
}

return true;
```

froriz5: Updated 👍

Issue #810, opened 2019-12-05
[TextInputLayout] Allow hint to float always


mauronra:
**Is your feature request related to a problem? Please describe.**
Currently it is impossible to set a hint to both TextInputLayout and TextInputEditText without both overlapping, in some cases it would be nice to show both hints to the user.

**Describe the solution you'd like**
Create an attribute such as:

`android:hintAlwaysFloat="true"`

Or something like it.

**Describe alternatives you've considered**
The easiest alternative to achieve such design is to set a TextView and an EditText within a container and handle both of them.

I've also [seen](https://stackoverflow.com/a/55270916/3624477) the use of reflection to achieve such behaviour, however this brings undesired problems.

**Additional context**
Actual: 

```
<com.google.android.material.textfield.TextInputLayout
            android:id="@+id/input"
            android:layout_width="match_parent"
            android:layout_height="wrap_content"
            android:hint="Name">

            <com.google.android.material.textfield.TextInputEditText
                android:layout_width="match_parent"
                android:layout_height="wrap_content"
                android:hint="e. g. Mauro N."
                android:inputType="text"/>

</com.google.android.material.textfield.TextInputLayout>
```

![image](https://user-images.githubusercontent.com/38300107/70277922-48556f00-1781-11ea-9774-8968a63a79c2.png)

Suggested:

```
<com.google.android.material.textfield.TextInputLayout
            android:id="@+id/input"
            android:layout_width="match_parent"
            android:layout_height="wrap_content"
            android:hint="Name"
            android:hintAlwaysFloat="true">

            <com.google.android.material.textfield.TextInputEditText
                android:layout_width="match_parent"
                android:layout_height="wrap_content"
                android:hint="e. g. Mauro N."
                android:inputType="text"/>

</com.google.android.material.textfield.TextInputLayout>
```

![image](https://user-images.githubusercontent.com/38300107/70277961-56a38b00-1781-11ea-9fe1-6031ea6dd06b.png)

leticiarossi:
Can't this be solved by using the `placeholderText` attribute instead?

ymarian:
Hi I think @leticiarossi it's right, the component supports this via placeHolderText

```xml
  <com.google.android.material.textfield.TextInputLayout
      android:id="@+id/text_input_username"
      style="@style/Widget.MaterialComponents.TextInputLayout.FilledBox"
      android:layout_width="wrap_content"
      android:layout_height="wrap_content"
      android:hint="Name"
[...]

PR #1028, opened 2020-02-17
[TextInputLayout] Add hintAlwaysFloat attribute to TextInputLayout

lupajz:
This adds attribute for keeping TextInputLayouts hint in collapsed state.
Refences issue #810


lupajz:
@googlebot I signed it!

ymarian:
I commented on the issue with the preferred solution

zx2c4:
The real solution to this problem is this trivial fix: https://github.com/material-components/material-components-android/pull/1163

PR #1049, opened 2020-02-22
[Motion] Updated documentation

dcampogiani:
In the example, it's always used "shared_element_container" as the transition name, but in the Navigation Architecture Component snippet "container" is used.


dcampogiani:
@googlebot I signed it!

hunterstich:
Thanks for catching this!

PR #1096, opened 2020-03-08
[TextInputLayout] Remove perform click in DropdownMenuEndIconDelegate

### LINKED ISSUE #1080: Exposed Dropdown Menu - Click sounds twice when clicking
**Description:**
When implementing the menu on Android as per https://material.io/develop/android/components/menu/ if your device or emulator (tested and verified with both) and you click on the field, you hear two quick clicks as if the listener was called twice. If you click the little drop down arrow instead of the text field, you only hear a single sound.

**Expected behavior:** Only a single sound (or maybe single event) should be fired on click

**Android API version:** Only tried on Android 10

**Material Library version:** 1.2.0-alpha05

collinbus:
#### [TextInputLayout] DropDownMenuEndIconDelegate

By removing the performClick from the EditText's OnClickListener. There won't be any click performed as it is already happening when you click it yourself. Android takes care of the click sound and the performClick call caused it to happen the second time.

closes #1080

### DIFF
--- a/lib/java/com/google/android/material/textfield/DropdownMenuEndIconDelegate.java
+++ b/lib/java/com/google/android/material/textfield/DropdownMenuEndIconDelegate.java
@@ -349,7 +349,6 @@ class DropdownMenuEndIconDelegate extends EndIconDelegate {
                 dropdownPopupDirty = false;
               }
               showHideDropdown(editText);
-              v.performClick();
             }
             return false;
           }


collinbus:
@googlebot I signed it!

leticiarossi:
Thanks for the PR! I already have an internal fix for this issue waiting to be synced out here to github, so will close this.

PR #1124, opened 2020-03-19
[Motion] Fix some code formatting issues in documentation

ansman:
* There was additional spaces in multiple places
* Some quotation marks where `“` instead of `"`
* Several lines had `;` which is not needed
* `findViewById` requires you to specify the type of view now
* The bundle argument in `Activity.onCreate` needs to be nullable and named `savedInstanceState`


ansman:
The build seemed to have failed for an unrelated reason

alfian8787:
LICENSE

PR #1163, opened 2020-03-28
[TextInputLayout] Do not overlay the layout hint and the edit hint

zx2c4:
Everybody wants to have a label for a edit text box. But some people
also want to have some default gray value shown when the box doesn't
have anything put in there. This used to be called the "hint". But now,
the "hint" is actually used as a label. Sort of:

If the hint is set on the TextInputEditText, it operates like a normal
hint, being visible when there's no text inside. If a hint is set on the
TextInputLayout, however, it operates like a label, unless the
TextInputEditText has no text, in which case, it operates like a hint.
The result is that one winds up with two hints, one on top of the other
in an illegible mess:

image

I'm currently hacking around this by subclassing TextInputEditText with
this horror:

```kt
    @Override
    override fun getText(): Editable? {
        val text = super.getText()
        if (!text.isNullOrEmpty())
            return text
        /* We want this expression in TextInputLayout.java to be true if there's a hint set:
         *        final boolean hasText = editText != null && !TextUtils.isEmpty(editText.getText());
         * But for everyone else it should return the real value, so we check the caller.
         */
        if (!hint.isNullOrEmpty() && Thread.currentThread().stackTrace[3].className == TextInputLayout::class.qualifiedName)
            return SpannableStringBuilder(hint)
        return text
    }
```

It doesn't get worse than that. This commit fixes the issue by avoiding
the expanded hint in the case that the enclosing EditText already has a
hint. It works:

image

### DIFF
--- a/lib/java/com/google/android/material/textfield/TextInputLayout.java
+++ b/lib/java/com/google/android/material/textfield/TextInputLayout.java
@@ -1301,7 +1301,7 @@ public class TextInputLayout extends LinearLayout {
 
   private void updateLabelState(boolean animate, boolean force) {
     final boolean isEnabled = isEnabled();
-    final boolean hasText = editText != null && !TextUtils.isEmpty(editText.getText());
+    final boolean hasText = editText != null && (!TextUtils.isEmpty(editText.getText()) || !TextUtils.isEmpty(editText.getHint()));
     final boolean hasFocus = editText != null && editText.hasFocus();
     final boolean errorShouldBeShown = indicatorViewController.errorShouldBeShown();
 


leticiarossi:
I think this is similar to https://github.com/material-components/material-components-android/issues/810

Can't you just use the hint on the `TextInputLayout` along with the `placeholderText` (https://github.com/material-components/material-components-android/issues/810#issuecomment-592543524)?

zx2c4:
`android:hint` is a standard attribute. This patch makes the desired behavior trivially supported by standard attributes. Why wouldn't you want that?

zx2c4:
And actually, @leticiarossi, the behavior is different. placeholderText only shows when the control is selected, whereas a proper hint (this PR) always shows. Please merge this bug fix.

zx2c4:
The thing is that what you just mentioned is trivially possible with this commit. This commit doesn't change any existing behavior at all. It only fixes existing brokenness (see <https://user-images.githubusercontent.com/10643/77832716-b097da00-70fd-11ea-809e-3c6928c66b62.png>), and replaces it with something that the user clearly meant with the XML (see <https://user-images.githubusercontent.com/10643/77832735-c86f5e00-70fd-11ea-8cdd-3b1b9eb86c5b.png>). Fixing that bug doesn't impact any other behavior and doesn't clash with your plans for anything else. This is simply a bug fix, not something that needs design review.

ivanbartsov:
For what it's worth, I also vote for this fix to be merged, kudos @zx2c4 .

C2H6O:
Can we get this merged in? Sometimes you want to have the title of the view to be different from the hint and the hint to be shown at all times.

zx2c4:
Indeed, months later and this is still not merged. Wondering what's up?

alshamiri5 (approved):

Issue #1235, opened 2020-04-22
[TextInputLayout] Icon is not using its own state to update the color


rewgoes:
Side-effect of b17d042f

**Description:** neither `errorIconView`, `endIconView` and `startIconView` are using their states when updating their color.

**Expected behavior:** icons should be able to update their state if they are checked, focused, hovered, activated, etc.

**Source code:**
Create a color selector with state_checked:
```xml
<selector xmlns:android="http://schemas.android.com/apk/res/android">
  <item android:color="?attr/colorPrimary" android:state_checked="true"/>
  <item android:alpha="0.38" android:color="?attr/colorOnSurface" android:state_enabled="false"/>
  <item android:alpha="0.54" android:color="?attr/colorOnSurface"/>
</selector>
```

The icon does not change it's color when checked, as it currently only uses the TextInputLayout state in `updateIconColorOnState`.

**Android API version:** any (tested with 29 and 24)

**Material Library version:** 1.2.0-alpha06 (tested with master branch)

**Device:** any (tested with S10+ and Nexus 6)

rewgoes:
@leticiarossi, could you please check bug that I've found?

PR #1238, opened 2020-04-22
[TextInputLayout] set endIconView as active when TextInputEditText gain focus

### LINKED ISSUE #879: [TextInputLayout] TextInputLayout not getting focus
**Description:** Seem like TextInputLayout is not getting focus when typeing text (The TextInputEditText does getting focused just TextInputLayout doesn't).
This problem is preventing from colorstatelist for icon tint for example to be ignored (color under state_focused is not getting called because TextInputLayout never gains focus).

**Expected behavior:** TextInputLayout should be focused (or maybe set some other state to be true like activated etc..) when typing to edit text

**Source code:** The code snippet which is causing this issue. Please consider attaching a minimal sample app that reproduces the issue.

**Android API version:** 28

**Material Library version:** 1.1.0-bet01

rewgoes:
closes #879 

This feature allows the endIconView to be set as activated when TextInputEditText gains focus.

The feature is blocked by #1235 (186c8f0786b57c406694501389ba31c91bc128e2)

### DIFF
--- a/lib/java/com/google/android/material/textfield/TextInputEditText.java
+++ b/lib/java/com/google/android/material/textfield/TextInputEditText.java
@@ -223,4 +223,13 @@ public class TextInputEditText extends AppCompatEditText {
       return "";
     }
   }
+
+  @Override
+  protected void onFocusChanged(boolean focused, int direction, Rect previouslyFocusedRect) {
+    super.onFocusChanged(focused, direction, previouslyFocusedRect);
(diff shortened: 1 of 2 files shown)


leticiarossi:
Thanks for the PR! I don't think it makes sense to make the icon activated when the text field gains focus. Once your other PR is merged in, that issue in #879 with the focused color of the icon tint being ignored will be solved, so will be closing this one.

PR #1260, opened 2020-04-29
[MaterialButtonToggleGroup] Single visible button corners

### LINKED ISSUE #1259: [MaterialButtonToggleGroup] Single visible button corners
**Description:** If a `MaterialButtonToggleGroup` contains more than one button but only one is visible it gets the shape as if was the first button in the group, even though it's the only (visible).

**Expected behavior:** The single visible button should have its original shape.

**Source code:** The issue seems to be happening in `getNewCornerData(...)`, where we don't consider that child views can be gone.

**Material Library version:** Original found in `1.1.0`, but it's also occuring in `1.2.0-alpha06`

Anigif:
Closes #1259.

### DIFF
--- a/lib/java/com/google/android/material/button/MaterialButtonToggleGroup.java
+++ b/lib/java/com/google/android/material/button/MaterialButtonToggleGroup.java
@@ -662,6 +662,8 @@ public class MaterialButtonToggleGroup extends LinearLayout {
       int firstVisibleChildIndex,
       int lastVisibleChildIndex) {
     CornerData cornerData = originalCornerData.get(index);
+	
+	// If only one (visible) child exists, use its original corners
     if (firstVisibleChildIndex == lastVisibleChildIndex) {
       return cornerData;
     }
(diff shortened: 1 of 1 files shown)


lib/java/com/google/android/material/button/MaterialButtonToggleGroup.java
```diff
@@ -661,9 +661,8 @@ private CornerData getNewCornerData(
       int index,
       int firstVisibleChildIndex,
       int lastVisibleChildIndex) {
-    int childCount = getChildCount();
     CornerData cornerData = originalCornerData.get(index);
-    if (childCount == 1 || firstVisibleChildIndex == lastVisibleChildIndex) {
```
ymarian: maybe add a comment that this condition means "only one visible child"

Anigif: I've added a comment, hope it makes sense

Issue #78, opened 2018-05-01
Add new `Dividers` component guidance


sfdexter:
https://material.io/go/design-dividers

garretyoder:
Is this widget supposed to use onSurface or onBackground as it's default background color by default? The material guidelines https://material.io/design/components/dividers.html are unclear on this. I can implement this since it's a quick few lines of code. Also, is this widget intended to have any intrinsic padding? Again, the guidelines left this out.

ldjcmu:
Closing this issue as we have no plans to implement this in the library directly. For a divider, you should use a color depending upon the color of the container it is on top of.

leticiarossi:
The divider component has been added to the library https://github.com/material-components/material-components-android/commit/782dffa8b719ff4129590380896d2619531906dc

PR #1320, opened 2020-05-23
[Divider] Added a full-bleed divider UI component

ethanhsuhsu:
> Created a full-bleed divider that follows the Material spec and has its
> color supplied by a Material theme.
> 
> Though this is the "base implementation" of a Material Divider, already
> includes most of the needed features:
> 
> * Proper divider height (based on the Material Spec)
> * Proper color transparency (based on the Material Spec)
> * Adaptable coloring from a Material theme
> * Dark theme support (from Material Theming)
> * Customizable attributes; the height, color and width can be changed

When creating Material lists, I found that it helped to split up its needed parts into "subcomponents." However, I later realized that these "subcomponents" could also individually be used internally and externally by developers. For the repository, a divider component makes sense. Infact, MDC-Android even includes small divider implementations!

Back in #78, dividers were mainly shrugged off as being _simple to implement_. However, with the introduction of dark mode and much confusion around implementing dividers ([stackoverflow](https://stackoverflow.com/questions/5049852/android-drawing-separator-divider-line-in-layout) 
 has 20+ answers and [this helpful divider guide](https://medium.com/@szholdiyarov/how-to-add-divider-to-list-and-recycler-views-858344450401) has multiple pages) the need for developer guidance on dividers is growing. 

In addition, I think working on a smaller component will help establish an open-source workflow, since the process of adding new-components through PRs has never been discussed before.

In the base implementation of dividers, I added support for a traditional full-bleed divider. I had the divider extend View and be configured to let developers add it through XML. Everything is documented, and all necessary attributes are annotated.

I think by adding dividers, Material Components Android will be more compelling solution for UIs.

In future contributions, I plan on improving this component and making it production ready:
* In the next contribution, I am going to add inset divider support to this component. The divider will only display a line up to a certain View or distance.
* Public and private API documentation, to make understanding the component easier.
* A catalog app example of using Material dividers
* UI tests to allow Dividers to be production ready

If any of these features should be added into this PR, I am happy to add them in 👍 

@dsn5ft @leticiarossi

### DIFF
--- a/lib/java/com/google/android/material/divider/res/values/attrs.xml
+++ b/lib/java/com/google/android/material/divider/res/values/attrs.xml
@@ -1,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+  ~ Copyright (C) 2020 The Android Open Source Project
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~
+  ~      http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+-->
+<resources>
+
+  <!-- Style to use for Dividers in this theme, usually to be used as a view. -->
+  <attr name="dividerStyle" format="reference" />
+  <attr name="dividerAlpha" format="float" />
+  <attr name="dividerSize" format="dimension"/>
+  <attr name="dividerColor" format="color"/>
+  <attr name="dividerInset" format="dimension"/>
+</resources>
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -48,6 +48,7 @@ def srcDirs = [
   'com/google/android/material/color',
   'com/google/android/material/datepicker',
   'com/google/android/material/dialog',
+  'com/google/android/material/divider',
   'com/google/android/material/drawable',
   'com/google/android/material/elevation',
   'com/google/android/material/expandable',
--- a/lib/java/com/google/android/material/divider/MaterialDivider.java
+++ b/lib/java/com/google/android/material/divider/MaterialDivider.java
@@ -1,0 +1,193 @@
+/*
+ * Copyright 2020 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.android.material.divider;
+
+import android.content.Context;
+import android.content.res.TypedArray;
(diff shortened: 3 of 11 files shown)


ethanhsuhsu:
Check out the component's design doc for more details and a full roadmap: https://docs.google.com/document/d/1an2ZJhco2om-6hSMI2W1Py7I8RmCOwiJ7u1znF7u4Ls/edit?usp=sharing

lib/java/com/google/android/material/divider/res/values/attrs.xml
```diff
@@ -0,0 +1,24 @@
...
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+-->
+<resources>
+
+  <!-- Style to use for Dividers in this theme, usually to be used as a view. -->
+  <attr name="dividerStyle" format="reference" />
+
+  <declare-styleable name="Divider" />
```
ricknout, line 22: Could this styleable include `<attr name="thickness" format="dimension" />` which could then be parsed/applied in the `MaterialDivider` class instead of using the hardcoded dimen? I also think the general name should change from "thickness" to "height"

ethanhsuhsu: @ricknout I originally allowed for this behavior, but then realized that developers are required to define the `android:layout_height` attribute. It would be confusing to have two different height attributes, it is simpler to just have a multi-purpose `android:layout_height`:

* If the developer wants to use Material's default divider height, they set the divider height to `WRAP_CONTENT`
* If the developer wants to customoize the height, they can set `android:layout_height` to a custom height.

ethanhsuhsu:
@ricknout Though a drawable-based implementation would be more friendly for us to maintain, it retains much of the confusion for developers. Here is a comparison between a view-based approach and a drawable-based approach:

| Criteria | Drawable approach | View approach |
|---------|:--------------------:|:----------------:|
| Simple implementation | ✅ | ✅ |
| Material theming | ✅ | ✅ |
| Attribute modifications | ❌ | ✅ |
| View-based insets | ❌ | ✅ |
| Customization |  ❌ | ✅ |

The main limitation of a drawable that is used in a RecyclerView is that it cannot be easily edited after-the-fact. A view-based approach would allow developers to have a reference to their divider to implement custom logic. Performing the same on an ItemDecoration means setting a new ItemDecoration, which is an expensive operation. Using a View also future-proofs this component, as future needs could be more easily met with a custom View. 

I understand that the final API may not be my decision, but I believe a drawable approach could disable some of a divider's functionality. However, if the last three points do not matter, than I would rate your implementation to be "cleaner".

lib/java/com/google/android/material/divider/res/values/attrs.xml
ricknout: Fair enough. Although I still find "thickness" to be a word that is not well known on Android in general. Perhaps "size" would be better?

ricknout:
I agree with your points. I see the view approach as most useful when using dividers directly in layouts. However, arguably one of the most common place a divider is used - in a RecyclerView - requires the drawable / item decoration approach. We can't exclude that use case when providing material dividers.

Having thought about this a bit more, to cover all bases I would suggest:
- A base drawable that can be used in all of the scenarios I previously mentioned
- **To be used in layouts:** A `MaterialDivider` view class which _uses_ this drawable as its background (along with additional customization attributes like insets, tint, etc. applied on top of this).
- **To be used with RecyclerView:** An additional `MaterialDividerItemDecoration` class which, like the view, uses the drawable and supports all of the attributes the `MaterialDivider` view does (you can see how DividerItemDecoration parses attrs in the [source](https://android.googlesource.com/platform/frameworks/support/+/androidx-master-dev/recyclerview/recyclerview/src/main/java/androidx/recyclerview/widget/DividerItemDecoration.java#67))

lib/java/com/google/android/material/divider/res/values/attrs.xml
ethanhsuhsu: Yep, that seems better

ethanhsuhsu:
Closed due to more recent updates...

PR #1325, opened 2020-05-25
[Shape] Update ShapeAppearanceModel to use AndroidX @Px Annotations

ethanhsuhsu:
> Allowed `cornerSize` parameters to be labeled as pixel dimensions to help better specify the input. The previously used @Dimension annotation is more vague, and does not inform developers on what dimension type they need to pass in.

When recently using Material Shape Theming, I was unsure of whether it allowed developers to pass in dp or pixel dimensions. Unfortunately, I ended up having to manually test to find out the result. By being more clear with annotations, other developers will not have to run into this issue ;)


ymarian:
For the record you can check that the Dimension annotation has a value parameter that defaults to PX.

PR #1328, opened 2020-05-26
[Shape] Allowed ShapeableImageView to dynamically update

### LINKED ISSUE #1327: [Shape] ShapeableImageView does not update ShapeAppearanceModel
**Description:** In certain situations, calling setShapeAppearanceModel on a ShapeableImageView does not update the edge and corner treatments. This happens as `requestLayout()` is not very aggressive in updating the View.

For example, calling setShapeAppearanceModel in the onBind method of a RecyclerView does not update directly to the screen.

[Video example](https://drive.google.com/file/d/1OzxxzyXY-aDUnfiLlhwxk-cmjj5WikkE/view?usp=sharing)

**Android API version:** Affects all versions of Android. Tested on API 29 (Android 10)

**Material Library version:** Latest - 1.2.0-alpha06

**Device:** Pixel 3

ethanhsuhsu:
closes #1327 

>After the ShapeAppearanceModel is changed, ShapeableImageView will now automatically redraw itself to use the latest ShapeAppearanceModel when drawing edges and corners.

When updating the ShapeAppearanceModel of a ShapeableImageView, the View sometimes does not update with the latest changes. This issue is an unconsidered corner case, however ShapeableImageView should be able to support updates in any part of the View rendering lifecycle.

### DIFF
--- a/lib/java/com/google/android/material/imageview/ShapeableImageView.java
+++ b/lib/java/com/google/android/material/imageview/ShapeableImageView.java
@@ -159,7 +159,12 @@ public class ShapeableImageView extends AppCompatImageView implements Shapeable 
   @Override
   public void setShapeAppearanceModel(@NonNull ShapeAppearanceModel shapeAppearanceModel) {
     this.shapeAppearanceModel = shapeAppearanceModel;
-    requestLayout();
+    updateShapeAppearanceModel();
+  }
+
+  private void updateShapeAppearanceModel(){
+    this.onSizeChanged(getWidth(), getHeight(), getWidth(), getHeight());
+    invalidate();
   }
 
   @NonNull


lib/java/com/google/android/material/imageview/ShapeableImageView.java
```diff
@@ -159,7 +159,8 @@ private void drawStroke(Canvas canvas) {
   @Override
   public void setShapeAppearanceModel(@NonNull ShapeAppearanceModel shapeAppearanceModel) {
     this.shapeAppearanceModel = shapeAppearanceModel;
-    requestLayout();
+    /*Refreshing this View to apply the newest shapeAppearanceModel */
+    this.onSizeChanged(getWidth(),getHeight(),getWidth(),getHeight());
```
ymarian, line 163: Thanks for the PR!

You still need to call requestLayout to redraw. You can change it to invalidate now.
onSizeChanged won't force a redraw.  Please do the following:

- Remove comment
- Extract logic from onSizeChanged to an `updateShapeAppearance(int width, int height)` method. 
- Call that method here before invalidate and on `onSizeChanged`.
- Watch the formatting (spaces after commas)

ethanhsuhsu: Hi @ymarian, I apologize if I misunderstood your comments. I moved the "refresh" part of `setShapeAppearanceModel` to a new `updateShapeAppearanceModel` and ran `updateShape...` from `setShape...`. I agree that the update logic should be decoupled from the setter. Thanks for the review 😉

ethanhsuhsu:
@ymarian, I updated my commit in accordance with your change requests. Can you approve my changes?

lib/java/com/google/android/material/imageview/ShapeableImageView.java
ymarian: Oh yeah I was saying something different, no worries. I'll make the changes and submit this. Thanks for finding the bug!

ethanhsuhsu: Cool, I just needed this bug fixed for an upcoming Lists PR...

ethanhsuhsu: I am going to leave this PR open just in case I encounter this issue again after you submit the patch. My dev environment is very specific.

PR #1367, opened 2020-06-03
[Lists] Added adaptable list items elements

ethanhsuhsu:
> [Lists] Added adaptable list items elements
> 
> Modified the existing components to automatically adjust their
> visibilities based on their internal contents. For example, the
> AdaptableTextView becomes visible if it has text to display, and becomes
> gone if has no text to display ("").
> 
> This eases the public API as developers will no longer have to call
> setVisibility() to enable list item elements before using them. In a
> kind of reactive manner, list item elements can be enabled and bound
> to data in one step.
> 
> Example:
> 
> In the new API:
> 
> `item.getPrimaryText().setText("My Text")` 
> 
> Is equivalent to this in the old API:
> 
> `item.getPrimaryText().setVisibility(VISIBLE)`
> `item.getPrimaryText().setText("My Text")`

Extension of #1350 and contingent on #1350 

Through adaptable elements, developers can simply set the contents of the list item without having to worry about View visibility and enabling/disabling elements. When the developers sets the image of a `AdaptableShapeableImageView` or sets the text of a `AdaptableTextView`, it will automatically set itself to be `View.VISIBLE` and display the contents. This dramatically simplifies developer interaction with MaterialListItem.
API Usage:

```java
@Override
    public MaterialListItem onCreateViewHolder(@NonNull ViewGroup parent, int viewType) {
        
        //Initialize new list item
        MaterialListItem item = new MaterialListItem(parent);
    }

@Override
    public void onBindViewHolder(@NonNull MaterialListItem item, int position) {
        
        //Customize the contents of the list item based on data and the items will automatically become visible!
        item.getVisual().setImageDrawable(...);
        item.getPrimaryText().setText(...);
}
``` 

Depending on the course of action, the majority of future action should be taken on #1350. This pull request helps show the potential use cases for Material Lists, and how it can simplify UI development for Android Developers. @ricknout, this is a cleaner list implementation and should be easier to teach others about.

### DIFF
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -54,6 +54,7 @@ def srcDirs = [
   'com/google/android/material/floatingactionbutton',
   'com/google/android/material/imageview',
   'com/google/android/material/internal',
+  'com/google/android/material/lists',
   'com/google/android/material/math',
   'com/google/android/material/menu',
   'com/google/android/material/navigation',
--- a/lib/java/com/google/android/material/lists/item/MaterialListItem.java
+++ b/lib/java/com/google/android/material/lists/item/MaterialListItem.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright 2020 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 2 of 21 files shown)


ethanhsuhsu:
The progress in this PR will be instead moved into #1350...

Issue #1373, opened 2020-06-04
bottomSheet behavior in android 10 the peekheight is messed up because of gestureInsetBottomIgnored by default is false


rafikFares:
as since version 1.2.0-alpha06 and above the feature of gestureInsetBottomIgnored  was made, and set false by default and only activate for phones with android 10+, so when testing it with default "false" in my onePlus with android 10 the peekheight is messed up

Expected behavior: peekHeight of only 10dp or whatever value i gave it

Source code: app:gestureInsetBottomIgnored="false" or doing nothing because by default is false

Android API version: Android API 10+

Material Library version: 1.2.0-alpha06+

Device: OnePlus 6 with android 10 and OnePlus 7 with android 10

to fix it i had to make app:gestureInsetBottomIgnored="true" but not much people know that this attribute exist so it may cause them a weird behavior in their app when they will update their dependency version

link to the commit of the feature : https://github.com/material-components/material-components-android/commit/4e28d9c9162d8f4c73dfa1b397f52ef5e32ed726

dsn5ft:
This shouldn't be an issue anymore as we've improved the peek height gesture inset behavior to only add the additional inset when it's needed (https://github.com/material-components/material-components-android/commit/7b620035a78f15284aed82ec6c09d5f6eb4249dc).

PR #1375, opened 2020-06-04
make gestureInsetBottomIgnored true by default to fix peekHeight in a…

rafikFares:
Solve : https://github.com/material-components/material-components-android/issues/1373


dsn5ft:
This shouldn't be an issue anymore as we've improved the peek height gesture inset behavior to only add the additional inset when it's needed (https://github.com/material-components/material-components-android/commit/7b620035a78f15284aed82ec6c09d5f6eb4249dc).

PR #1391, opened 2020-06-08
[ProgressIndicator] Fix for non dismissable progress indicator

### LINKED ISSUE #1382: [ProgressBarIndicator] Don't hide if calling show/hide is too fast
**Material design version 1.3.0-alpha01**
**Test device: Galaxy S10+ Android 10, Galaxy J5 Android 6.0.1**

ProgressBarIndicator didn't hide if I called show/hide too fast. I need to call in next time to hide it or call immediately **set progressBarIndicator.isGone = true** but it will not have animation or **delay(1second)** before progressBarIndicator.hide(). It is the same the case when I use ContentLoadingProgressBar.

```kotlin
launch {
     progressBarIndicator.show()

     val videos = videoModel.getVideos()
     videosAdapter.addData(videos.toMutableList())

     //delay(1000L)
     progressBarIndicator.hide() // don't hide if time between show and hide not enough long
 }

```

```xml
<com.google.android.material.appbar.AppBarLayout
            android:id="@+id/appBar"
            app:elevation="0dp">

            <FrameLayout
                app:layout_scrollFlags="scroll|enterAlways">

                <com.google.android.material.appbar.MaterialToolbar
                    android:id="@+id/toolbar"
                    android:layout_height="?attr/actionBarSize"
                    android:background="?themePrimary"
                    app:title="@string/videos" />

                <com.google.android.material.progressindicator.ProgressIndicator
                    android:id="@+id/progressBar"
                    style="@style/Widget.MaterialComponents.ProgressIndicator.Linear.Indeterminate"
                    android:layout_gravity="bottom"
                    android:visibility="gone"
                    app:growMode="outgoing"
                    app:indicatorColor="?colorSecondary"
                    tools:visibility="visible"/>

            </FrameLayout>

</com.google.android.material.appbar.AppBarLayout>
```

matpag:
This should fix the problem with `ProgressIndicator` , that prevent them to dismiss if `hide()` is called too quickly.

We need to evaluate the animation state every time, so I removed the check which stops logic when animations are running.

This was reproducible on catalog app fast clicking "Show" and "Hide" in the "Indeterminate Progress Indicator Demo" page.
With the fix applied, the progress indicator are dismissed correctly.

Let me know if this is ok for you.

Closes #1382

### DIFF
--- a/lib/java/com/google/android/material/progressindicator/DrawableWithAnimatedVisibilityChange.java
+++ b/lib/java/com/google/android/material/progressindicator/DrawableWithAnimatedVisibilityChange.java
@@ -206,11 +206,6 @@ abstract class DrawableWithAnimatedVisibilityChange extends Drawable implements 
     boolean shouldAnimate =
         animationDesired && progressIndicator.getGrowMode() != ProgressIndicator.GROW_MODE_NONE;
 
-    // We don't want to change visibility while show/hide animation is running.
-    if (showAnimator.isRunning() || hideAnimator.isRunning()) {
-      return false;
-    }
-
     // Cancels any running animations.
     showAnimator.cancel();
     hideAnimator.cancel();


ymarian:
Thanks for working on this! Seems to be an improvement but hide() quickly after show() still doesn't work, tested in the catalog app. I want to dig a bit deeper to get the right solution.

matpag:
In the class `ProgressIndicatorIndeterminateDemoFragment`, I've tested using this show button callback in the initialize method:
```
showButton.setOnClickListener(
	v -> {
	  for (ProgressIndicator progressIndicator : indicatorList) {
		resetProgressIndicator(progressIndicator);
		progressIndicator.show();
		progressIndicator.hide();
	  }
	});
```
I thought this was supposed the minimum call time between show and hide.
I've tested again now and it seems to work.

pekingme:
Hi @matpag, thanks for being active on this new component. The conditional statement you removed in this commit is based on the consideration of the animation continuity. Meaning if the component is showing halfway with a grow animation and started to hide, it won't immediately fully shown and start to play the hide animation. Will consult with a design if this is ok.

matpag:
@pekingme You are right, I proceeded in this way because to me the animation is still great, even if the indicator starts hiding when not fully shown.

If someone doesn't want that, in the linked issue I proposed to add a new delay property to the progress indicator component so that the developers can add a "minimum hide delay" to customize this behaviour. (I can work on that too when this is fixed).

Thank you for taking the time to respond.

pekingme:
Sorry, I missed the original issue thread. That’s a good idea. Let me double check if this commit affects other things before approving it.

PR #1395, opened 2020-06-09
[Slider] Keyboard navigation fix, part 2

### LINKED ISSUE #1381: [Slider] Keyboard navigation is broken in RTL
Requires #1364. Working on a PR offline.

**Description:**

##### Focus handling
Before #1364 `RangeSlider` didn't handle passing keyboard focus correctly at all. Now it doesn't handle passing keyboard focus only in RTL - the sides are reversed.

##### Value changing
Right now, Plus key decrements and Minus key increments in RTL. **Fixing this will change behavior from the original `AbsSeekBar`**, which is also broken from this point of view.

**Expected behavior:**
Pressing Plus key increments the value of active thumb (moves the thumb left) in RTL.
Pressing Minus key decrements the value of active thumb (moves the thumb right) in RTL.
Pressing Left arrow key moves focus to left (either to another thumb or another view).
Pressing Right arrow key moves focus to right (either to another thumb or another view).

**Source code:**
These methods require changing, at minimum:
* `BaseSlider#onKeyDown`
* `BaseSlider#calculateIncrementForKey`
* `BaseSlider#focusThumbOnFocusGained`

**Android API version:** Any

**Material Library version:** 1.2.0-alpha01

**Device:** Pixel 3a Android R AVD with Google Play Intel x86 Atom_64 System Image r4

consp1racy:
Closes #1381.

Does anybody know how to test RTL in Robolectric?

It runs with `targetSdkVersion` 14 (it's unset so it takes it from `minSdkVersion`) and the only fix is having `targetSdkVersion` 17+ in build.gradle (RTL is API 17+ only) so Robolectric picks it up. Then I tried setting `@Config(qualifiers="+ldrtl)`. Then I tried setting default locale to `ar_XB` and updating system configuration with it (mimic `RuntimeEnvironment.setQualifiers(...)`). Then I tried manually setting RTL on the slider. None of this worked.

### DIFF
--- a/lib/java/com/google/android/material/slider/BaseSlider.java
+++ b/lib/java/com/google/android/material/slider/BaseSlider.java
@@ -1362,7 +1362,7 @@ abstract class BaseSlider<
     float right = normalizeValue(max);
 
     // In RTL we draw things in reverse, so swap the left and right range values
-    if (ViewCompat.getLayoutDirection(this) == ViewCompat.LAYOUT_DIRECTION_RTL) {
+    if (isRtl()) {
       return new float[] {right, left};
     } else {
       return new float[] {left, right};
@@ -1389,7 +1389,7 @@ abstract class BaseSlider<
    */
   private float normalizeValue(float value) {
     float normalized = (value - valueFrom) / (valueTo - valueFrom);
-    if (ViewCompat.getLayoutDirection(this) == ViewCompat.LAYOUT_DIRECTION_RTL) {
+    if (isRtl()) {
       return 1 - normalized;
     }
     return normalized;
@@ -1660,7 +1660,7 @@ abstract class BaseSlider<
     double position = snapPosition(touchPosition);
 
     // We might need to invert the touch position to get the correct value.
-    if (ViewCompat.getLayoutDirection(this) == ViewCompat.LAYOUT_DIRECTION_RTL) {
+    if (isRtl()) {
       position = 1 - position;
     }
     return (float) (position * (valueTo - valueFrom) + valueFrom);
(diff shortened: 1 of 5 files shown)


ymarian:
RE: testing rtl, You probably just need to recreate the slider or trigger a configuration change in the activity since it's already created before setting the qualifiers. 

On a side note it'd be awesome if you want to migrate this test to the new ActivityScenario API.

consp1racy:
For `ActivityScenario` the activity (here `AppCompatActivity`) needs to be declared in the manifest with a launcher intent filter. There's a million manifests in MDC, I wish I knew where the java test manifest for slider goes. Can Robolectric even can pick those up? Anyway, the ones that are already there (e.g. `javatests/com/google/android/material/slider/AndroidManifest.xml`) get ignored.

    java.lang.RuntimeException: Unable to resolve activity for Intent { act=android.intent.action.MAIN cat=[android.intent.category.LAUNCHER] cmp=com.google.android.material.test/androidx.appcompat.app.AppCompatActivity } -- see https://github.com/robolectric/robolectric/pull/4736 for details
        at org.robolectric.android.fakes.RoboMonitoringInstrumentation.startActivitySyncInternal(RoboMonitoringInstrumentation.java:48)
        at org.robolectric.android.internal.LocalActivityInvoker.startActivity(LocalActivityInvoker.java:34)
        at androidx.test.core.app.ActivityScenario.launch(ActivityScenario.java:207)

Shuffling around MDC internals is way beyond the scope.

lib/build.gradle
```diff
@@ -102,6 +102,8 @@ android {
 
   defaultConfig {
     minSdkVersion 14
+    //noinspection ExpiredTargetSdkVersion
+    targetSdkVersion 17
```
ymarian, line 106: I'm not super comfortable making these changes in the PR. We would have to check internally if there is any workaround or ramification from this. I think we'll probably need to create a separate test module

Can you add `@Ignore("Fix RTL support for robolectric tests")` To the new tests? We can investigate this internally! Thanks so much for working on that!

consp1racy:
wth image

consp1racy:
@ymarian Thanks, that's great news.

I noticed the `@Ignore` got moved from the individual RTL tests to the whole class. I had to put the annotation on each method on purpose because otherwise the tests would still run on my machine - the class annotation was ignored by the runner.

ymarian:
@consp1racy hmm it worked for and the build seems to be passing here:

https://travis-ci.org/github/material-components/material-components-android/jobs/699796539

Do you still have issues locally? I can roll that back

consp1racy:
@ymarian The whole thing blows up because the min sdk version is not 17, so annotating only the RTL methods wouldn't work.
obrazek

Once Robolectric to picks up test manifests this can easily be fixed, let's just do that.

PR #1408, opened 2020-06-16
[CollapsingToolbarLayout] Slight misalignment during transition in multiline mode

### LINKED ISSUE #1052: [CollapsingToolbarLayout] Slight misalignment during transition in multiline mode
**Description:** This affects the multiline mode of the CollapsingToolbarLayout that was added as an experimental feature in 1.2.0-alpha05 (see also: #413). Upon closer inspection of the new implementation, I found that there is a slight misalignment in the positions of the expanded text and the overlays drawn during the transition, as seen in the screenshot below. Also, when the text snaps to the completely collapsed state, there is a slight upwards jump of the text position. Both of these problems can be seen in the catalog app and were not present in my original version in #413.

Screenshot_1582471001

**Expected behavior:** Smooth transition without misalignments. This worked correctly in our original PR:

Demo image

**Source code:** see catalog app -> Top App Bar -> Collapsing Demo (Multiline title)

**Android API version:** 28

**Material Library version:** 1.2.0-alpha05

**Device:** Android Emulator

I haven't found the source of the issue yet, but it should be possible to find when comparing my implementation in #413 with the one added by @ymarian in 83e4e47d2225cd281bef195d1baf035117cda4cd.

brudaswen:
Closes #1052

As explained in #1052 the alignment of the collapsing text was not correct. This became even more prominent if the textSize between expanded and collapsed became bigger.

## Before
image

## After
image

### DIFF
--- a/lib/java/com/google/android/material/internal/CollapsingTextHelper.java
+++ b/lib/java/com/google/android/material/internal/CollapsingTextHelper.java
@@ -713,7 +713,7 @@ public final class CollapsingTextHelper {
     // Collapsed text
     textPaint.setAlpha((int) (collapsedTextBlend * originalAlpha));
     canvas.drawText(
-        textToDrawCollapsed, 0, textToDrawCollapsed.length(), 0, -ascent / scale, textPaint);
+        textToDrawCollapsed, 0, textToDrawCollapsed.length(), 0, -ascent, textPaint);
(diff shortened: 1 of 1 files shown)


brudaswen:
@googlebot I signed it!

JorisBodin:
Perfect, thank you :D

PR #1432, opened 2020-06-22
[CollapsingToolbarLayout] Use correct layout to calculate expanded bounds

SonOfBowser:
Fix an issue introduced in 83e4e47d2225cd281bef195d1baf035117cda4cd
which affected multiline collapsing toolbars
when the expanded size was much larger or smaller
than the collapsed size.
The stored `textLayoutHeight` is only used to
calculate the expanded bounds but is generated
after the call to `calculateUsingTextSize(collapsedTextSize)`
which updates `textLayout` to the collapsed size.
This change just moves it to after
`calculateUsingTextSize(expandedTextSize)` so
`textLayout` is using the expanded size.

Note this isn't very visible in the catalog app
because the text sizes are similar but you'll
notice after applying this that the bottom
margin now applies correctly. It's more noticeable
if you set the expanded text size to one of the
larger sizes.


SonOfBowser:
@googlebot I signed it!

Issue #1234, opened 2020-04-22
[MaterialAlertDialog] Button theming broken on 1.2.0-alpha06 for Bridge theme


rlatapy-luna:
**Description:** Full description of issue here
AlertDialog theming seems broken in 1.2.0-alpha06 with Bridge theme.

**Expected behavior:**
|alpha05 + bridge|alpha06 + bridge|alpha06 (expected behavior)|
|:--:|:--:|:--:|
|![image](https://user-images.githubusercontent.com/45555889/80015461-0def3300-84d2-11ea-9f98-1ff2b53c0242.png)|![image](https://user-images.githubusercontent.com/45555889/80015336-da140d80-84d1-11ea-881d-d6a330303372.png)|![image](https://user-images.githubusercontent.com/45555889/80015368-e8fac000-84d1-11ea-9b3f-84289447aa24.png)|

With `backgroundTint` I can manage to have the correct color with alpha05 + bridge but not with alpha06 + bridge

**Source code:**  https://github.com/rlatapy-luna/dialogbuttonstyleissue

```
<style name="AppTheme" parent="Theme.MaterialComponents.DayNight.NoActionBar.Bridge">
    <item name="materialAlertDialogTheme">@style/ThemeOverlay.Custom.MaterialAlertDialog</item>
</style>
<style name="ThemeOverlay.Custom.MaterialAlertDialog" parent="ThemeOverlay.MaterialComponents.MaterialAlertDialog">
    <item name="buttonBarPositiveButtonStyle">@style/Widget.MaterialComponents.Button</item>
    <item name="buttonBarNeutralButtonStyle">@style/Widget.MaterialComponents.Button.OutlinedButton</item>
</style>

```
**Android API version:** 29

**Material Library version:** 1.2.0-alpha06

**Device:** Pixel 3XL

gabrielemariotti:
I am not sure that it is a bug.
You are using a `Bridge` theme and it means that each button is a  `AppCompatButton` instead of a `MaterialButton`.

<img width="459" alt="Schermata 2020-08-15 alle 20 54 23" src="https://user-images.githubusercontent.com/2583078/90319839-66908300-df3c-11ea-9a64-e08cd85198c2.png">

You can't use an `OutlinedButton` or a `Filled` style with an `AppCompatButton` since some attributes are not used.

drchen:
It seems we can close the issue. : )

Issue #1201, opened 2020-04-10
[CollapsingToolbarLayout] A title without space isn't displayed correctly


esnaultdev:
**Description:**
![Screenshot_1586525028_small](https://user-images.githubusercontent.com/7100677/78994543-51b17680-7b7b-11ea-92b0-5868ff08c006.png)

Using a title without spaces in a CollapsingToolbarLayout results in the text being laid out below the last line.

**Expected behavior:**
The title should break as it does in the screenshot, but all the lines should be displayed properly. Since the title gravity is bottom and the maxLines 3, in this example two lines should be displayed.

**Source code:**
Using the catalog source code:
Updating `cat_topappbar_collapsing_multiline_fragment.xml` to have bigger margins on the `CollapsingToolbarLayout`:
```
app:expandedTitleMarginStart="72dp"
app:expandedTitleMarginEnd="56dp"
```

Updating the `cat_topappbar_collapsing_multiline_demo_toolbar_title` string value to "Thisisalongtitlewithoutspace".

Resulting files:

`catalog/java/io/material/catalog/topappbar/res/layout/cat_topappbar_collapsing_multiline_fragment.xml`
```
...
<com.google.android.material.appbar.CollapsingToolbarLayout
      android:id="@+id/collapsingtoolbarlayout"
      style="?attr/catalogToolbarStyle"
      android:layout_width="match_parent"
      android:layout_height="match_parent"
      app:collapsedTitleTextAppearance="?attr/textAppearanceHeadline6"
      app:expandedTitleGravity="bottom"
      app:expandedTitleMarginBottom="24dp"
      app:expandedTitleMarginStart="72dp"
      app:expandedTitleMarginEnd="56dp"
      app:expandedTitleTextAppearance="?attr/textAppearanceHeadline5"
      app:maxLines="3"
      app:layout_scrollFlags="scroll|exitUntilCollapsed|snap">
...
 ```

`catalog/java/io/material/catalog/topappbar/res/values/strings.xml`
```
...
<string name="cat_topappbar_collapsing_multiline_demo_toolbar_title" description="Long title [CHAR_LIMIT=NONE]">Thisisalongtitlewithoutspace</string>
...
```

**Android API version:** API 29

**Material Library version:** 1.2.0-alpha05

**Device:** Android Emulator - Pixel 3

drchen:
Looks like it's been fixed.

Issue #1458, opened 2020-07-03
CollapsingToolbarLayout maxlines acting not as expected.


lectricas:
**Description:** Hi. I'm trying to make my CollapsingToolbarLayout to have two lines, which I do with 
`app:maxLines="2"`

There is some margins in expanded title, so I removed it like this:

```
            app:expandedTitleMarginBottom="0dp"
            app:expandedTitleMarginEnd="0dp"
            app:expandedTitleMarginStart="0dp"
            app:expandedTitleMarginTop="0dp"
```

Now my second line is cut out of view. I need to set 
`app:expandedTitleMarginBottom="8dp"`
in order to see the full two line title.
When I remove `app:maxLines="2"` it works as expected. 

It is not clear for me why this happens. Also CollapsingToolbarLayout is very mysterious, consider writing some docs on how this thing works.

<img width="490" alt="Screenshot 2020-07-03 at 15 06 09" src="https://user-images.githubusercontent.com/8782219/86467806-c50ff200-bd3e-11ea-9546-2efb050f85e3.png">

SonOfBowser:
Have you tried this on a snapshot? I think this is the same thing I fixed in https://github.com/material-components/material-components-android/pull/1432

PR #1462, opened 2020-07-04
[CollapsingToolbarLayout] Fix CollapsingToolbar multiline position

ghus-raba:
closes https://github.com/material-components/material-components-android/issues/1201 and https://github.com/material-components/material-components-android/issues/1458

This fixes position of multiline text in expanded state (best visible if bottom margin is set to 0 as in https://github.com/material-components/material-components-android/issues/1458) as well as some jump when transitioning to collapsed state.

### DIFF
--- a/lib/java/com/google/android/material/appbar/CollapsingToolbarLayout.java
+++ b/lib/java/com/google/android/material/appbar/CollapsingToolbarLayout.java
@@ -482,7 +482,7 @@ public class CollapsingToolbarLayout extends FrameLayout {
         collapsingTextHelper.setCollapsedBounds(
             tmpRect.left + (isRtl ? toolbar.getTitleMarginEnd() : toolbar.getTitleMarginStart()),
             tmpRect.top + maxOffset + toolbar.getTitleMarginTop(),
-            tmpRect.right + (isRtl ? toolbar.getTitleMarginStart() : toolbar.getTitleMarginEnd()),
+            tmpRect.right - (isRtl ? toolbar.getTitleMarginStart() : toolbar.getTitleMarginEnd()),
             tmpRect.bottom + maxOffset - toolbar.getTitleMarginBottom());
 
         // Update the expanded bounds
--- a/lib/java/com/google/android/material/internal/CollapsingTextHelper.java
+++ b/lib/java/com/google/android/material/internal/CollapsingTextHelper.java
@@ -548,18 +548,15 @@ public final class CollapsingTextHelper {
 
     switch (collapsedAbsGravity & Gravity.VERTICAL_GRAVITY_MASK) {
       case Gravity.BOTTOM:
-        collapsedDrawY = collapsedBounds.bottom;
+        collapsedDrawY = collapsedBounds.bottom + textPaint.ascent();
         break;
       case Gravity.TOP:
-        collapsedDrawY = collapsedBounds.top - textPaint.ascent();
+        collapsedDrawY = collapsedBounds.top;
         break;
       case Gravity.CENTER_VERTICAL:
       default:
-        float textHeight = textPaint.descent() - textPaint.ascent();
-        float textOffset = (textHeight / 2) - textPaint.descent();
-        collapsedDrawY = shouldDrawMultiline()
-            ? collapsedBounds.centerY() - (textHeight / 2)
-            : collapsedBounds.centerY() + textOffset;
+        float textOffset = (textPaint.descent() - textPaint.ascent()) / 2;
+        collapsedDrawY = collapsedBounds.centerY() - textOffset;
         break;
     }
 
@@ -577,13 +574,12 @@ public final class CollapsingTextHelper {
     }
 
     calculateUsingTextSize(expandedTextSize);
-
-    float textLayoutHeight = textLayout != null ? textLayout.getHeight() : 0;
+    float expandedTextHeight = textLayout != null ? textLayout.getHeight() : 0;
 
     float measuredWidth = textToDraw != null
         ? textPaint.measureText(textToDraw, 0, textToDraw.length()) : 0;
     width = textLayout != null && maxLines > 1 && !isRtl
-        ? textLayout.getLineWidth(0)
+        ? textLayout.getWidth()
         : measuredWidth;
     expandedFirstLineDrawX = textLayout != null ? textLayout.getLineLeft(0) : 0;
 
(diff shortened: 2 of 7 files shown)


ghus-raba:
> You'll have to rebase this on master, we don't take PRs directly to a release branch, and there have been a few fixes to Multiline there. In addition please squash it into one commit.

Done. The main issues seemed to be also fixed in master (though I would swear I checked that before). 

What's left here is some improvement to multiline demo + small fix for computing collapsed bounds. Also I kept some cleanup in collapsing text helper. 

Also I would ask to consider picking these fixes for 1.2 - as it is currently, the multiline feature is unusable for us in 1.2 and I would love to avoid waiting for 1.3 to become stable.

ymarian:
Can you elaborate a bit on the changes and what was broken? 

Things change substantially after this for example the margin bottom was previously from the baseline of the text but now it seems it’s from the text descent. Additionally seems you fixed margin start for RTL? But again a lot of things change from this, even TextFields. So would like to understand some of the changes and how this was tested

ghus-raba:
> Can you elaborate a bit on the changes and what was broken?

Sure. To start off easy, everything under `catalog` folder is just adding a menu to multiline demo that let's us change the value of `maxLines` for collapsing toolbar to 1, 2 or 3 for easier testing.

For the additional changes, I will comment next to the relevant code.

> Additionally seems you fixed margin start for RTL?

Maybe, but not intentionally 😄. I did not check it in RTL.

> how this was tested?

Just manually inside the catalog by using various settings eg. from linked issues. If you know of a way to write tests for this, I would love to.

lib/java/com/google/android/material/appbar/CollapsingToolbarLayout.java
```diff
@@ -482,7 +482,7 @@ protected void onLayout(boolean changed, int left, int top, int right, int botto
         collapsingTextHelper.setCollapsedBounds(
             tmpRect.left + (isRtl ? toolbar.getTitleMarginEnd() : toolbar.getTitleMarginStart()),
             tmpRect.top + maxOffset + toolbar.getTitleMarginTop(),
-            tmpRect.right + (isRtl ? toolbar.getTitleMarginStart() : toolbar.getTitleMarginEnd()),
+            tmpRect.right - (isRtl ? toolbar.getTitleMarginStart() : toolbar.getTitleMarginEnd()),
```
ghus-raba, line 485: This fixes text going out of screen in collapsed state when end margin is set.

lib/java/com/google/android/material/internal/CollapsingTextHelper.java
```diff
@@ -593,19 +590,15 @@ private void calculateBaseOffsets() {
             isRtl ? ViewCompat.LAYOUT_DIRECTION_RTL : ViewCompat.LAYOUT_DIRECTION_LTR);
     switch (expandedAbsGravity & Gravity.VERTICAL_GRAVITY_MASK) {
       case Gravity.BOTTOM:
-        float offset = shouldDrawMultiline() ? textLayoutHeight - textPaint.descent() : 0;
-        expandedDrawY = expandedBounds.bottom - offset;
+        expandedDrawY = expandedBounds.bottom - expandedTextHeight;
```
ghus-raba, line 593: These changes are for `expandedDrawY` and `collapsedDrawY ` to always mark top edge of the text to draw. Before, it was top edge for multiline, but bottom (baseline actually) for single line.

lib/java/com/google/android/material/internal/CollapsingTextHelper.java
```diff
@@ -678,15 +669,15 @@ public void draw(@NonNull Canvas canvas) {
 
       if (drawTexture) {
         // If we should use a texture, draw it instead of text
-        canvas.drawBitmap(expandedTitleTexture, x, y + ascent, texturePaint);
```
ghus-raba, line 681: Since `y` marks top of the text to draw, no need to add (actually subtract, see why it is a bit confusing?) the `ascent`

lib/java/com/google/android/material/internal/CollapsingTextHelper.java
```diff
@@ -699,30 +690,27 @@ private boolean shouldDrawMultiline() {
...
 
-    // position the overlays
-    canvas.translate(x - currentExpandedX, 0);
-
     // Collapsed text
     textPaint.setAlpha((int) (collapsedTextBlend * originalAlpha));
-    canvas.drawText(
-        textToDrawCollapsed, 0, textToDrawCollapsed.length(), 0, -ascent, textPaint);
+    canvas.drawText(textToDrawCollapsed, 0, textToDrawCollapsed.length(), 0,
+        textLayout.getLineBaseline(0), textPaint);
```
ghus-raba, line 704: Since `canvas.drawText(...)` expects the y to be the text baseline and canvas here is translated to top of the text, we get the baseline of the first line. 

Now that I look at it, it is probably true, that `-ascent == textLayout.getLineBaseline(0)`, but the latter is IMO clearer.

lib/java/com/google/android/material/internal/CollapsingTextHelper.java
```diff
@@ -699,30 +690,27 @@ private boolean shouldDrawMultiline() {
...
+      @NonNull Canvas canvas, float currentExpandedX, float y) {
     int originalAlpha = textPaint.getAlpha();
     // positon expanded text appropriately
     canvas.translate(currentExpandedX, y);
     // Expanded text
     textPaint.setAlpha((int) (expandedTextBlend * originalAlpha));
     textLayout.draw(canvas);
 
-    // position the overlays
-    canvas.translate(x - currentExpandedX, 0);
```
ghus-raba, line 711: Here I admit, that I am not sure what was this line supposed to do, so this might not be the smartest thing to do. 

However, we want to draw the single line in the same place as the multiline above via `textLayout.draw(canvas)`, so moving away from that point does not seem right.

ghus-raba:
> the margin bottom was previously from the baseline of the text but now it seems it’s from the text descent

You are right, it took me a while to realise. I must have forgotten to offset it after moving the anchor to top. 

On the other hand, it feels more natural this way, as it is same behavior as when we set bottom margin to a TextView. Plus IMHO it is a bit odd, that if we set `app:expandedTitleMarginBottom="0dp"`, part of the text under the baseline will get cut off.

Should I return the offset by descent then?

ymarian:
I agree it's nicer that it matches what TextView does. Thanks for the comments. Let me discuss with the rest of the team. We have a lot of screenshots tests internally that need to be updated, so it will take us more time to merge this in because of that

ymarian:
@ghus-raba ok we talked about this. The spec does use baseline for the margin here:
So we'll need to keep that, sorry for the back and forth.

Screen Shot 2020-07-08 at 2 02 14 PM

ghus-raba:
Ok, I offset it again, so the bottom margin is counted from baseline. 

Let me know if there is anything else to do.

PR #1483, opened 2020-07-15
[TabLayout] Selected indicator paint can be null

### LINKED ISSUE #1481: [TabLayout] Indicator color always influences indicator drawable
**Description:** Indicator color always sets tint to indicator drawable even when the color was cleared using style or in layout.
Problem is that [indicator paint](https://github.com/material-components/material-components-android/blob/f945a0649b9cddc13afafeb80df7218f1aa85ef8/lib/java/com/google/android/material/tabs/TabLayout.java#L2825) is never null and thus is always set to indicator drawable in [onDraw](https://github.com/material-components/material-components-android/blob/f945a0649b9cddc13afafeb80df7218f1aa85ef8/lib/java/com/google/android/material/tabs/TabLayout.java#L3180).

So right now there is no option to keep the color of a drawable. 

- Setting `app:tabIndicatorColor="@null"` - hides indicator drawable (alpha 0).
- Setting `app:tabIndicatorColor` to any color - sets color to the drawable (intended behavior).
- Not setting `app:tabIndicatorColor` - uses primaryColor for drawable (intended behavior).

**Expected behavior:** Indicator color does not affect drawable when it is not set.

**Source code:** 

Layout:
```xml
<com.google.android.material.tabs.TabLayout
                android:layout_height="56dp"
                app:tabGravity="center"
                app:tabIndicator="@drawable/tabbar_indicator_gradient_blue"
                app:tabIndicatorColor="@null"
                app:tabIndicatorFullWidth="false"
                app:tabIndicatorHeight="3dp"
                app:tabMode="fixed"/>
```
Drawable (tabbar_indicator_gradient_blue):
```xml
<shape xmlns:android="http://schemas.android.com/apk/res/android"
    android:shape="rectangle">
    <gradient
        android:angle="0"
        android:endColor="#00b0eb"
        android:startColor="#0086b3"/>
    <corners android:topLeftRadius="3dp" android:topRightRadius="3dp"/>
</shape>
```

In this case the indicator dissapears because the indicator color is applied even when it should not.

**Android API version:** 24+

**Material Library version:** 1.2.0-beta01

**Device:** Huawei P20 lite (API 28)

Del-S:
Selected indicator paint can be null to allow using drawable color in TabLayout.

closes #1481 

When you set `app:tabIndicatorColor="@null"` it will clear the indicator color and it will not influence drawable anymore.

### DIFF
--- a/lib/java/com/google/android/material/tabs/TabLayout.java
+++ b/lib/java/com/google/android/material/tabs/TabLayout.java
@@ -2822,7 +2822,7 @@ public class TabLayout extends HorizontalScrollView {
 
   class SlidingTabIndicator extends LinearLayout {
     private int selectedIndicatorHeight;
-    @NonNull private final Paint selectedIndicatorPaint;
+    private Paint selectedIndicatorPaint;
     @NonNull private final GradientDrawable defaultSelectionIndicator;
 
     int selectedPosition = -1;
@@ -2840,11 +2840,19 @@ public class TabLayout extends HorizontalScrollView {
     SlidingTabIndicator(Context context) {
       super(context);
       setWillNotDraw(false);
-      selectedIndicatorPaint = new Paint();
       defaultSelectionIndicator = new GradientDrawable();
     }
 
     void setSelectedIndicatorColor(int color) {
+      if (color == 0) {
+        selectedIndicatorPaint = null;
+        return;
+      }
+
+      if (selectedIndicatorPaint == null) {
+        selectedIndicatorPaint = new Paint();
+      }
+
       if (selectedIndicatorPaint.getColor() != color) {
         selectedIndicatorPaint.setColor(color);
         ViewCompat.postInvalidateOnAnimation(this);


Del-S:
@googlebot I signed it!

Del-S:
@googlebot I fixed it.

lib/java/com/google/android/material/tabs/TabLayout.java
```diff
@@ -2840,11 +2840,19 @@ private float approximateLineWidth(@NonNull Layout layout, int line, float textS
     SlidingTabIndicator(Context context) {
       super(context);
       setWillNotDraw(false);
-      selectedIndicatorPaint = new Paint();
       defaultSelectionIndicator = new GradientDrawable();
     }
 
     void setSelectedIndicatorColor(int color) {
+      if (color == 0) {
```
dsn5ft, line 2847: a problem here is that the constant value for `Color.WHITE` is 0. so i think this change would disallow white for the selected indicator color?

dsn5ft: oh never mind, `Color.TRANSPARENT` is 0. i still wonder if that's an issue.

dsn5ft: thinking about this a little more 😅, i think this is actually ok. because if the color is `Color.TRANSPARENT`, then we set the paint to null, and we don't draw the indicator, which is sort of the same as drawing it transparently. will import this PR and change the `0` to `Color.TRANSPARENT` to make this a little more obvious.

Del-S: Awesome. Thank you. :+1:

aszkup:
@dsn5ft Hi can you say when this feature will be released?

dsn5ft:
Hi @aszkup, it should be included in the next 1.3.0 alpha release, sometime within the next few weeks.

Del-S:
Can confirm that it works properly in **1.3.0-rc01**.

carstenhag:
@dsn5ft Can behavior changed like these please also be included in release notes? We have noticed now after some months of using 1.3.0 that the tab indicator is now showing up, when we usually always had it hidden :P.

I just re-checked the release notes and it was not mentioned.

Issue #1494, opened 2020-07-20
[BottomNavigationView] Add LabelVisibilityMode LABEL_VISIBILITY_UNSELECTED


charleston10:
The component contains some visualization properties, but does not include the alternative mode for visibility `insibility`

Contains
LABEL_VISIBILITY_AUTO, LABEL_VISIBILITY_SELECTED, LABEL_VISIBILITY_LABELED, LABEL_VISIBILITY_UNLABELED

But not contain LABEL_VISIBILITY_UNSELECTED

In this case when clicking on menu the Icon is showing but not show label

dsn5ft:
I don't think this behavior is in the Material spec for Bottom Navigation. Can you explain why you would want to show the labels only on unselected Bottom Navigation items? That seems counterintuitive.

PR #1495, opened 2020-07-20
[BottomNavigationView] Add new mode visibility "unselected"

charleston10:
https://github.com/material-components/material-components-android/issues/1494

A new behavior has been added for the label display modes
how to show only the icon, this mode and the inverse of `LABEL_VISIBILITY_SELECTED`

To take advantage of the behavior of `LABEL_VISIBILITY_SELECTED` I separated the method into a function `setCheckedVisibilityMode`, in this way it maintains the default behavior and to reverse the behavior a negation was used


charleston10:
@googlebot I fixed it.

charleston10:
no cla

PR #1497, opened 2020-07-21
[BottomNavigationView] add label mode unselected

charleston10:
#1494

A new behavior has been added for the label display modes
how to show only the icon, this mode and the inverse of LABEL_VISIBILITY_SELECTED

To take advantage of the behavior of LABEL_VISIBILITY_SELECTED I separated the method into a function setCheckedVisibilityMode, in this way it maintains the default behavior and to reverse the behavior a negation was used


dsn5ft:
Closing for now, let's continue the discussion on #1494 if needed.

Issue #1512, opened 2020-07-24
[MaterialTimePicker] Allow to be informed of MaterialTimePicker dialog events


mpost:
Similar to the `MaterialDatePicker` the `MaterialTimePicker` should have a dismiss, cancel, negative and positive listener. This brings the two dialogs closer to each other which should be a desirable trait. The positive listener would render the current `MaterialTimePicker.setListener()` obsolete.

Some more observations around the api and to bring the time and date dialogs closer:

* The `MaterialTimePicker` should also use a builder approach to create its instance the current newInstance() method feels like a not fully fleshed out api.
* The `MaterialTimePicker` uses the `TimeFormat` interface to store the two flags while the input mode is on the dialog class itself. To be in line with the date dialog all flags should be directly on the time dialog.
* The `MaterialTimePicker` should allow to set a title similarly to the date dialog.

PR #1517, opened 2020-07-25
[MaterialTimePicker] Added listeners

gabrielemariotti:
Added dismiss, cancel, negative and positive listeners as `MaterialDatePicker` #1512.
Updated the demo with the new positive listener.


ymarian:
Thanks @gabrielemariotti! I did a few cosmetics changes, and changed things to NonNull. I also removed the previous method rather than deprecating. Since this is only released in Alpha there is no need to preserve binary compatibility.

gabrielemariotti:
@ymarian Thanks for feedback.

PR #1518, opened 2020-07-25
Fix datepicker RTL month buttons icon

### LINKED ISSUE #1485: RTL issue with MaterialDatePicker
There is an issue with RTL layout for MaterialDatePicker, the arrows for the months is reversed:
 
Capture

matpag:
I switched the button icons when the layout is RTL.

Can this be enough to fix the problem?
Waiting for a feedback.

Fixes #1485

### DIFF
--- a/lib/java/com/google/android/material/datepicker/res/layout/mtrl_calendar_month_navigation.xml
+++ b/lib/java/com/google/android/material/datepicker/res/layout/mtrl_calendar_month_navigation.xml
@@ -51,7 +51,7 @@
     android:gravity="center"
     android:insetBottom="0dp"
     android:insetTop="0dp"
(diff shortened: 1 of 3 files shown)


dsn5ft:
Hi @matpag, thanks for the PR! I'm thinking of solving this slightly differently, with resource qualifiers and aliases, so that we can set the icons once in XML and not have to rely on doing it programmatically after the fact. Will sync that fix out soon and link the commit to this PR.

matpag:
@dsn5ft Your approach is clear, I had half of this implementation ready.
I updated my PR, pls take another look :)

dsn5ft:
Nice! That's almost exactly what I had. Will add you as a co-author 😁.

Issue #1380, opened 2020-06-04
[MaterialDatePicker] Disable InputModes


nielsz:
Currently, using `setInputMode` we can set the default Input mode, e.q. INPUT_MODE_CALENDAR or INPUT_MODE_TEXT. 

What I would like is to disable INPUT_MODE_TEXT at all, so just show the INPUT_MODE_CALENDAR mode, without an option to toggle between the two modes.

The API that I can think of is something like:
```
MaterialDatePicker.Builder.datePicker()
            .setDefaultInputMode(INPUT_MODE_TEXT)
            .setInputModes(listOf(INPUT_MODE_CALENDAR, INPUT_MODE_TEXT))
            .build()
```
So, renaming  the setInputMode to setDefaultInputMode, and allow a list of inputModes to be set as possibilities. If there's only one option, don't show the toggle.

dsn5ft:
One of the main purposes of the input mode toggle and text input mode is to give an alternative mode of entering the date for accessibility users who may not be able to use the calendar view. For that reason we will not be making it easier to disable the mode.

SikanderBakht:
Input mode is working for English language only. For other languages its not working properly.

PR #1520, opened 2020-07-25
[MaterialDatePicker] Option to hide the inputmode toggle in header

gabrielemariotti:
Added an option to hide the toggle in header to change the input mode.
Updated the demo with the new option.

To hide the toggle just use:
`builder.showToggleInputMode(false);`

The default value of the new option is true.
#1380

### DIFF
--- a/catalog/java/io/material/catalog/datepicker/DatePickerMainDemoFragment.java
+++ b/catalog/java/io/material/catalog/datepicker/DatePickerMainDemoFragment.java
@@ -99,6 +99,7 @@ public class DatePickerMainDemoFragment extends DemoFragment {
     final RadioGroup opening = root.findViewById(R.id.cat_picker_opening_month_group);
     final RadioGroup selection = root.findViewById(R.id.cat_picker_selection_group);
     final RadioGroup inputMode = root.findViewById(R.id.cat_picker_input_mode_group);
+    final RadioGroup toggleInputMode = root.findViewById(R.id.cat_picker_toggle_input_mode_group);
 
     launcher.setOnClickListener(
         v -> {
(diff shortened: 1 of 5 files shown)


dsn5ft:
One of the purposes of the input mode toggle is to give an alternative mode of entering the date for accessibility users who may not be able to use the calendar view. For that reason I don't think we should make it any easier to disable the mode.

gabrielemariotti:
@dsn5ft feel free to close the PR.
I've only pushed the PR on the issue labeled as feature request and I am not the author of the issue.

Just a different point of view.
This option can work also with the input mode = text and not only with the calendar view.

In any case I can't understand why you should remove an option/feature.
To be clear I totally agree with you about:

> input mode toggle is to give an alternative mode of entering the date for accessibility users who may not be able to use the calendar view
 
and it means that the responsibility of the accessibility is in charge of the library that **must** provide an alternative method, but it doesn't mean that the library should force the devs/companies to use this alternative.
Just give the devs/companies/designer the option to remove the toggle if there are cases where it is not suggested to show it.

In any case currently it is quite simple to have the same result with a style:
https://stackoverflow.com/questions/62090970/how-to-change-the-color-of-the-materialdatepicker-in-keyboard-mode-also-it-i/62091571#62091571

dsn5ft:
Gotcha, I'll close the PR for now since we don't have an API for the developer to add an alternative mode. But honestly even if we did have that API, I'd be worried that developers would just hide the toggle without providing an alternative, because they think the button doesn't look good or something like that. So in the future I think if anything we'd be more likely to allow for swapping the text input mode for a different custom a11y mode, as opposed to hiding the toggle altogether.

Agreed that we shouldn't always be forcing things as a library, but this is an important enough piece of functionality that we want to be available and consistent.

gabrielemariotti:
@dsn5ft 
Thanks for feedback.

I think the linked issue #1380 should be close for the same reason.

Issue #1376, opened 2020-06-04
[Theme] Lint should warn if setting colorAccent or colorPrimaryDark


nickbutcher:
`colorAccent` and `colorPrimaryDark` from `AppCompat` are superseded by newer theming attributes (`colorSecondary` and `colorPrimaryVariant` respectively) so setting them directly in your theme is likely a mistake. This seems to be causing confusion to developers:

https://twitter.com/ArchieQuiones1/status/1268390022538911746

What about adding a Lint check to the library to warn if developers specify these attributes.

PR #1523, opened 2020-07-26
Lint checks for colorAccent and colorPrimaryDark

gabrielemariotti:
Added lint checks:
- ThemeCheckColorAccent : Detects usages colorAccent in style resources
- ThemeCheckColorPrimaryDark : Detects usages colorPrimaryDark in style resources

#1376

### DIFF
--- a/catalog/build.gradle
+++ b/catalog/build.gradle
@@ -18,6 +18,8 @@ dependencies {
 
   api compatibility("recyclerViewSelection")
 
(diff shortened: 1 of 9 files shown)


gabrielemariotti:
Just an example of the result:

Schermata 2020-07-26 alle 08 02 45

There is also a raw README.md (obviously the lib is not published).

PR #1528, opened 2020-07-27
[MaterialDatePicker] placeholder text in input picker

gabrielemariotti:
According to [material design doc](https://material.io/components/pickers#anatomy) the input picker should display a placeholder text with the date format.

Added the placeholder text in input pickers with single and range selectors.

Schermata 2020-07-28 alle 00 31 38

Schermata 2020-07-28 alle 00 30 44


dsn5ft:
Thanks @gabrielemariotti!

PR #1529, opened 2020-07-27
[MaterialDatePicker] clear_text as endIconMode in input picker

gabrielemariotti:
Added a `clear_text` icon as `endIconMode` in the input picker for single and range selectors.

I am not sure it is compliant with your material design guideline of MaterialDatePicker.
It is never displayed in the screens reported in the doc but it is not specify as "don't".

When a date (or range) is selected it could be useful to display the clear_text action to clear the selected date in a quick way.

Schermata 2020-07-28 alle 00 58 25


dsn5ft:
Thanks @gabrielemariotti, I think the designers intentionally left out the clear button for date pickers because of limited horizontal space, and since it should be easy enough to backspace to clear the few date input characters.

https://material.io/components/pickers#mobile-pickers

image

gabrielemariotti:
@dsn5ft 
Thanks for feedback

PR #1534, opened 2020-07-28
[Slider] Added XML enabled property to constructor attrs evaluation

### LINKED ISSUE #1531: [Slider] Attribute "android:enabled" not working
**Description:**
When using the default slider, if provided the `android:enabled` attribute inside a layout, the slider is not disabled.

**Expected behavior:** 
I expect that using  `android:enabled`  has the same effect than `slider.isEnabled = false` (kotlin) which works properly.

**Source code:** 

```xml
<com.google.android.material.slider.Slider
            android:layout_width="200dp"
            android:enabled="false"/>
```

**Android API version:** 
* Execution env : API 22
* Target env : API 29

**Material Library version:** 1.3.0-alpha02

**Device:** No name tablet running API 22

matpag:
I followed what you did for `android:enabled` property in `TextInputLayout` [here](https://github.com/material-components/material-components-android/blob/b82d41fe55e9538509299a1ee90798b29839392c/lib/java/com/google/android/material/textfield/res/values/attrs.xml#L24) and replicated it for Slider.
Because the evaluation happens in `BaseSlider` class, this change will work for both `Slider` and `RangeSlider`

Closes #1531

### DIFF
--- a/lib/java/com/google/android/material/slider/BaseSlider.java
+++ b/lib/java/com/google/android/material/slider/BaseSlider.java
@@ -102,7 +102,7 @@ import java.util.List;
  * displays the selected value using letters to indicate magnitude (e.g.: 1.5K, 3M, 12B, etc..).
  *
  * <p>With the default style {@link
- * com.google.android.material.R.style#Widget_MaterialComponents_Slider}, colorPrimary and
+ * com.google.android.material.R.style#Widget.MaterialComponents.Slider}, colorPrimary and
  * colorOnPrimary are used to customize the color of the slider when enabled, and colorOnSurface is
  * used when disabled. The following attributes are used to customize the slider's appearance
  * further:
@@ -156,6 +156,7 @@ import java.util.List;
  * @attr ref com.google.android.material.R.styleable#Slider_android_stepSize
  * @attr ref com.google.android.material.R.styleable#Slider_android_valueFrom
  * @attr ref com.google.android.material.R.styleable#Slider_android_valueTo
+ * @attr ref com.google.android.material.R.styleable#Slider_android_enabled
  * @attr ref com.google.android.material.R.styleable#Slider_haloColor
  * @attr ref com.google.android.material.R.styleable#Slider_haloRadius
  * @attr ref com.google.android.material.R.styleable#Slider_labelBehavior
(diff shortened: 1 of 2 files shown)


lib/java/com/google/android/material/slider/BaseSlider.java
```diff
@@ -102,7 +102,7 @@
  * displays the selected value using letters to indicate magnitude (e.g.: 1.5K, 3M, 12B, etc..).
  *
  * <p>With the default style {@link
- * com.google.android.material.R.style#Widget_MaterialComponents_Slider}, colorPrimary and
+ * com.google.android.material.R.style#Widget.MaterialComponents.Slider}, colorPrimary and
```
dsn5ft, line 105: this should be `R.style#Widget_MaterialComponents_Slider` since that's how it would be referenced programmatically

lib/java/com/google/android/material/slider/BaseSlider.java
```diff
@@ -349,6 +350,8 @@ public TooltipDrawable createTooltipDrawable() {
 
     accessibilityManager =
         (AccessibilityManager) getContext().getSystemService(Context.ACCESSIBILITY_SERVICE);
+
+
```
dsn5ft, line 354: extra lines

lib/java/com/google/android/material/slider/BaseSlider.java
```diff
@@ -1206,6 +1212,11 @@ public void setTrackInactiveTintList(@NonNull ColorStateList trackColor) {
     invalidate();
   }
 
+  /**
+   * Enable or disable the Slider
+   * @see #isEnabled()
+   * @attr ref com.google.android.material.R.styleable#Slider_android_enabled
+   */
```
dsn5ft, line 1219: this comment shouldn't be needed, as by default the doc from the parent method is inherited

lib/java/com/google/android/material/slider/BaseSlider.java
matpag: I thought this had to be a valid Javadoc link, sorry 😅

lib/java/com/google/android/material/slider/BaseSlider.java
matpag: You are right 👍

PR #1556, opened 2020-07-31
[MaterialAlertDialog] Changed default elevation to 24 dp

### LINKED ISSUE #1255: [MaterialAlertDialog] Wrong elevation
**Description:** 
I'm having an issue getting my custom dialog to appear as the same surface color as the MaterialAlertDialog in dark mode and it appears that the MaterialAlertDialog's elevation is set to 16dp instead of 24dp.

I am referencing the official guidelines  [here](https://material.io/components/dialogs) for material design dialog elevation which states that "Dialogs are displayed at 24dp elevation and can display a shadow" but following up the theme inheritance I found that the elevation of MaterialAlertDialog is set at 16dp instead.

The chain of inheritance is as follows:

The default theme overlay defined in `MaterialAlertDialogBuilder` class's `MATERIAL_ALERT_DIALOG_THEME_OVERLAY` constant references and the respective parent
->`@style/ThemeOverlay.MaterialComponents.MaterialAlertDialog`  
->`Base.ThemeOverlay.MaterialComponents.MaterialAlertDialog` 
->`Base.V14.ThemeOverlay.MaterialComponents.MaterialAlertDialog` 
->`ThemeOverlay.AppCompat.Dialog.Alert`
-> `Base.ThemeOverlay.AppCompat.Dialog.Alert` 
->`Base.ThemeOverlay.AppCompat.Dialog` 
->`Base.V21.ThemeOverlay.AppCompat.Dialog` (in API 21 onwards) which defines the item
 `<item name="android:windowElevation">@dimen/abc_floating_window_z</item>`

The `@dimen/abc_floating_window_z` dimension has the value of 16dp which does not align with the material spec, causing custom widget that follows the anatomy of the material spec to have inconsistent shadow in light theme and different alpha values in dark theme. 

`android:windowElevation` attribute should be set in one of the MaterialComponents base style to override the value from appCompat library

**Expected behavior:** 
Alert dialog elevation is set at 24dp

**Android API version:** 
Android 10 (API level 29)

**Material Library version:** 
1.1.0-rc2

**Device:** 
Oneplus 7T Pro

gabrielemariotti:
According to the [material design doc](https://material.io/design/environment/elevation.html#default-elevations)  the default elevation for the `MaterialAlertDialog` should be `24dp`.

Schermata 2020-07-31 alle 23 17 15

Currently the `MaterialAlertDialog` seems to use a value of `16dp` inheriting the value from `<item name="android:windowElevation">@dimen/abc_floating_window_z</item>` defined in the AppCompat theme.  
Using the dark mode with a `CardView `with `16dp` and a `CardView `with `24dp` you can check:

current

With this PR the value is overridden in the default `ThemeOverlay.MaterialComponents.MaterialAlertDialog`.

It closes #1255.

### DIFF
--- a/lib/java/com/google/android/material/dialog/res/values-v21/themes_base.xml
+++ b/lib/java/com/google/android/material/dialog/res/values-v21/themes_base.xml
@@ -44,4 +44,9 @@
 
   <style name="Base.Theme.MaterialComponents.Light.Dialog" parent="Base.V21.Theme.MaterialComponents.Light.Dialog"/>
 
+  <style name="Base.ThemeOverlay.MaterialComponents.MaterialAlertDialog" parent="Base.V21.ThemeOverlay.MaterialComponents.MaterialAlertDialog"/>
+
+  <style name="Base.V21.ThemeOverlay.MaterialComponents.MaterialAlertDialog" parent="Base.V14.ThemeOverlay.MaterialComponents.MaterialAlertDialog">
+    <item name="android:windowElevation">@dimen/mtrl_alert_dialog_elevation</item>
+  </style>
 </resources>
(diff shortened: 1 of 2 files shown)


dsn5ft:
Thanks @gabrielemariotti, we were sort of holding off on this for now since the very large elevation of 24dp could cause some accessibility contrast issues for dialogs with the dark theme elevation overlay.

gabrielemariotti:
@dsn5ft Thanks for feedback.

drchen:
I'll close the PR since we are working on a new version of Material design and this becomes obsolete.

PR #1566, opened 2020-08-05
[TextInputLayout] Exposed the style of the ExposedDropDownMenu

gabrielemariotti:
Currently the ExposedDropDownMenu provided by `DropdownMenuEndIconDelegate` doesn't use the `shapeAppearance` defined in the app theme and the corner radius are defined by code.

With this change:
- the ExposedDropDownMenu uses the `shapeAppearance` defined in the app theme
- there is a new attribute `exposedDropDownMenuStyle` defined in the `TextInputLayout` style with allows to customize the `shapeAppearance` /`shapeAppearanceOverlay` (and some other properties) of the ExposedDropDownMenu.

This change can be tested with the current catalog (changing the  `shapeAppearance`).

Schermata 2020-08-05 alle 21 32 38
Schermata 2020-08-05 alle 21 32 48

Schermata 2020-08-05 alle 21 33 14

### DIFF
--- a/lib/java/com/google/android/material/textfield/DropdownMenuEndIconDelegate.java
+++ b/lib/java/com/google/android/material/textfield/DropdownMenuEndIconDelegate.java
@@ -28,6 +28,7 @@ import android.animation.ValueAnimator.AnimatorUpdateListener;
 import android.annotation.SuppressLint;
 import android.content.Context;
 import android.content.res.ColorStateList;
+import android.content.res.TypedArray;
 import android.graphics.Color;
 import android.graphics.drawable.Drawable;
 import android.graphics.drawable.LayerDrawable;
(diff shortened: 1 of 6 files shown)


leticiarossi:
Thanks for the PR! I think it would be confusing to have a `exposedDropDownMenuStyle` since we have the TextInputLayout ExposedDropdownMenu styles and the AutoCompleteTextView styles. 

Since we only want to affect the dropdown's popup background, I think it would make more sense to have an attribute like `popupBackgroundShapeAppearance`. I wonder if we could remove the logic that sets the background from the `DropdownMenuEndIconDelegate` and move it to the `MaterialAutoCompleteTextView` class instead, where it could read that attribute (that would then be set inside of the existent `Widget.MaterialComponents.AutoCompleteTextView.*` styles.

Would you be interested in trying that?

gabrielemariotti:
Thanks for your feedback.
I will update the PR.

Are you interested to set in the `Widget.MaterialComponents.AutoCompleteTextView.*` styles also attributes like
`popupVerticalPadding`and `popupElevation`?

leticiarossi:
We already set `android:dropDownVerticalOffset` and `android:popupElevation` so it'd make sense to also read then in the class! I think it would also make sense to have popup*Padding attrs (right now we only set vertical padding but developers may want to set different paddings too)

gabrielemariotti:
Moving all the logic that sets the background to the `MaterialAutoCompleteTextView` class, how do you prefer to set the popupBackground with an `AppCompatAutoCompleteTextView` (for example with a `Bridge` theme)?

gabrielemariotti:
I've just updated the PR moving the all the logic that sets the background to the `MaterialAutoCompleteTextView` class.

jsdeyro:
Will these be able to customize the ExposedDropDownMenu dropdown's popup background? If so, when this will be updated?

PR #1575, opened 2020-08-06
[CollapsingToolbarLayout] When scrimAnimator already created we must update the duration when calling `setScrimAnimationDuration`

extmkv:
**Component:** CollapsingToolbarLayout

**Problem:** When calling `setScrimAnimationDuration` if the `scrimAnimator` is already created he duration of the animator must be updated.
On `animateScrim` the null verification protects the duration to be re-defined.


rddsantos:
Looks good!

extmkv:
@googlebot I signed it!

drchen (approved):

PR #1580, opened 2020-08-06
[TextInputLayout] background color of the indicatorArea when error is displayed

gabrielemariotti:
Sometimes in the `TextInputLayout` the error text is not full visible.
Instead of changing the `errorTextColor` could be useful an option to change the background color leaving the std errorTextColor.
Example:
Schermata 2020-08-06 alle 18 57 41

This change adds an option to tint the background color of the indicatorArea when an error is displayed.  
The default background doesn't change the current behavior.

Schermata 2020-08-06 alle 18 57 57

Example:

```
  <com.google.android.material.textfield.TextInputLayout
      app:indicatorAreaBackgroundColorOnError="@color/indicator_area_backgroundcolor_selector"
```

or:

`labelTextField.setIndicatorAreaBackgroundColorOnError(...);`

It is related to #1234

### DIFF
--- a/lib/java/com/google/android/material/textfield/res-public/values/public.xml
+++ b/lib/java/com/google/android/material/textfield/res-public/values/public.xml
@@ -57,6 +57,7 @@
   <public name="errorIconDrawable" type="attr"/>
   <public name="errorIconTint" type="attr"/>
   <public name="errorIconTintMode" type="attr"/>
+  <public name="indicatorAreaBackgroundColoraOnError" type="attr"/>
   <public name="helperText" type="attr"/>
   <public name="helperTextEnabled" type="attr"/>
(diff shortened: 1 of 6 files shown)


lib/java/com/google/android/material/textfield/res-public/values/public.xml
```diff
@@ -57,6 +57,7 @@
   <public name="errorIconDrawable" type="attr"/>
   <public name="errorIconTint" type="attr"/>
   <public name="errorIconTintMode" type="attr"/>
+  <public name="indicatorAreaBackgroundColoraOnError" type="attr"/>
```
matpag, line 60: There is an extra "a" in the name
```suggestion
  <public name="indicatorAreaBackgroundColorOnError" type="attr"/>
```

Issue #1272, opened 2020-05-04
[Exposed DropDown Menu] Does a simple way to know when user click on arrow icon exists?


user-fal:
Hi team,

I struggled couple days ago. I have a use case where I should populate exposed dropdown menu when user click on arrow to show the dropdown.

Does callback exist?

Thanks

PR #1581, opened 2020-08-06
[TextInputLayout] set the endIconOnClickListener also with endIconMode

gabrielemariotti:
Currently the `endIconMode` overrides the `endIconOnClickListener`.
With this change it is possible to add an `endIconOnClickListener` also when the `endIconMode` is set.  
The listener is called after the `onClickListener` defined in the `EndIconDelegate`.

### DIFF
--- a/lib/java/com/google/android/material/textfield/TextInputLayout.java
+++ b/lib/java/com/google/android/material/textfield/TextInputLayout.java
@@ -380,6 +380,7 @@ public class TextInputLayout extends LinearLayout {
   @Nullable private Drawable endDummyDrawable;
   private int endDummyDrawableWidth;
   private Drawable originalEditTextEndDrawable;
+  protected OnClickListener endIconOnClickListener;
   private OnLongClickListener endIconOnLongClickListener;
   private OnLongClickListener errorIconOnLongClickListener;
   @NonNull private final CheckableImageButton errorIconView;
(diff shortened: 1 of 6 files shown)


user-fal:
Hi @gabrielemariotti,
Does #1272 was finally resolved?

lib/java/com/google/android/material/textfield/TextInputLayout.java
```diff
@@ -380,6 +380,7 @@
   @Nullable private Drawable endDummyDrawable;
   private int endDummyDrawableWidth;
   private Drawable originalEditTextEndDrawable;
+  protected OnClickListener endIconOnClickListener;
```
drchen, line 383: Why do we need this and where is it being used?

lib/java/com/google/android/material/textfield/TextInputLayout.java
```diff
@@ -3056,6 +3057,13 @@ public int getEndIconMode() {
    *     will have
    */
   public void setEndIconOnClickListener(@Nullable OnClickListener endIconOnClickListener) {
+    this.endIconOnClickListener = endIconOnClickListener;
+    if (getEndIconMode() == END_ICON_NONE || getEndIconMode() == END_ICON_CUSTOM) {
```
drchen, line 3061: What about other modes?

drchen (changes requested):

lib/java/com/google/android/material/textfield/TextInputLayout.java
gabrielemariotti: It is used in the `EndIconDelegate#callOriginalEndIconOnClickListener` method.

Issue #1076, opened 2020-03-01
[Snackbar] Setting LayoutParams for SnackbarLayout doesn't work


Zoro38:
**Description:**

There is no effect when trying to set LayoutParams for the SnackbarLayout (i.e. adding some margins).
I tried to add margins to the Snackbar because I wanted it not to stick to the bottom edges of the screen. Instead, I wanted it to float just like shown in the Material design guidelines.

This bug does not occur on Library version 1.0.0. I've reproduced this issue with version 1.1.0 and also with the latest alpha (1.2.0-alpha05).

**Expected behavior:**

![expected](https://user-images.githubusercontent.com/18499748/75632463-9224e880-5bfc-11ea-9122-1f3a61f06167.png)

**Actual behavior:**

![actual](https://user-images.githubusercontent.com/18499748/75632466-97823300-5bfc-11ea-8f1d-84ae7469911a.png)

**Source code:** 

```
Snackbar snackbar = Snackbar.make(getView(), "Test", Snackbar.LENGTH_LONG);
Snackbar.SnackbarLayout snackbarLayout = (Snackbar.SnackbarLayout) snackbar.getView();
CoordinatorLayout.LayoutParams layoutParams = (CoordinatorLayout.LayoutParams) 
snackbarLayout.getLayoutParams();
layoutParams.setMargins(32, 0, 32, 32);
snackbarLayout.setLayoutParams(layoutParams);
snackbar.show();
```

**Android API version:** 29

**Material Library version:** 1.1.0 and 1.2.0-alpha05

**Device:** Samsung Galaxy S10 and Android Emulator (Pixel 2 API 29)

jairrab:
Can confirm this issue. The last version where the margin behavior was working properly was with version 1.1.0-alpha10.  After that, the margin is not working.

KirillBorodin:
Guys, I am just faced this issue right before release the app to Google Play. Fortunately before, and not after...

ankushyerawar:
Facing same issue could anyone help!!

danielgomezrico:
I believe you are using the wrong style to setup the Snackbar, you need to use this in the app theme:
> <item name="snackbarStyle">@style/Widget.MaterialComponents.Snackbar</item>

Instead of `@style/Widget.MaterialComponents.Snackbar.FullWidth`.

dpreussler:
it seems there is a general issue here, setting margins dont seem to work via xml.

None of these work:

```
[...]

PR #1583, opened 2020-08-07
[Snackbar] bugfix: setting layoutParams for snackbarLayout `Open #1076`

cchencool:
Solve issue:[Setting LayoutParams for SnackbarLayout doesn't work bug]
https://github.com/material-components/material-components-android/issues/1076

Reason: BaseTransientBottomBar#updateMargins ignored the original layoutParams in the view and simply cover it with a new value.


cchencool:
@googlebot I signed it!

PR #1588, opened 2020-08-07
[Slider] Fix the validation of the slider step size and values

### LINKED ISSUE #1587: [RangeSlider/Slider] Invalid validation of step size
**Description:** The validation of slider values is incorrect and thus some valid values are not allowed.

**Expected behavior:** I'm trying to set up a slider with the range `0.1..10.0` and a step of `0.1` should be valid but the slider throws an exception:
```
    java.lang.IllegalStateException: The stepSize(0.1) must be 0, or a factor of the valueFrom(0.1)-valueTo(10.0) range
        at com.google.android.material.slider.BaseSlider.validateStepSize(BaseSlider.java:456)
        at com.google.android.material.slider.BaseSlider.validateConfigurationIfDirty(BaseSlider.java:490)
        at com.google.android.material.slider.BaseSlider.calculateTicksCoordinates(BaseSlider.java:1268)
        at com.google.android.material.slider.BaseSlider.onSizeChanged(BaseSlider.java:1261)
```

**Source code:**
```kotlin
val slider: Slider = ...
slider.valueFrom = 0.1f
slider.valueTo = 10f
slider.stepSize = 0.1f
```

**Android API version:** Android 10 (API 29)

**Material Library version:** 1.2.0

**Device:** Google Pixel 4XL

ansman:
This fixes #1587

### DIFF
--- a/lib/java/com/google/android/material/slider/BaseSlider.java
+++ b/lib/java/com/google/android/material/slider/BaseSlider.java
@@ -469,7 +469,8 @@ abstract class BaseSlider<
   }
 
   private void validateStepSize() {
-    if (stepSize > 0.0f && ((valueTo - valueFrom) / stepSize) % 1 > THRESHOLD) {
+    float remainder = ((valueTo - valueFrom) / stepSize) % 1;
+    if (stepSize > 0.0f && remainder > THRESHOLD && remainder < 1 - THRESHOLD) {
       throw new IllegalStateException(
           String.format(
               EXCEPTION_ILLEGAL_STEP_SIZE,
@@ -489,7 +490,8 @@ abstract class BaseSlider<
                 Float.toString(valueFrom),
                 Float.toString(valueTo)));
       }
-      if (stepSize > 0.0f && ((valueFrom - value) / stepSize) % 1 > THRESHOLD) {
+      float remainder = ((value - valueFrom) / stepSize) % 1;
+      if (stepSize > 0.0f && remainder > THRESHOLD && remainder < 1 - THRESHOLD) {
         throw new IllegalStateException(
             String.format(
                 EXCEPTION_ILLEGAL_DISCRETE_VALUE,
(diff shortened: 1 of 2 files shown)


ansman:
This is the commit that fixes the issue: https://github.com/material-components/material-components-android/pull/1588/commits/a8960218d364b6a2bce90ac8c44e782c71d1c05f

ansman:
@googlebot I fixed it.

dsn5ft:
Thanks @ansman! Because of how our commit syncing works, this should be based on master, and then once the commit is synced back here, then we can cherry pick it into the 1.2 branch.

ansman:
Got it, will fix

lib/java/com/google/android/material/slider/BaseSlider.java
```diff
@@ -489,7 +490,8 @@ private void validateValues() {
                 Float.toString(valueFrom),
                 Float.toString(valueTo)));
       }
-      if (stepSize > 0.0f && ((valueFrom - value) / stepSize) % 1 > THRESHOLD) {
```
ansman, line 492: Doing `valueFrom - value` will always produce 0 or a negative value since `value >= valueFrom` which means that this check didn't actually validate the value and would allow any value.

I believe the correct check is `value - valueFrom`

lib/javatests/com/google/android/material/slider/SliderConfigTest.java
```diff
@@ -119,6 +119,29 @@
...
+          null,
+        },
+        new Object[] {
+          /* valueFrom = */ 0.1f,
+          /* valueTo = */ 10f,
+          /* stepValue = */ 0.1f,
+          /* value = */ 0.1999999f, // This emulates a rounding error from 0.2
+          null,
+        },
+        new Object[][] {
```
ansman, line 137: This weird double array is needed because `ImmutableList` only have overloads for a limited number of values then it switches to `vararg` and since we're using arrays here it becomes confused and thinks that the list type is `Object` rather than `Object[]` which causes compile issues. Using a double array works just fine though.

ansman:
I'm not sure why the Travis build fails. Building locally works just fine. It seems like the failure is unrelated to my changes.

wcshi:
Quick update, there's a delay on this PR because the team is internally discussing alternative solutions because this change breaks another test case.

gabrielemariotti:
As alternative you can evaluate the use of `BigDecimal` to avoid all rounding issues due to float.

PR #1598, opened 2020-08-08
[TextInputLayout] Added method to set hint with a String resource

### LINKED ISSUE #959: [TextInputLayout] Expose a setHint function that accepts a String Resource
**Current Problem:**
At the moment, the only way to set the hint of a `TextInputLayout` is using a fully formed CharSequence defined:

```public void setHint(@Nullable CharSequence hint)``` [Source](https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/textfield/TextInputLayout.java#L1356)

**Proposed Solution:**
Ideally, there should also exist a public function that can set the hint using a String Resource.
Something like:

```public void setHint(@StringRes int hint)```

### PR DESCRIPTION
**Pull request by gabrielemariotti**
It closes #959

### DIFF
--- a/lib/java/com/google/android/material/textfield/TextInputLayout.java
+++ b/lib/java/com/google/android/material/textfield/TextInputLayout.java
@@ -1471,7 +1471,7 @@ public class TextInputLayout extends LinearLayout {
    * @param textHintId The resource id of the text to display in the floating label
    */
   public void setHint(@StringRes int textHintId) {
-    setHint(getContext().getText(textHintId));
+    setHint(textHintId != 0 ? getResources().getText(textHintId) : null);
   }
 
   private void setHintInternal(CharSequence hint) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: lib/java/com/google/android/material/textfield/TextInputLayout.java [hunk 1]
```diff
@@ -1463,6 +1463,17 @@ public void setHint(@Nullable CharSequence hint) {
...
 
+  /**
+   * Set the hint to be displayed in the floating label, if enabled, using the given resource id.
+   *
+   * @see #setHintEnabled(boolean)
+   * @see #setHint(CharSequence)
+   * @param textHintId The resource id of the text to display in the floating label
+   */
+  public void setHint(@StringRes int textHintId) {
+    setHint(getContext().getText(textHintId));
```
wcshi, line 1474: Looking how we do it in similar methods, please call setHint(textHintId != 0 ? getResources().getText(textHintId) : null)

gabrielemariotti: Done

PR #1603, opened 2020-08-10
[ShapeableImageView] Respect strokeWidth in ShapeMask

simekadam:
Problem:
When stroke is set it is rendered so that half of the stroke is inside of the bounds of the ImageView and half is outside. The problem is that anything that is outside of the bounds of the View gets clipped and thus won't be visible. That leads to incorrectly rendered stroke. Normally this should be fixed by telling the parent to not clip
its children, but that doesn't work here as here it gets actually clipped by the view itself by its ShapeMask.

Changes:
Offset the shape mask by half of the stroke width on all sides.

Screenshot:
Here is a screenshot from a little sample app I built: 
On the top is ShapeableImageView, on the bottom is a patched version. 
image

Sample app:
[ImageViewFix.zip](https://github.com/material-components/material-components-android/files/5050571/ImageViewFix.zip)

### DIFF
--- a/lib/java/com/google/android/material/imageview/ShapeableImageView.java
+++ b/lib/java/com/google/android/material/imageview/ShapeableImageView.java
@@ -153,8 +153,12 @@ public class ShapeableImageView extends AppCompatImageView implements Shapeable 
   }
 
   private void updateShapeMask(int width, int height) {
+    final float strokeOffset = getStrokeWidth() / 2f;
     destination.set(
-        getPaddingLeft(), getPaddingTop(), width - getPaddingRight(), height - getPaddingBottom());
+        getPaddingLeft() + strokeOffset,
+        getPaddingTop() + strokeOffset,
+        width - getPaddingRight() - strokeOffset,
+        height - getPaddingBottom() - strokeOffset);
     pathProvider.calculatePath(shapeAppearanceModel, 1f /*interpolation*/, destination, path);
     // Remove path from rect to draw with clear paint.
     maskPath.rewind();


simekadam:
@googlebot I signed it!

simekadam:
I was also able to "fix" this by removing not disable the hw buffer rendering [here](https://github.com/material-components/material-components-android/blob/d8a7a9afd58aea8e13a0d71d8ea693c7f7962395/lib/java/com/google/android/material/imageview/ShapeableImageView.java#L125) but that then seems to break the masking done [here](https://github.com/material-components/material-components-android/commit/20d2efed827f622e83b5a32a17835bb52b723ffb#diff-c619ca095752b93f73de0774131f8248R83).

simekadam:
I was also able to "fix" this by removing not disable the hw buffer rendering [here](https://github.com/material-components/material-components-android/blob/d8a7a9afd58aea8e13a0d71d8ea693c7f7962395/lib/java/com/google/android/material/imageview/ShapeableImageView.java#L125) but that then seems to break the masking done [here](https://github.com/material-components/material-components-android/commit/20d2efed827f622e83b5a32a17835bb52b723ffb#diff-c619ca095752b93f73de0774131f8248R83).

ymarian:
This is only works for circles. Does adding more padding not work for your case? We are considering adding an additional attribute for insets

simekadam:
Padding indeed works well, but I found it a bit iffy to be having to set both padding and stroke and then having to maintain these two in sync. I have tried some cut corners, but indeed

```
 <item name="cornerFamily">cut</item>
```
is still broken.

PR #1645, opened 2020-08-22
[TextInputLayout] Hint always collapsed capability

### LINKED ISSUE #1644: [TextInputLayout] Add the capability to have the hint always collapsed
I've got a requirement to not show the hint in an expanded manner at any point but always at the top left above the edittext. Unfortunately there does not seem to be any way of doing this right now.

Just an attribute that would keep it collapsed at all times

You can cheat and have a custom edittext report its text as non empty just to the `TextInputLayout` but that is a very hacky solution

I'll put up a pull request with what I have in mind

vkondrav:
closes #1644


matpag:
I think you should provide a setter for the property

PR #1656, opened 2020-08-26
[Theme] Color utility to resolving material theme color attrs programmatically

gabrielemariotti:
Added a class utility to resolve material theme color attrs programmatically using the `MaterialColors`.

### DIFF
--- a/lib/java/com/google/android/material/color/MaterialColorUtils.java
+++ b/lib/java/com/google/android/material/color/MaterialColorUtils.java
@@ -1,0 +1,282 @@
+/*
+ * Copyright 2020 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


lib/java/com/google/android/material/color/MaterialColorUtils.java
```diff
@@ -0,0 +1,282 @@
...
+package com.google.android.material.color;
+
+import android.content.Context;
+import android.view.View;
+import androidx.annotation.ColorInt;
+import androidx.annotation.NonNull;
+import com.google.android.material.R;
+
+/** Utility methods for {@link MaterialColors}. */
+public class MaterialColorUtils {
```
drchen, line 26: I think it will be better to define this as an enum class under MaterialColors, say, MaterialColors.Color.PRIMARY, etc., and provide getColor() method for the enum class.

drchen (changes requested):

thatomoekoa (approved):

PR #1661, opened 2020-08-26
[TextInputLayout] Fixed wrong hint position when an error is shown and field is empty and unfocused

### LINKED ISSUE #1275: TextInputLayout  error hint animaiton
When error is set on TextInputLayout if the TextInputEditText is empty and  there is no focus at the field the hint label doesn't slide down. (see screenshot). If there is no error, the hint animation works fine (the label slides down )
Screen Shot 2020-05-05 at 11 46 33

gabrielemariotti:
It closes #1275.

**Currently** as described in the issue when an error is shown and the field is empty and unfocused:
-  the position of the hint text is collapsed

Schermata 2020-08-26 alle 23 30 54

Following the https://material.io/components/text-fields the hint text should be expanded with the error color.

Schermata 2020-08-26 alle 23 32 32

It can be reproduced with the catalog demo.

**With this PR**:

Unfocused, with an error and without text:
Schermata 2020-08-26 alle 23 27 20

Focused, with an error and with text:
Schermata 2020-08-26 alle 23 26 56

Unfocused, with an error and with text:
Schermata 2020-08-26 alle 23 27 09

### DIFF
--- a/lib/java/com/google/android/material/textfield/TextInputLayout.java
+++ b/lib/java/com/google/android/material/textfield/TextInputLayout.java
@@ -1484,13 +1484,14 @@ public class TextInputLayout extends LinearLayout {
       collapsingTextHelper.setExpandedTextColor(ColorStateList.valueOf(disabledHintColor));
     } else if (errorShouldBeShown) {
       collapsingTextHelper.setCollapsedTextColor(indicatorViewController.getErrorViewTextColors());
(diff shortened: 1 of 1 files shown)


leticiarossi:
Thanks! I think the expanded label should remain gray (the specs need to be updated), but I can change that when I pull the PR in!

PR #1678, opened 2020-09-02
[MaterialToolbar] Add Shape and shape customization to MaterialToolbar

alorma:
#### Don't forget:

I've recently seen that `MaterialToolbar` uses MaterialShapeDrawable to customize the color on dark mode.

But unfortunately it has no effect if we want to customize the shape of it, as we needed to use a custom xml drawable, or set `MaterialShapeDrawable` manually.

This PR adds shape customization to `MaterialToolbar`

## Screenshots
Rounded | Cut | Dark
---|---|---
 |  | 

Questions:

* How can I make it work with theme configuration dialog fragment?

### DIFF
--- a/lib/java/com/google/android/material/appbar/MaterialToolbar.java
+++ b/lib/java/com/google/android/material/appbar/MaterialToolbar.java
@@ -16,25 +16,25 @@
 
 package com.google.android.material.appbar;
 
-import com.google.android.material.R;
-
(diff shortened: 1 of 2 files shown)


alorma:
Then how can we implement something like this:

image

ymarian:
Good point. We just don't want it to be part of the shape system. However we still allow:

`(toolbar.background as MaterialShapeDrawable)`

So you can customize it

Issue #1570, opened 2020-08-06
[BottomSheetDialogFragment] Since Ver. 1.2.0 different dismiss behavior


bobekos:
**Description:**
Since Version 1.2.0 and above the dismiss gesture detection is working differently then in 1.1.0.
Where I could easily dismiss the dialog in version 1.1.0, i cannot do it since version 1.2.0. First I thought about the new attribute "gestureInsetBottomIgnored" but that seems to have nothing to do with it. In version 1.2.0 the callback "close the dialog" seems to depend on a certain length of the dismiss (swipe down) gesture. Difficult to describe, best take a look at the video below.

**Expected behavior:**
Dismiss gesture is working like in the Version 1.1.0.

Video Version 1.1.0 https://we.tl/t-DOOuTP8TbB
Video Version 1.2.0 https://we.tl/t-w4fyZ8OjMo

[video_demo.zip](https://github.com/material-components/material-components-android/files/5033200/video_demo.zip)

**Source code:**
You must do nothing, no special themes or other settings. This simple thing is already working differently between the two versions.

```


//just normal dialog (no special theme)
class BottomSheetTest : BottomSheetDialogFragment() {

    override fun onCreateView(
        inflater: LayoutInflater,
        container: ViewGroup?,
        savedInstanceState: Bundle?
    ): View? = inflater.inflate(R.layout.fragment_test, container, false)

}

//default show
BottomSheetTest().show(supportFragmentManager, TAG)
```

**Android API version:**
Android 10

**Material Library version:**
1.1.0 and 1.2.0

**Device:**
Pixel 3a

bobekos:
It seems like this method here pass other peekHeight to the result:

```
boolean shouldHide(@NonNull View child, float yvel) {
    if (skipCollapsed) {
      return true;
    }
    if (child.getTop() < collapsedOffset) {
      // It should not hide, but collapse.
      return false;
    }
    int peek = calculatePeekHeight();
    final float newTop = child.getTop() + yvel * HIDE_FRICTION;
    return Math.abs(newTop - collapsedOffset) / (float) peek > HIDE_THRESHOLD;
  }
```

In the Version 1.1.0 the peek was always 0 when peek_height was set to auto. So this means that the result was always true.

dsn5ft:
Hi @bobekos, thanks for sharing the code snippet, but I'm unable to reproduce this. Can you share a minimal sample app with the full setup that reproduces the issue?

bobekos:
[...]

PR #1697, opened 2020-09-09
[BottomSheet] Add the possibility to set the hide threshold of the bottom sheet

bobekos:
It would be nice to have the possibility to set the hide threshold of the bottom sheet when the user want to dismiss it. This pull request allows you to do that.

Related to the issue #1570

### DIFF
--- a/lib/java/com/google/android/material/bottomsheet/BottomSheetBehavior.java
+++ b/lib/java/com/google/android/material/bottomsheet/BottomSheetBehavior.java
@@ -188,7 +188,7 @@ public class BottomSheetBehavior<V extends View> extends CoordinatorLayout.Behav
 
   private static final int SIGNIFICANT_VEL_THRESHOLD = 500;
 
-  private static final float HIDE_THRESHOLD = 0.5f;
+  private static final float DEFAULT_HIDE_THRESHOLD = 0.5f;
 
   private static final float HIDE_FRICTION = 0.1f;
 
@@ -214,6 +214,9 @@ public class BottomSheetBehavior<V extends View> extends CoordinatorLayout.Behav
 
   /** True if Behavior has a non-null value for the @shapeAppearance attribute */
   private boolean shapeThemingEnabled;
+
+  /** Threshold to control the hidden state trigger */
+  private float hideThreshold;
 
   private MaterialShapeDrawable materialShapeDrawable;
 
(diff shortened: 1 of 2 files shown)


raajkumars:
I am not sure whether this is something we would want the app developers to change manually. Instead this should be handled dynamically by the motion subsystem.

hunterstich:
Hey @bobekos,

I'm speaking with our design/interaction team about whether we want to allow customization of the threshold from app to app. Is there any additional info you can give us on why you need this? What makes you want to change the threshold? Does the current threshold not feel right? Or are you trying to achieve something different using bottom sheet?

bobekos:
Hi sorry for the late response. Not the current threshold is feeling bad but yeah the 'new' dismiss behaviour is bad. Since 1.2.0 the dismiss behaviour has changed. I already reported that as issue (see #1570) but the change which got fixed has affect only on bottomsheet where the peekheight was smaller than the bottomsheet itself.

But you just need to compare the dismiss behaviour before and after the version 1.2.0. For example the demo app and the theme bottomsheet (where you can change the app theme) or better the new version of google play app. I think they also use material library above 1.1.0. Just try to dismiss the bottomsheet when you long clicked a app. It is not normal in my opinion.

So in case you think the behaviour is correct i decided to create this pull request to add the possibility make the behaviour the same like in version 1.1.0

hunterstich:
@bobekos got it. 

So, do you mean e944d1b2a6ee5d9d5a338de0c0061f7b02790f77 _did_ fix one problem but bottom sheet is still too difficult to dismiss is certain scenarios? The Google Play Store's bottom sheet looks like it just hasn't updated to use the fix from e944d1b2a6ee5d9d5a338de0c0061f7b02790f77. For the catalog, the theme switcher bottom sheet seems to work fine for me when built off the current `master`.

Design suggests we don't open up the option to customize the hide threshold of bottom sheet in order to keep behavior consistent and predictable across apps. I think if we need to tweak the behavior, we should do so with A) UX design's input/spec and B) make the changes are made internally so bottom sheet behavior remains consistent from app to app.

Thoughts?

bobekos:
@hunterstich 

But have you really compared the two versions? Here i maked two videos one from the current master and one from the release-1.1 branch -> 
[master.zip](https://github.com/material-components/material-components-android/files/5391470/master.zip)
[1.1.zip](https://github.com/material-components/material-components-android/files/5391475/1.1.zip)

I mean this cannot be overlooked. The question is now if the current behavior is correct. In my opinion it is not. So I thought that with my pull request everyone can decide for himself. But yeah you are right the behavior should be consistent over all apps and shouldn't be changed in this "dramatically" way.

What should i say? :D At the end you can easily block this pull request if you think that now the behavior is correct. I mean who says that the old one is also the correct one?

hunterstich:
@bobekos I see what you mean.

Even if we do update the behavior, it sounds like design is in favor of keeping the threshold private so all apps share the same, predictable behavior. 

I'm going to go ahead and close this in favor of a feature request asking design to look into what the correct gesture spec should be - https://github.com/material-components/material-components-android/issues/1808

PR #1711, opened 2020-09-11
[Snackbar] Changed dimens according to spec

gabrielemariotti:
According to the spec:

Schermata 2020-09-12 alle 00 54 39

Currently:

- The text has a padding of `12dp` from the `SnackbarContentLayout`  inherited from `Widget.Design.Snackbar` + `12dp` from `Widget.MaterialComponents.Snackbar.TextView`.
- The action button has a padding of `12dp` from the `SnackbarContentLayout`.

Schermata 2020-09-12 alle 00 41 31

With this PR:

Schermata 2020-09-12 alle 00 26 32

It can be tested with the demo catalog.


Monday1908 (changes requested):
API36

PR #1769, opened 2020-10-01
[ShapeableImageView] Call invalidateOutline when shape is changed

### LINKED ISSUE #1489: [ShapeableImageView] Padding is applied to background, unlike normal ImageView
**Description:** In ShapeableImageView, `android:padding` gets applied to both the image itself and the shaped background. This is unusual compared to most views (though maybe typical for Shapeables, not sure).

**Expected behavior:**
What the normal ImageView does: pads the image, but not the background.
Top: ShapeableImageView with padding. Bottom: normal ImageView with padding.
image

It seems like the current behavior should be achieved by margin rather than padding, and that padding should leave the background alone, like it does in most Views.

**Source code:**
```xml
<ShapeableImageView
    android:layout_width="48dp"
    android:layout_height="48dp"

    android:background="?colorSecondary"
    android:padding="8dp"

    app:shapeAppearance="?shapeAppearanceSmall"
    app:srcCompat="@drawable/my_icon" />
```

**Android API version:** 21-29

**Material Library version:** 1.2.0-rc01

**Device:** Pixel 3 emulator

Looking at the ShapeableImageView source, it appears to be clear where the shape is set to honor the padding. I'd like to know if the current behavior is expected for some reason. If it is, I'd like to know if the behavior I want can be accommodated by some new attribute or something.

If there's a clear path forward I'd be happy to open a PR. And if that PR could potentially land in 1.2, I'd be happy to open a PR quickly. :) Let me know!

drewhamilton:
I discovered this while using the included Catalog app to look into #1489. But this PR does *not* that issue.

Previously, ShapeableImageView was not updating its shadow outline provider when its shape was changed at runtime. This would lead to shadows around the original shape while the view had the appearance of the new shape.

This could not be observed in the Catalog app, because the Catalog app was incorrectly using `app:elevation`, an invalid attribute; rather than `android:elevation`. So the views in the Catalog had no elevation at all and thus the odd elevation behavior was not visible.

This PR fixes the issue by calling `invalidateOutline` at the end of `setShapeAppearanceModel`, and applies `android:elevation` in the Catalog app to view the result.

### DIFF
--- a/catalog/java/io/material/catalog/imageview/res/layout/catalog_imageview.xml
+++ b/catalog/java/io/material/catalog/imageview/res/layout/catalog_imageview.xml
@@ -22,6 +22,7 @@
   android:id="@+id/main_viewGroup"
   android:layout_width="match_parent"
   android:layout_height="wrap_content"
+  android:clipChildren="false"
   android:padding="@dimen/cat_imageview_padding"
   android:orientation="vertical">
 
@@ -66,7 +67,7 @@
     app:strokeColor="?attr/colorSecondary"
     app:shapeAppearance="?attr/shapeAppearanceMediumComponent"
     app:strokeWidth="1dp"
-    app:elevation="5dp"
+    android:elevation="5dp"
     android:padding="8dp"
     app:srcCompat="@drawable/dog_image" />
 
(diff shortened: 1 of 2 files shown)


drewhamilton:
@googlebot I signed it!

PR #1786, opened 2020-10-07
[Slider] Invalidate after new ColorStateList and return early if same

lernerb:
When using the Slider component:
* Trigger invalidate() call after a new ColorStateList is applied 
* Exit early if the the fill color (ColorStateList) is the same as the currently set fill color, to avoid extra draws.

There's no issue currently filed for this, but when using the `Slider.setThumbTintList` in a project i'm working on, I noticed that the fill color doesn't **always** update when setting it - this should fix it!


raajkumars (approved):
LGTM

PR #1790, opened 2020-10-09
[MaterialButton] Added methods to apply horizontal insets

### LINKED ISSUE #1770: [MaterialButton] Change horizontal button insets programmatically
Follow-up of #526 that has been closed by adding vertical insets APIs but no APIs for horizontal insets yet.

gabrielemariotti:
It closes #1770 

This PR adds methods to support setting horizontal insets programatically.

### DIFF
--- a/lib/java/com/google/android/material/button/MaterialButtonHelper.java
+++ b/lib/java/com/google/android/material/button/MaterialButtonHelper.java
@@ -432,4 +432,43 @@ class MaterialButtonHelper {
     return insetTop;
   }
 
+
+  public void setInsetLeft(@Dimension int newInsetLeft) {
+    setHorizontalInsets(newInsetLeft, insetRight);
+  }
+
+  public int getInsetLeft() {
+    return insetLeft;
+  }
+
+  public void setInsetRight(@Dimension int newInsetRight) {
+    setHorizontalInsets(insetLeft, newInsetRight);
+  }
+
+  private void setHorizontalInsets(@Dimension int newInsetLeft, @Dimension int newInsetRight) {
+    // Store padding before setting background, since background overwrites padding values
+    int paddingStart = ViewCompat.getPaddingStart(materialButton);
+    int paddingTop = materialButton.getPaddingTop();
+    int paddingEnd = ViewCompat.getPaddingEnd(materialButton);
+    int paddingBottom = materialButton.getPaddingBottom();
+    int oldInsetLeft = insetLeft;
+    int oldInsetRight = insetRight;
+    insetLeft = newInsetLeft;
+    insetRight = newInsetRight;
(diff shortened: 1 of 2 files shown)


lib/java/com/google/android/material/button/MaterialButtonHelper.java
```diff
@@ -432,4 +432,43 @@ public int getInsetTop() {
...
+    int paddingBottom = materialButton.getPaddingBottom();
+    int oldInsetLeft = insetLeft;
+    int oldInsetRight = insetRight;
+    insetLeft = newInsetLeft;
+    insetRight = newInsetRight;
+    if (!backgroundOverwritten) {
+      updateBackground();
+    }
+    // Set the stored padding values
+    ViewCompat.setPaddingRelative(
```
hunterstich, line 462: If we're using `setPaddingRelative` here, don't we want to expose `setInsetStart` and `setInsetEnd` since internally `setPaddingRelative` will swap the left/right values depending on the layout direction? 

If we want to support setting `left` and `right` values, we'd need to change this to `ViewCompat.setPadding()`, if I understand correctly.

hunterstich (commented):
Hey @gabrielemariotti,

Thanks for the PR! Added a few comments. Let me know what you think.

lib/java/com/google/android/material/button/MaterialButtonHelper.java
gabrielemariotti: I agree. It is better `setInsetStart` and `setInsetEnd`.

gabrielemariotti: @hunterstich In the current code in the `loadFromAttributes` you are using `setPaddingRelative` adding `paddingStart + insetLeft`. Let me know what you prefer. Maybe the best solution is to use `insetLeft`, `paddingLeft` and `ViewCompat.setPadding()` everywhere.
https://github.com/material-components/material-components-android/blob/5e3ccdc73cc9f71bbba425ea838619468b6195f9/lib/java/com/google/android/material/button/MaterialButtonHelper.java#L128-L133

hunterstich: Oh good catch 

I agree, it would be nice to correct/avoid the mix of paddingStart|End and left|right. Using `paddingLeft|Right` + `insetLeft|Right` and `ViewCompat.setPadding` everywhere sounds like a good solution to me. Off the top of my head, I can't think of any behavior that this might break, but once you update `MaterialButtonHelper` and I pull it in and run all our test, we can double check.

In the future (or now if you want...), we can add in support for `insetStart|End` and `ViewCompat.setPaddingRelative`.

daeni01 (approved):

Issue #832, opened 2019-12-16
[Slider] Touch event being handled incorrectly when used inside a scrolling container


rubensousa:
**Description:** Unlike SeekBar, Slider doesn't check if it's inside a scrolling container before starting the drag process inside onTouchEvent. This causes scroll events to be wrongly intercepted  
by Slider.

| Current behavior | Expected behavior |
| ------------- | ------------- |
| ![scroll_bug](https://user-images.githubusercontent.com/10662096/70946852-d9fd9000-204f-11ea-850f-0e3f48b57570.gif) | ![scroll_fix](https://user-images.githubusercontent.com/10662096/70946882-e84bac00-204f-11ea-9d99-ec32d082ae15.gif)

I'm currently working on a PR for this

PR #1847, opened 2020-10-31
[Slider] Fix slider doesn't slide in Horizontal Recycler View

### LINKED ISSUE #1843: [Slider] Slider doesn't slide in Horizontal RecyclerView
**Description:** Slider doesn't slide in Horizontal RecyclerView.

**Expected behavior:** Slider should slide in Horizontal RecyclerView.

**Android API version:** 30

**Material Library version:** 1.2.1

**Device:** Emulator Pixel 3

Subuday:
Closes #1843

### DIFF
--- a/lib/java/com/google/android/material/slider/BaseSlider.java
+++ b/lib/java/com/google/android/material/slider/BaseSlider.java
@@ -1699,6 +1699,10 @@ abstract class BaseSlider<
         break;
       case MotionEvent.ACTION_MOVE:
         if (!thumbIsPressed) {
+          // Check if we're trying to scroll instead of dragging this Slider
+          if (isInScrollingContainer() && abs(x - touchDownX) < scaledTouchSlop) {
+            return false;
+          }
           getParent().requestDisallowInterceptTouchEvent(true);
           onStartTrackingTouch();
         }
--- a/lib/java/com/google/android/material/slider/BaseSlider.java
+++ b/lib/java/com/google/android/material/slider/BaseSlider.java
@@ -1679,7 +1679,7 @@ abstract class BaseSlider<
 
         // If we're inside a scrolling container,
         // we should start dragging in ACTION_MOVE
-        if (isInScrollingContainer()) {
+        if (isInVerticalScrollingContainer()) {
           break;
         }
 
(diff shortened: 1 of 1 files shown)


lib/java/com/google/android/material/slider/BaseSlider.java
```diff
@@ -1699,10 +1699,6 @@ public boolean onTouchEvent(@NonNull MotionEvent event) {
         break;
       case MotionEvent.ACTION_MOVE:
         if (!thumbIsPressed) {
-          // Check if we're trying to scroll instead of dragging this Slider
-          if (isInScrollingContainer() && abs(x - touchDownX) < scaledTouchSlop) {
-            return false;
-          }
```
dsn5ft, line 1705: @Subuday I don't think we can just remove this. It was added to fix issues with vertical scrolling:

https://github.com/material-components/material-components-android/issues/832
https://github.com/material-components/material-components-android/pull/833

Maybe instead of `isInScrollingContainer()` we could use `View#canScrollVertically` and `View#canScrollHorizontally` on the scrolling parent somehow for this x touch check.

cc: @cketcham

cketcham: I'm wondering if it's even a good idea to put a slider in a horizontally scrolling view? It wouldn't be possible to distinguish between an attempt to change the value on the slider or scroll. This would force the user to touch outside the Slider to perform a scroll on the parent.

dsn5ft: That's a good point, but I think the same thing could be said of putting a Slider in a TabLayout/ViewPager. I feel like it should be possible, but to swipe or scroll the user just has to touch outside the Slider like you said.

And maybe a small optimization could be that if the Slider is at its full left/right position and can't move its thumb any more depending on the direction of the swipe/scroll, then that would result in a swipe/scroll.

Subuday: @dsn5ft Yes, it's a good idea to check if it's in vertical or horizontal scrolled view.

Subuday: @cketcham As for my case, it's dragging card like in Badoo which implemented using RecyclerView has slider inside.

lib/java/com/google/android/material/slider/BaseSlider.java
```diff
@@ -2035,20 +2039,30 @@ private void invalidateTrack() {
...
-    ViewParent p = getParent();
-    while (p instanceof ViewGroup) {
-      if (((ViewGroup) p).shouldDelayChildPressedState()) {
+    return isInVerticalScrollingContainer() || isInHorizontalScrollingContainer();
+  }
+
+  private boolean isInVerticalScrollingContainer() {
+    ViewParent parent = getParent();
+    while (parent instanceof View) {
+      if (((View) parent).canScrollVertically(1) ||
```
cketcham, line 2049: Maybe here and in isInHorizontalScrollingContainer() we should also check shouldDelayChildPressedState(). Even if the parent could scroll, if shouldDelayChildPressedState() is false then I think that it should behave like it isn't in a scrolling container

cketcham: In fact, maybe it's good enough to just remove `isInScrollingContainer()` and use `isInVerticalScrollingContainer()` with that logic. The other call to `isInScrollingContainer()` in `ACTION_DOWN` should also be `isInVerticalScrollingContainer()`. No point in waiting until `ACTION_MOVE` if we're just going to capture the touch event anyways.

So maybe remove `isInScrollingContainer()` and `isInHorizontalScrollingContainer()` and just use `isInVerticalScrollingContainer()` with a check for `shouldDelayChildPressedState()`

cketcham (changes requested):

cketcham (approved):
I think this makes sense. At least you should be able to change the value of the slider if you're in a horizontally scrolling view.

PR #1887, opened 2020-11-19
[Snackbar] Fixes remaining NPEs from #917

vonox7:
There could happen NPEs when dismissing Snackbars that have not been attached to the view yet or fading in Snackbars that have already been dismissed. https://github.com/material-components/material-components-android/issues/917#issuecomment-576577838 describes this issue, which can be reproduced on Android 7.0, Android 8.0.0, Android 9.0. 

To test this PR, I created snackbars that were dismissed within 0-200ms and within 0-200ms created again. Without this PR, the app crashes (on e.g. an Android 8.0.0 emulator) within some seconds. With this PR, the app can run for at least 20 minutes without a crash.

Therefor the following exception does not happen any more:
```
java.lang.NullPointerException: Attempt to invoke virtual method 'void android.view.View.invalidate(boolean)' on a null object reference
    at android.view.View.setFlags(View.java:13338)
    at android.view.View.setVisibility(View.java:9382)
    at com.google.android.material.snackbar.BaseTransientBottomBar$9.run(BaseTransientBottomBar.java:848)
    at android.os.Handler.handleCallback(Handler.java:789)
    at android.os.Handler.dispatchMessage(Handler.java:98)
    at android.os.Looper.loop(Looper.java:164)
    at android.app.ActivityThread.main(ActivityThread.java:6541)
    at java.lang.reflect.Method.invoke(Native Method)
    at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:240)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:767)
```

Issue #917, opened 2020-01-17
[BaseTransientBottomBar] 1.1.0-rc01 causes View.invalidate(boolean) NullPointerException


Naitbit:
**Description:** 
After updating from Material 1.0.0 to 1.1.0-rc01 I saw multiple reports of crashes with BaseTransientBottomBar (and no app specific code in stacktrace).
Stacktraces look like:
Huawei Android 7
```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'void android.view.View.invalidate(boolean)' on a null object reference
       at android.view.View.setFlags + 11550(View.java:11550)
       at android.view.View.setVisibility + 8046(View.java:8046)
       at com.google.android.material.snackbar.BaseTransientBottomBar$9.run + 849(BaseTransientBottomBar.java:849)
       at android.os.Handler.handleCallback + 761(Handler.java:761)
       at android.os.Handler.dispatchMessage + 98(Handler.java:98)
       at android.os.Looper.loop + 156(Looper.java:156)
       at android.app.ActivityThread.main + 6617(ActivityThread.java:6617)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run + 942(ZygoteInit.java:942)
       at com.android.internal.os.ZygoteInit.main + 832(ZygoteInit.java:832)
```
Samsung Android 9
```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'void android.view.View.invalidate(boolean)' on a null object reference
       at android.view.View.setFlags + 15721(View.java:15721)
       at android.view.View.setVisibility + 10789(View.java:10789)
       at com.google.android.material.snackbar.BaseTransientBottomBar$9.run + 849(BaseTransientBottomBar.java:849)
       at android.os.Handler.handleCallback + 873(Handler.java:873)
       at android.os.Handler.dispatchMessage + 99(Handler.java:99)
       at android.os.Looper.loop + 214(Looper.java:214)
       at android.app.ActivityThread.main + 7094(ActivityThread.java:7094)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run + 494(RuntimeInit.java:494)
       at com.android.internal.os.ZygoteInit.main + 975(ZygoteInit.java:975)
```
Xiaomi Android 9
```
[...]

PR #1938, opened 2020-12-18
[Snackbar] Fix last remaining NPE from #917

vonox7:
The commit in PR https://github.com/material-components/material-components-android/pull/1887 fixed all NPEs from issue #917. But the commit https://github.com/material-components/material-components-android/commit/ebd322314719fb65bdfa7299274ed4662a29d88c did only apply 3 of the 4 changes. Ommiting the first change seems logical when looking at the code, so I understand @leticiarossi why you did not apply the first null-check.
But: Without the null-check the NPE mentioned in #917 still occours. We also could not believe it, this is why we made in the last days several releases and testing it “in the field” via jitpack. As result:

1. When releasing with https://github.com/vonox7/material-components-android/commit/331d064eb02cb4ff1a2b84f274b8d2c98848e9d4, no NPEs happened any more.
2. When releasing with https://github.com/material-components/material-components-android/commit/c6deb0639763409b132f79dd7039158e69dc6ffb most of the NPEs were fixed, but not all of them. An example crash is:
```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'void android.view.View.invalidate(boolean)' on a null object reference
       at android.view.View.setFlags(View.java:15732)
       at android.view.View.setVisibility(View.java:10800)
       at com.google.android.material.snackbar.BaseTransientBottomBar$10.run(BaseTransientBottomBar.java:896)
       at android.os.Handler.handleCallback(Handler.java:873)
       at android.os.Handler.dispatchMessage(Handler.java:99)
       at android.os.Looper.loop(Looper.java:214)
       at android.app.ActivityThread.main(ActivityThread.java:7050)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:494)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:965)
```
3. When releasing again with https://github.com/vonox7/material-components-android/commit/331d064eb02cb4ff1a2b84f274b8d2c98848e9d4, no NPEs happened any more.
4. We could not believe it, so we reverted the single other change that happened in BaseTransientBottomBar in a release with https://github.com/zigavehovec/material-components-android/commit/f2e475675e74e5195e85d0f30770f4469ac7b685: The same crashes as in release from step 2 happened.
5. We did now a final release with https://github.com/zigavehovec/material-components-android/commit/d9057289bdaae87ffbf9f729b587ffd695feb8f0. No NPEs happen any more.

I must confess that I am not 100% sure why exactly this check fixes the NPE, but i can report that without this null-check a lot of crashes occour, and with this check the material-components-android library doesn’t crash any more on our use-cases. Therefore I ask politely to also add this NPE-check to the library so that it won’t crash any more on Android 8 and Android 9 devices.

### DIFF
--- a/lib/java/com/google/android/material/snackbar/BaseTransientBottomBar.java
+++ b/lib/java/com/google/android/material/snackbar/BaseTransientBottomBar.java
@@ -773,7 +773,9 @@ public abstract class BaseTransientBottomBar<B extends BaseTransientBottomBar<B>
 
       // Set view to INVISIBLE so it doesn't flash on the screen before the inset adjustment is
       // handled and the enter animation is started
-      view.setVisibility(View.INVISIBLE);
+      if (view.getParent() != null) {
+        view.setVisibility(View.INVISIBLE);
+      }
       targetParent.addView(this.view);
     }
 


ikim24 (approved):

ikim24:
Hi, we've had to roll this back as it was causing a flicker in the snackbar. I think we'll need a better of understanding of what flow causes these NPEs before figuring out how to handle them.

lib/java/com/google/android/material/snackbar/BaseTransientBottomBar.java
```diff
@@ -773,7 +773,9 @@ public void run() {
 
       // Set view to INVISIBLE so it doesn't flash on the screen before the inset adjustment is
       // handled and the enter animation is started
-      view.setVisibility(View.INVISIBLE);
+      if (view.getParent() != null) {
```
drchen, line 776: This will never be executed give L764. 

Shouldn't we just switch the order of L776 and L777 to fix the issue?

drchen (changes requested):

vonox7:
Closing because `if (view.getParent() != null) {` check is in master: https://github.com/material-components/material-components-android/blob/release-1.8/lib/java/com/google/android/material/snackbar/BaseTransientBottomBar.java#L915

Issue #1922, opened 2020-12-11
[AppBarLayout] Add setExpandedEx() to show minHeight


EaseTheWorld:
When 'minHeight' is set and scrollFlags includes 'enterAlwaysCollapsed',
scroll up shows only minHeight. ('Collapsed' state)

and setExpanded(true) shows full height. ('Expanded' state)

but setExpanded(false) hide appbar completely.('Hidden' state)
so I can't go back to 'Collapsed' state programmatically.

I suggest setExpandedMode() to support these Hidden/Collapsed/Expanded state.

I tested by modifying BaseBehavior.onLayoutChild(),
  - adding PENDING_ACTION_EXPAND_MIN_HEIGHT,
  - offset=abl.getDownNestedPreScrollRange() - abl.getUpNestedPreScrollRange()
    will do.
I know the name is confusing. but PENDING_ACTION_COLLAPSED is already used for hidden state. :(

If owner agrees this feature and decide interface, I will upload PR.

EaseTheWorld:
My suggestion for interface is
setExpandedEx(int mode, boolean animate)
and expose three mode
  MODE_EXPAND_HIDE
  MODE_EXPAND_COLLAPSED
  MODE_EXPAND_EXPANDED

PR #1985, opened 2021-01-15
[AppBarLayout] Introduce expand api

EaseTheWorld:
This solves #1922 in more general way.

- expand(AppBarLayout.EXPAND_NONE) = setExpanded(false)
- expand(AppBarLayout.EXPAND_ALL) = setExpanded(true)
- expand(AppBarLayout.EXPAND_ENTERED) : new feature. same result as scroll down and entering
- expand(positive value) : new feature. shows the given height from the bottom of appbarlayout


EaseTheWorld:
@googlebot I signed it!

PR #1997, opened 2021-01-18
[MaterialButton] fix(button): respect intrinsic icon width when iconGravity is view{Start/End}

### LINKED ISSUE #1996: [MaterialButton] Text misalignment with icon with viewStart alignment
**Description:** 
TextAlignment is being offset based on the icon intrinsicWidth when set as `viewStart` causing the button text to not be properly centered. This is more prominently seen when the button's width is set at match_constraints as opposed to wrap_content.

**Expected behavior:** 
Button text to remain horizontally centered regardless of icon placement.

**Android API version:** 
API 29 and 30 tested so far

**Material Library version:** 
'com.google.android.material:material:1.2.1'

**Device:** 
Pixel 3 XL and Pixel 4 XL (so far)

bmc08gt:
To ensure the button text remains centered in the container, the icon's intrinsicWidth
needs to be taken into consideration.

closes #1996

### DIFF
--- a/lib/java/com/google/android/material/button/MaterialButton.java
+++ b/lib/java/com/google/android/material/button/MaterialButton.java
@@ -766,6 +766,7 @@ public class MaterialButton extends AppCompatButton implements Checkable, Shapea
     // Forced icon update
     if (needsIconReset) {
       resetIconDrawable();
+      adjustTextFromIcon();
       return;
     }
 
@@ -781,6 +782,17 @@ public class MaterialButton extends AppCompatButton implements Checkable, Shapea
 
     if (hasIconChanged) {
       resetIconDrawable();
+      adjustTextFromIcon();
+    }
+  }
+
+  private void adjustTextFromIcon() {
+    if (icon != null) {
+      if (isIconViewStart()) {
+          setIconPadding(-icon.getIntrinsicWidth());
+      } else if (isIconViewEnd()) {
+        setIconPadding(icon.getIntrinsicWidth());
+      }
     }
   }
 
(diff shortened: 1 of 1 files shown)


bmc08gt:
@googlebot I signed it!

DearDhruv (approved):

SupremeFergs (approved):

maxfie1d (approved):

lib/java/com/google/android/material/button/MaterialButton.java
```diff
@@ -781,6 +782,17 @@ private void updateIcon(boolean needsIconReset) {
 
     if (hasIconChanged) {
       resetIconDrawable();
+      adjustTextFromIcon();
```
drchen, line 785: Put this inside resetIconDrawable()?

lib/java/com/google/android/material/button/MaterialButton.java
```diff
@@ -781,6 +782,17 @@ private void updateIcon(boolean needsIconReset) {
...
+    }
+  }
+
+  private void adjustTextFromIcon() {
+    if (icon != null) {
+      if (isIconViewStart()) {
+          setIconPadding(-icon.getIntrinsicWidth());
+      } else if (isIconViewEnd()) {
+        setIconPadding(icon.getIntrinsicWidth());
```
drchen, line 794: Why do we need different logic here? Isn't the purpose "consuming" the icon width with a negative padding?

drchen (changes requested):
I think the original behavior is WAI, according to the Material design spec:
https://material.io/components/buttons#interactive-demo

TBH I don't think we can just change it since it's an existing behavior conforming to the spec and being used everywhere. 

If you really need this behavior, I'll suggest you just customizing  your MaterialButton to implement the behavior.

bmc08gt:
Yep I've done just that for a project.

PR #2094, opened 2021-02-23
[BottomSheetBehavior] add minFitToContentsOffset

emakar:
I needed BottomSheetBehavior to work with top margin being set to its view.
Currently this is buggy and gives various card jumps.
I also need this behavior with fitToContents = true

While supporting special margin treatment I've realised another option is possible - adding new property, i.e. minFitToContentsOffset

So, I've included both commits and currently prefer the one with minFitToContentsOffset


emakar:
I've done more testing and left only special top margin treatment

This is ready for review

Lingviston:
One update on this one?

PR #1871, opened 2020-11-14
[ShapeableImageView] Support contentPadding

drewhamilton:
Closes #1489. Adds `contentPadding` APIs to ShapeableImageView. This is achieved by communicating the sum of padding and contentPadding as a single value to the superclass (AppCompatImageView), but tracking them separately in this class. This maintains the existing behavior of ShapeableImageView padding, which pads the background, but takes advantage of the fact that AppCompatImageView does _not_ pad the background.

Frankly the internal logic is a nightmare, and hard to wrap my head around even after writing it. But it's the only way I can get it to work while maintaining behavioral compatibility, and it seems to work as expected with some manual testing.

I'm happy to add tests but I can't figure out how they work in this project. Is there something special I need to do to get the IDE to recognize the test sources?

PR #2280, opened 2021-07-01
[ShapeableImageView] Fix padding bugs

### LINKED ISSUE #2063: [ShapeableImageView] contentPadding not working when set via Kotlin
**Description:** 
When setting `contentPadding` on a `ShapeableImageView` in code (Kotlin or Java), it does not work as expected. Setting `app:contentPadding` via `XML` works fine.

**Expected behavior:**
Both elements should match and be like the top element.
image

**Source code:**

`MainActivity.kt`
```kotlin
package com.example.bug

import androidx.appcompat.app.AppCompatActivity
import android.os.Bundle
import android.util.TypedValue
import com.google.android.material.imageview.ShapeableImageView

class MainActivity : AppCompatActivity() {
    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)
        setContentView(R.layout.activity_main)

        val px = TypedValue.applyDimension(
            TypedValue.COMPLEX_UNIT_DIP,
            8f,
            resources.displayMetrics
        ).toInt()

        findViewById<ShapeableImageView>(R.id.kotlin)?.setContentPadding(px, px, px, px)
    }
}
```

`activity_main.xml`
```xml
<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
    tools:context=".MainActivity">

    <com.google.android.material.imageview.ShapeableImageView
        android:id="@+id/xml"
        android:layout_width="50dp"
        android:layout_height="50dp"
        android:background="@color/black"
        app:contentPadding="8dp"
        app:layout_constraintBottom_toTopOf="@id/kotlin"
        app:layout_constraintEnd_toEndOf="parent"
        app:layout_constraintStart_toStartOf="parent"
        app:layout_constraintTop_toTopOf="parent"
        app:shapeAppearanceOverlay="@style/CircleImageView"
        app:srcCompat="@color/teal_700" />

    <com.google.android.material.imageview.ShapeableImageView
        android:id="@+id/kotlin"
        android:layout_width="50dp"
        android:layout_height="50dp"
        android:background="@color/black"
        app:shapeAppearanceOverlay="@style/CircleImageView"
        app:srcCompat="@color/teal_700"
        app:layout_constraintStart_toStartOf="parent"
        app:layout_constraintEnd_toEndOf="parent"
        app:layout_constraintTop_toBottomOf="@id/xml"
        app:layout_constraintBottom_toBottomOf="parent"/>

</androidx.constraintlayout.widget.ConstraintLayout>
```

`styles.xml`
```xml
<resources>
    <style name="CircleImageView">
    </style>
</resources>
```

**Android API version:** 30

**Material Library version:** 1.3.0

**Device:** Pixel 3 and others

drewhamilton:
#### Don't forget:

Fixes #2063.

Ensures that content padding values are not propagated to their respective `super` padding values until after `onMeasure` has completed with `isLayoutDirectionResolved() == true` at least once. This was already the case inside `onMeasure`, but not the case in `setContentPadding` and `getPadding*` methods.

The internal logic required to make this work is hard to reason about, so I added `ShapeableImageViewTest` to test that both types of padding work as expected on multiple Android SDKs.

### DIFF
--- a/lib/java/com/google/android/material/imageview/ShapeableImageView.java
+++ b/lib/java/com/google/android/material/imageview/ShapeableImageView.java
@@ -37,8 +37,6 @@ import android.graphics.Rect;
 import android.graphics.RectF;
 import android.os.Build.VERSION;
 import android.os.Build.VERSION_CODES;
-import androidx.appcompat.content.res.AppCompatResources;
-import androidx.appcompat.widget.AppCompatImageView;
 import android.util.AttributeSet;
 import android.view.View;
 import android.view.ViewOutlineProvider;
@@ -48,6 +46,8 @@ import androidx.annotation.Dimension;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.RequiresApi;
+import androidx.appcompat.content.res.AppCompatResources;
+import androidx.appcompat.widget.AppCompatImageView;
 import com.google.android.material.resources.MaterialResources;
 import com.google.android.material.shape.MaterialShapeDrawable;
 import com.google.android.material.shape.ShapeAppearanceModel;
@@ -169,28 +169,25 @@ public class ShapeableImageView extends AppCompatImageView implements Shapeable 
       return;
     }
 
-    if (VERSION.SDK_INT > 19 && !isLayoutDirectionResolved()) {
-      return;
+    if ((VERSION.SDK_INT < 19 || isLayoutDirectionResolved())
+        && !hasAdjustedPaddingAfterLayoutDirectionResolved) {
+      // Finally, update the super padding to be the combined `android:padding` and
+      //  `app:contentPadding`, in keeping with ShapeableImageView's crazy internal padding contract:
+      if (VERSION.SDK_INT >= 21 && (isPaddingRelative() || isContentPaddingRelative())) {
+        setPaddingRelative(
+            super.getPaddingStart(),
+            super.getPaddingTop(),
+            super.getPaddingEnd(),
+            super.getPaddingBottom());
+      } else {
+        setPadding(
+            super.getPaddingLeft(),
+            super.getPaddingTop(),
+            super.getPaddingRight(),
+            super.getPaddingBottom());
+      }
+      hasAdjustedPaddingAfterLayoutDirectionResolved = true;
     }
-
-    hasAdjustedPaddingAfterLayoutDirectionResolved = true;
-
-    // Update the super padding to be the combined `android:padding` and
-    // `app:contentPadding`, keeping with ShapeableImageView's internal padding contract:
-    if (VERSION.SDK_INT >= 21 && (isPaddingRelative() || isContentPaddingRelative())) {
-      setPaddingRelative(
-          super.getPaddingStart(),
-          super.getPaddingTop(),
-          super.getPaddingEnd(),
-          super.getPaddingBottom());
-      return;
-    }
-
-    setPadding(
-        super.getPaddingLeft(),
-        super.getPaddingTop(),
-        super.getPaddingRight(),
-        super.getPaddingBottom());
   }
 
   @Override
--- a/lib/java/com/google/android/material/imageview/ShapeableImageView.java
+++ b/lib/java/com/google/android/material/imageview/ShapeableImageView.java
@@ -169,13 +169,12 @@ public class ShapeableImageView extends AppCompatImageView implements Shapeable 
       return;
     }
 
-    if ((VERSION.SDK_INT < 19 || isLayoutDirectionResolved())
-        && !hasAdjustedPaddingAfterLayoutDirectionResolved) {
+    if (VERSION.SDK_INT < 19 || isLayoutDirectionResolved()) {
       hasAdjustedPaddingAfterLayoutDirectionResolved = true;
 
       // Finally, update the super padding to be the combined `android:padding` and
       //  `app:contentPadding`, in keeping with ShapeableImageView's crazy internal padding contract:
-      if (VERSION.SDK_INT >= 21 && (isPaddingRelative() || isContentPaddingRelative())) {
+      if (VERSION.SDK_INT >= 17 && (isPaddingRelative() || isContentPaddingRelative())) {
         setPaddingRelative(
             super.getPaddingStart(),
             super.getPaddingTop(),
@@ -378,6 +377,8 @@ public class ShapeableImageView extends AppCompatImageView implements Shapeable 
           top + getContentPaddingTop(),
           right + getContentPaddingRight(),
           bottom + getContentPaddingBottom());
+    } else {
+      super.setPadding(left, top, right, bottom);
     }
   }
 
@@ -399,6 +400,8 @@ public class ShapeableImageView extends AppCompatImageView implements Shapeable 
           top + getContentPaddingTop(),
           end + getContentPaddingEnd(),
           bottom + getContentPaddingBottom());
+    } else {
+      super.setPaddingRelative(start, top, end, bottom);
     }
   }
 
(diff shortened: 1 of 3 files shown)


lib/java/com/google/android/material/imageview/ShapeableImageView.java
```diff
@@ -169,28 +169,25 @@ protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
...
+            super.getPaddingTop(),
+            super.getPaddingEnd(),
+            super.getPaddingBottom());
+      } else {
+        setPadding(
+            super.getPaddingLeft(),
+            super.getPaddingTop(),
+            super.getPaddingRight(),
+            super.getPaddingBottom());
+      }
```
drewhamilton, line 189: When my previous PR #1871 was merged (outside of GitHub), this logic was refactored incorrectly and caused strange behavior on API 19. I've restored the original logic from #1871 along with the other logic changes/fixes in this PR.

lib/java/com/google/android/material/imageview/ShapeableImageView.java
```diff
@@ -219,13 +216,17 @@ public void setContentPadding(
...
-    // Super padding is equal to background padding + content padding. Adjust the content padding
-    //  portion of the super padding here:
-    super.setPadding(
-        super.getPaddingLeft() - leftContentPadding + left,
-        super.getPaddingTop() - topContentPadding + top,
-        super.getPaddingRight() - rightContentPadding + right,
-        super.getPaddingBottom() - bottomContentPadding + bottom);
+    // If onMeasure hasn't adjusted padding yet, wait for it to be set to avoid double-applying the
+    //  content padding in onMeasure:
+    if (hasAdjustedPaddingAfterLayoutDirectionResolved) {
```
drchen, line 221: Will this affect API >= 19?

lib/java/com/google/android/material/imageview/ShapeableImageView.java
```diff
@@ -244,14 +245,20 @@ public void setContentPadding(
...
-    // Super padding is equal to background padding + content padding. Adjust the content padding
-    //  portion of the super padding here:
-    super.setPaddingRelative(
-        super.getPaddingStart() - getContentPaddingStart() + start,
-        super.getPaddingTop() - topContentPadding + top,
-        super.getPaddingEnd() - getContentPaddingEnd() + end,
-        super.getPaddingBottom() - bottomContentPadding + bottom);
+    // If onMeasure hasn't adjusted padding yet, wait for it to be set to avoid double-applying the
+    //  content padding in onMeasure:
+    if (hasAdjustedPaddingAfterLayoutDirectionResolved) {
```
drchen, line 250: ditto

lib/java/com/google/android/material/imageview/ShapeableImageView.java
drewhamilton: Yes, `hasAdjustedPaddingAfterLayoutDirectionResolved` doesn't get set until after `onMeasure` runs after `isLayoutDirectionResolved` (which is an API 19+ method) becomes true. (So really it affects 19+ more than it affects pre-19.)

lib/java/com/google/android/material/imageview/ShapeableImageView.java
```diff
@@ -169,28 +169,25 @@ protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
...
-    // Update the super padding to be the combined `android:padding` and
-    // `app:contentPadding`, keeping with ShapeableImageView's internal padding contract:
-    if (VERSION.SDK_INT >= 21 && (isPaddingRelative() || isContentPaddingRelative())) {
-      setPaddingRelative(
-          super.getPaddingStart(),
-          super.getPaddingTop(),
-          super.getPaddingEnd(),
-          super.getPaddingBottom());
-      return;
+    if (VERSION.SDK_INT < 19 || isLayoutDirectionResolved()) {
```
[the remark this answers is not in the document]
drewhamilton: Fixed. I was having trouble reasoning about this logic structure last time, but now it's clear to me.

lib/java/com/google/android/material/imageview/ShapeableImageView.java
drchen: Got it. Don't you need to call setContentPadding() again when hasAdjustedPaddingAfterLayoutDirectionResolved is set to true? So the previous set content paddings can be applied?

drewhamilton: No. `setContentPadding` does two things:
1. It calls `super.setPadding` according to this class's internal padding contract. This is skipped if this flag is not set, but `super.setPadding` is also called immediately after the flag is set, so this call is covered either way.
2. It sets the private `contentPadding` variables, regardless of whether that flag is set.

The private `contentPadding` variables are only used in the `getContentPadding` and `getPadding` getters. `updateShapeMask` uses the `getPadding` getters to set the shape size, so now I wonder if I need to force a call to `updateShapeMask` immediately after that flag is set too...

lib/java/com/google/android/material/imageview/ShapeableImageView.java
```diff
@@ -219,13 +219,17 @@ public void setContentPadding(
...
-        super.getPaddingLeft() - leftContentPadding + left,
-        super.getPaddingTop() - topContentPadding + top,
-        super.getPaddingRight() - rightContentPadding + right,
-        super.getPaddingBottom() - bottomContentPadding + bottom);
+    // If onMeasure hasn't adjusted padding yet, wait for it to be set to avoid double-applying the
+    //  content padding in onMeasure:
+    if (hasAdjustedPaddingAfterLayoutDirectionResolved) {
+      // Super padding is equal to background padding + content padding. Adjust the content padding
+      //  portion of the super padding here:
+      super.setPadding(
```
drchen, line 227: It looks like there might be a bug if content padding are set twice, and the first time hasAdjustedPaddingAfterLayoutDirectionResolved is false and the second time it's true. 

At the second call, super's paddings are not actually updated with content paddings yet, so subtracting the content paddings will causes extra paddings being subtracted. Can you also add test cases for this situation after you fix it?

lib/java/com/google/android/material/imageview/ShapeableImageView.java
drchen: That's a good catch. I think onSizeChanged() will be called after setting paddings, right? That should be fine I guess.

lib/java/com/google/android/material/imageview/ShapeableImageView.java
drewhamilton: At your second call to `setContentPadding`, super's paddings **have** been updated, because `setPadding` is called in `onMeasure` immediately after `hasAdjust...` was set, and `setPadding` then called `super.setPadding` with the added differences.

In any case I'll look to add tests that check this.

lib/java/com/google/android/material/imageview/ShapeableImageView.java
drewhamilton: I think `onSizeChanged` is not called in that case actually. I'll need to try to force an error this way to be sure, and add a related test if possible

lib/javatests/com/google/android/material/imageview/ShapeableImageViewTest.java
```diff
@@ -0,0 +1,292 @@
...
+    assertThat(imageView.getContentPaddingBottom()).isEqualTo(8);
+    assertThat(imageView.getPaddingStart()).isEqualTo(1);
+    assertThat(imageView.getPaddingTop()).isEqualTo(2);
+    assertThat(imageView.getPaddingEnd()).isEqualTo(3);
+    assertThat(imageView.getPaddingBottom()).isEqualTo(4);
+
+    forceResolveLayoutDirection(imageView);
+    imageView.onMeasure(WIDTH_MEASURE_SPEC, HEIGHT_MEASURE_SPEC);
+
+    assertThat(imageView.getContentPaddingStart()).isEqualTo(5);
```
drchen, line 264: Can we also add tests to check if super.getPaddingTop/Bottom/Start/End() returns correct values?

PR #2331, opened 2021-08-14
Material button lollipop shape fix

### LINKED ISSUE #1797: [MaterialButton] Ripple doesn't respect shape if app:rippleColor was set on API 21
**Description:** Ripple doesn't respect shape if app:rippleColor was set on API 21. If no attribute was set ripple shape is correct.

**Expected behavior:** Ripple should be clipped to match the button's shape

**Source code:** sample project is attached

**Android API version:** 21

**Material Library version:** 1.2.1

**Device:** Emulator API 21

Video: https://drive.google.com/file/d/1-_Ua63dONtLKFUrkzwUMo4t8eGS6NJ45/view?usp=sharing

Sample project: https://drive.google.com/file/d/1Bzu4H3zX4LQxyVg0Ry9d2zmiI8hogVVr/view?usp=sharing

vicmns:
[Android Lollipop][Buttons] Fallback when updating drawable shapes. (closes #1797)

Changing a button shape in Lollipop doesn't affect the view ripple drawable, which overflows the background drawable:

This change introduces a fallback only for `Lollipop`, in which we recreate the background drawable when the shape changes (As it is currently done when changing the drawable insets):


vicmns:
@googlebot I signed it!

lib/java/com/google/android/material/button/MaterialButtonHelper.java
```diff
@@ -46,11 +45,15 @@
...
 import com.google.android.material.shape.Shapeable;
 
+import static androidx.annotation.RestrictTo.Scope.LIBRARY_GROUP;
+
 /** @hide */
 @RestrictTo(LIBRARY_GROUP)
 class MaterialButtonHelper {
 
-  private static final boolean IS_LOLLIPOP = VERSION.SDK_INT >= VERSION_CODES.LOLLIPOP;
+  private static final boolean IS_MIN_LOLLIPOP = VERSION.SDK_INT >= VERSION_CODES.LOLLIPOP;
```
drchen, line 54: IS_LOLLIPOP_OR_ABOVE?

drchen (approved):

lib/java/com/google/android/material/button/MaterialButtonHelper.java
vicmns: Good point, wasn't "IS_MIN" and old used convention in Android when gating min API logic? I'm pretty sure that I saw some time ago code using this prefix.

drchen: I guess Android uses atLeastXXXX in many places, which also works. : )

PR #2386, opened 2021-09-25
[AppBarLayout] Fix snap issue with transparent status bar

### LINKED ISSUE #2001: [AppBarLayout] Snap behavior bug
**Description:**
There is a bug with AppBarLayout in how it calculates `snapBottom` within `snapToChildIfNeeded`. The issue happens when the `AppBarLayout` has more than one child for example`CollapsingToolbarLayout` and another view below it such as `TabLayout`. After scrolling, the collapsed toolbar expands again automatically. 

The logical error is in the following code in `snapToChildIfNeeded`. 

```java
if (offsetChildIndex == abl.getChildCount() - 1) {
  // If this is the last child, we need to take the top inset into account
  snapBottom += abl.getTopInset();
}
```

The top inset is only used for the last child which in this case would be `TabLayout`. I think instead of the last child, the first child needs to take the top inset into account. So logic should be

```java
if (offsetChildIndex == 0) {
  snapBottom += abl.getTopInset();
}
```

**Expected behavior:** Screenshots and/or description of expected behavior

**Source code:** The code snippet which is causing this issue. Please consider attaching a minimal sample app that reproduces the issue.

**Android API version:** 30

**Material Library version:** 1.2.0

**Device:** Pixel 4a

Khang-NT:
closes #2001

### DIFF
--- a/catalog/java/io/material/catalog/topappbar/res/layout/cat_topappbar_scrolling_transparent_statusbar_activity.xml
+++ b/catalog/java/io/material/catalog/topappbar/res/layout/cat_topappbar_scrolling_transparent_statusbar_activity.xml
@@ -1,5 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!--
+<?xml version="1.0" encoding="utf-8"?><!--
   Copyright 2017 The Android Open Source Project
 
   Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


drchen (approved):
It looks good to me but we may need a thorough test against it. 

Will need to discuss with our TL before we can submit it.

drchen:
Let's use the one without the Catalog change. : )

I'll close this one.

PR #2448, opened 2021-10-29
[MaterialColors] MaterialColors returns the wrong value when color is TypedValued.TYPE_STRING

oOJohn6Oo:
> I came across this problem when implementing Dark Mode.

- Pic 1 is requesting a `android.R.attr.colorBackground`,I override it in my project theme, return `data` is ok.
Pic1

- Pic 2 is requesting a `android.R.attr.textColorPrimary` which I didn't override it, and I found the `TypedValue#Type` is `TYPE_STRING`, we can't just return `data`. 
Pic2

I found this issue, and I solved it, but I don’t know the reason for this design.


oOJohn6Oo:
@googlebot I signed it!

PR #2467, opened 2021-11-10
[Sliders] Make `BaseOnChangeListener` public

ZacSweers:
`Slider.OnChangeListener` and `Slider.OnSliderTouchListener` both extend it and are public APIs. This must be public too, otherwise access to those subtypes are actually a lint error in AGP 7.1+

### DIFF
--- a/lib/java/com/google/android/material/slider/BaseOnChangeListener.java
+++ b/lib/java/com/google/android/material/slider/BaseOnChangeListener.java
@@ -17,15 +17,10 @@
 package com.google.android.material.slider;
 
 import androidx.annotation.NonNull;
-import androidx.annotation.RestrictTo;
-import androidx.annotation.RestrictTo.Scope;
 
 /**
  * Interface definition for a callback invoked when a slider's value is changed.
- *
- * @hide
  */
-@RestrictTo(Scope.LIBRARY_GROUP)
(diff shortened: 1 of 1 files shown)


josefigueroa168:
Slider.OnSliderTouchListener actually extends BaseOnSliderTouchListener, so both of these base classes would need to be made public for the intended effect.

ZacSweers:
Do you want me to add that to this PR too?

josefigueroa168:
Sure, I think it makes sense as the issue is the same.

drchen:
If we explicitly override interface methods in public interfaces, will that remove the lint errors?

ZacSweers:
I don't believe so

dsn5ft:
This was resolved by https://github.com/material-components/material-components-android/commit/3cf0647da0239d65174fc6d651c37fae80048b66.

dsn5ft:
Actually I was just able to test this with AGP 7.1.0-rc01 and Android Studio Bumblebee 2021.1.1 RC1, and I think https://github.com/material-components/material-components-android/commit/3cf0647da0239d65174fc6d651c37fae80048b66 (making the interface methods explicit in each sub-type) does resolve this issue:

# With MDC 1.6.0-alpha01

image

# With MDC 1.6.0-alpha02

image

PR #2609, opened 2022-03-21
[Image View] Fix multiple contentPadding issues

### LINKED ISSUE #2063: [ShapeableImageView] contentPadding not working when set via Kotlin

jpmcosta:
#### Solution

Closes #2063.

Should fix the following `contentPadding` issues:
- `contentPadding` getters returning outdated values when `app:contentPaddingStart` or `app:contentPaddingEnd` are defined in xml;
- `contentPadding` being applied twice, if `setContentPadding` or `setContentPaddingRelative` are called before the view is measured;
  - in this case, the shape mask is also affected to have padding even if only `contentPadding` is defined.

#### Work in progress

### DIFF
--- a/lib/java/com/google/android/material/imageview/ShapeableImageView.java
+++ b/lib/java/com/google/android/material/imageview/ShapeableImageView.java
@@ -216,9 +216,6 @@ public class ShapeableImageView extends AppCompatImageView implements Shapeable 
    */
   public void setContentPadding(
       @Dimension int left, @Dimension int top, @Dimension int right, @Dimension int bottom) {
-    startContentPadding = UNDEFINED_PADDING;
-    endContentPadding = UNDEFINED_PADDING;
-
     // Super padding is equal to background padding + content padding. Adjust the content padding
     //  portion of the super padding here:
     super.setPadding(
@@ -231,6 +228,9 @@ public class ShapeableImageView extends AppCompatImageView implements Shapeable 
     topContentPadding = top;
     rightContentPadding = right;
     bottomContentPadding = bottom;
+
+    startContentPadding = UNDEFINED_PADDING;
+    endContentPadding = UNDEFINED_PADDING;
   }
 
   /**
@@ -256,6 +256,9 @@ public class ShapeableImageView extends AppCompatImageView implements Shapeable 
     topContentPadding = top;
     rightContentPadding = isRtl() ? start : end;
     bottomContentPadding = bottom;
+
+    startContentPadding = start;
+    endContentPadding = end;
   }
 
   private boolean isContentPaddingRelative() {
--- a/lib/java/com/google/android/material/imageview/ShapeableImageView.java
+++ b/lib/java/com/google/android/material/imageview/ShapeableImageView.java
@@ -82,7 +82,6 @@ public class ShapeableImageView extends AppCompatImageView implements Shapeable 
   @Dimension private int bottomContentPadding;
   @Dimension private int startContentPadding;
   @Dimension private int endContentPadding;
-  private boolean hasAdjustedPaddingAfterLayoutDirectionResolved = false;
 
   public ShapeableImageView(Context context) {
     this(context, null, 0);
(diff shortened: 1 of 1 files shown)


lib/java/com/google/android/material/imageview/ShapeableImageView.java
```diff
@@ -140,6 +134,24 @@ public ShapeableImageView(Context context, @Nullable AttributeSet attrs, int def
 
     attributes.recycle();
 
+    if (hasContentPadding()) {
```
drchen, line 137: This seems not right to update paddings before layout direction is resolved? The content padding values thus got might be wrong.

jpmcosta: It seems that, once padding is set in `View`, `resetResolvedPadding()` ensures that `resolvePadding()` will be called when appropriate. I also can't find any documentation that `setPadding()` or `setPaddingRelative()` should only be called when layout direction is resolved.

So, I would say that the previous workflow was unnecessary. However, I might be missing something and I'm sorry if that's the case.

I couldn't reproduce any issues while testing.

drchen: It's not about get/setPadding() but how we get content padding, which relies on View.getLayoutDirection() to decide which content padding should be used. If this call happens before the layout direction is resolved, there might be a problem.

jpmcosta: `View.resolvePadding()` should adjust the padding accordingly, once the layout direction is resolved.

Again, I might be missing something. Do you have an example, even if theoretical, that I can test?

drchen: View.resolvePadding() also makes use of View.getLayoutDirection() to decide if it's RTL, which, again, may not be resolved yet in the constructor.

jpmcosta: This is how I see it:
- `View.resolvePadding()` is usually not called in the constructor;
- In this case, it will be called, because we need to get the current padding values (e.g., calling `super.getPaddingStart()`, etc.);
- However, setting padding also calls `View.resetResolvedPaddingInternal()`, ensuring that `View.resolvePadding()` will be called again later, when `View.getLayoutDirection()` is properly resolved;
- In the meantime, even if `View.getLayoutDirection()` is not accurate, everything will be consistent.

So, I don't see any issues that don't already occur when setting/getting the regular padding.

drchen: So, imagine this scenario - 

If you set paddingStart and contentPaddingRight at the same time. Before RTL is resolved, in the overridden setPaddingRelative(), contentPaddingRight will be added to paddingEnd instead of paddingStart - which is wrong, because it should be added to paddingStart instead.

jpmcosta:
That's definitely an issue. Somehow I wasn't considering cases where you
would mix regular absolute padding with relative content padding, or the
other way around. 👍

lib/java/com/google/android/material/imageview/ShapeableImageView.java
jpmcosta: That's definitely an issue. Somehow I wasn't considering cases where one would mix absolute regular padding with relative content padding, or the other way around. 👍

Issue #2655, opened 2022-04-19
[MaterialTimePicker] Incorrect translation of the confirmation button


pubiqq:
**Description:** The confirmation button in DatePicker has an incorrect translation into some languages (Kazakh, Spanish, Latvian). Here is an example with the Kazakh language (should be "Жарайды"):
<table>
  <tr>
    <th>Material 1.4.0 (correct translation)</td>
    <th>Material 1.5.0 & 1.7.x (incorrect translation)</td>
   </tr> 
   <tr>
      <td><img src="https://user-images.githubusercontent.com/82187521/164058425-ce26ed96-9f23-41e2-ab0e-ab7525e51a29.png" width="300px"></td>
    <td><img src="https://user-images.githubusercontent.com/82187521/164058436-a6e1a55d-c5d9-41f6-922a-3b56f9eef2ce.png" width="300px"></td>
  </td>
  </tr>
</table>

**Material Library version:** 
- 1.4.0 works correctly
- 1.5.0, 1.6.0-beta01, 1.7.0-alpha01 work incorrectly

Issue #2577, opened 2022-02-23
[MaterialDatePicker] [MaterialTimePicker] Buttons are not displayed properly


mal7othify:
**Description:** Full description of issue here

**Expected behavior:** The MaterialDatePicker and MaterialTimePicker buttons (OK and Cancel) are not displayed properly in One Plus device. Refer to the screenshot below:
<img width="419" alt="Screen Shot 2022-02-23 at 7 42 47 AM" src="https://user-images.githubusercontent.com/35968097/155262983-46cefd5f-7b4c-46cf-85b4-58f5fe9479ba.png">

**Android API version:** 30 & 31

**Material Library version:** com.google.android.material:material:1.3.0 & com.google.android.material:material:1.5.0

**Device:** Device details:
Device Name: OnePlus 9 Pro 5G (edited)
Model: LE2125
OS Version: Android 12

To help us triage faster, please check to make sure you are using the [latest version](https://github.com/material-components/material-components-android/releases) of the library.

We also happily accept [pull requests](https://github.com/material-components/material-components-android/pulls).

drchen:
Can you provide the themes and the styles you are using? A reproducible sample will be even better.

drchen:
I'll close this issue for now due to no further info available. Please feel free to reopen it with more info.

Issue #1602, opened 2020-08-10
MaterialDatePicker OK button is not visible


huize:
This BUG appears only on the specific devices, the "OK" button is not visible.

#### Current behavior:
 Huawei P30
![61597040133](https://user-images.githubusercontent.com/2106901/89756146-46179180-db14-11ea-9039-0403b6642b8e.jpg)

#### Expected behavior:
Pixel 3:
![71597040166](https://user-images.githubusercontent.com/2106901/89756152-4ca60900-db14-11ea-8c74-946c21325dcb.jpg)

#### Source code:
```
 val datePicker =
                MaterialDatePicker.Builder.datePicker().setCalendarConstraints(
                    CalendarConstraints.Builder().setEnd(Date().time).build()
                ).build()
       
   datePicker.addOnPositiveButtonClickListener { newDateInMillis ->
              // do something
            }
   datePicker.showNow(supportFragmentManager!!, "datePicker")
```

matpag:
Can you try the latest version `1.3.0-alpha02` to see if the problem is still there?
Thank you

wcshi:
Duplicate of https://github.com/material-components/material-components-android/issues/1346

huize:
> Duplicate of #1346
Not the same bug，I update the lib to 1.3.0-alpha02 。 this bug  bug still exists。

huize:
> Can you try the latest version `1.3.0-alpha02` to see if the problem is still there?
> Thank you

Not the same bug，I update the lib to 1.3.0-alpha02 。 this bug bug still exists。

wcshi:
According to https://www.gsmarena.com/huawei_p30-9530.php, Huawei p30 should have the following resolution:
1080 x 2340 pixels, 19.5:9 ratio (~422 ppi density)

@huize what font size are you using in your custom style? A screenshot of your layout inspector view would be helpful.

huize:
> According to https://www.gsmarena.com/huawei_p30-9530.php, Huawei p30 should have the following resolution:
> 1080 x 2340 pixels, 19.5:9 ratio (~422 ppi density)
> 
> @huize what font size are you using in your custom style? A screenshot of your layout inspector view would be helpful.

@wcshi  This bug occurred on a device of a user from Google play,i just set the activity style. cutive font downloaded from google fonts
```
<style name="AppTheme" parent="Theme.MaterialComponents.Light">
        <!-- Customize your theme here. -->
[...]

Issue #1346, opened 2020-05-29
MaterialDatePicker OK button is not fully visible (3 dots)


ghost:
This BUG appears only on the specific devices, the "OK" button is not fully visible, and it displays as a "3 dots".

**Current behavior:**
Nokia 7 Plus
![85e24d84-ab07-44ea-a83f-576edfa8f8a9](https://user-images.githubusercontent.com/46451579/83291214-45937e00-a1e8-11ea-8af9-9117f1e4fa76.png)

**Expected behavior:** 
Pixel 3:
![IMAGE 2020-05-29 20:04:36](https://user-images.githubusercontent.com/46451579/83290881-b1c1b200-a1e7-11ea-930b-c13e7c389ff8.jpg)

**Source code:** 
Extension function:
```
fun AppCompatActivity.datePicker(
    titleId: Int? = null,
    minusYears: Int = 0,
    minusDay: Int = 0,
    startDate: Long? = null,
    datePattern: String = DMY,
    dateChosen: (String, Long) -> Unit
) {
    val endDate = LocalDate().minusYears(minusYears).minusDays(minusDay)
    MaterialDatePicker.Builder.datePicker().apply {
        titleId?.let { setTitleText(it) }
        setCalendarConstraints(CalendarConstraints.Builder().apply {
            startDate?.let { setStart(it) }
            setEnd(endDate.toDate().time)
        }.build())
    }.build().run {
        addOnPositiveButtonClickListener {
            dateChosen.invoke(DateTime(it).toLocalDate().toString(datePattern), it)
        }
        show(supportFragmentManager, toString())
    }
}
```
On the place of call:
```
datePicker(
                R.string.some_str,
                PAST_18_YEARS,
                PAST_1_DAY
            ) { date, timestamp ->
                espUser.birthDate = timestamp.mlsToSeconds()
            }
```
**Android API version:** - 29

**Material Library version:** 1.2.0-alpha06

**Device:** : Nokia 7 Plus

wcshi:
Based on [these specs](https://www.gsmarena.com/nokia_7_plus-9058.php), both the Nokia 7 Plus and the Pixel 3 have the same screen resolution: 1080 x 2160 pixels, 18:9 ratio (~403 ppi density)

Are you using a non-default screen display or font size setting?

PavelPetkevich:
Hello @wcshi , 

Thank you for reacting to the issue.

Nop, both devices have all settings by defaults...

[...]

PR #2656, opened 2022-04-19
Replace all Android strings with library ones

pubiqq:
[Context](https://github.com/material-components/material-components-android/issues/1346#issuecomment-933791071):
> It did look like the Android framework string is missing on certain OEM devices. (Material is using android:string/ok for the button.) I guess we can fix it by replacing the string with proper values.
---
- Fixes https://github.com/material-components/material-components-android/issues/2655
- Fixes https://github.com/material-components/material-components-android/issues/2577
- Fixes https://github.com/material-components/material-components-android/issues/1602
- Fixes https://github.com/material-components/material-components-android/issues/1346

### DIFF
--- a/lib/java/com/google/android/material/datepicker/res/values-af/strings.xml
+++ b/lib/java/com/google/android/material/datepicker/res/values-af/strings.xml
@@ -23,7 +23,8 @@
   <string name="mtrl_picker_range_header_selected">%1$s-%2$s</string>
   <string name="mtrl_picker_date_header_unselected">Gekose datum</string>
   <string name="mtrl_picker_date_header_selected">%1$s</string>
-  <string name="mtrl_picker_confirm">@android:string/ok</string>
+  <string name="mtrl_picker_confirm">"OK"</string>
+  <string name="mtrl_picker_cancel">"Kanselleer"</string>
   <string name="mtrl_picker_text_input_date_hint">Datum</string>
   <string name="mtrl_picker_text_input_date_range_start_hint">Begindatum</string>
   <string name="mtrl_picker_text_input_date_range_end_hint">Einddatum</string>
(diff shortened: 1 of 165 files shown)


pubiqq:
OK, I will update my PR as soon as all the translations appear.

pubiqq:
By the way, I noticed you partially added translations [here](https://github.com/material-components/material-components-android/commit/431e291b01a4862146bba73381cda749bbfe4363) and some of them differ from `android:string/ok`:

| Resource | @android:string/ok | [`431e291`](https://github.com/material-components/material-components-android/commit/431e291b01a4862146bba73381cda749bbfe4363) |
|-|-|-|
| values-as | `ঠিক আছে` | `OK` |
| values-hy | `Լավ` | `Եղավ` |
| values-ka | `OK` | `კარგი` |
| values-mr | `ठीक आहे` | `ओके` |
| values-or | `ଠିକ୍‍ ଅଛି` | `OK` |

In these cases, the translations will differ both from the old `MaterialDatePicker` implementation and from `android.app.DatePickerDialog`. It's not critical, nevertheless, I would like to know if you will keep the current translations or will try to unify them with standard Android strings?

P. S. As for the correctness of the translations, I can't say anything, specifically for my PR I took them from Android SDK 31 sources.

dsn5ft:
No need to update this PR, we will automatically add the translations to the library when they are available internally.

We use a different translation service, so the specific translations will not always match whatever is in the AOSP source - we will not attempt to maintain consistency with that.

pubiqq:
> No need to update this PR, we will automatically add the translations to the library when they are available internally.

I meant "update" = "leave the relevant comment and close" 😄

> ... the specific translations will not always match whatever is in the AOSP source - we will not attempt to maintain consistency with that.

OK, thanks for the answer!

dsn5ft:
Ah! Gotcha. I will actually link your PR to my commit coming here soon and the PR will be auto-closed. The translations will be synced out automatically within a week or two I believe.

pubiqq:
Oh, that's even better, thanks!

Issue #2028, opened 2021-02-03
[Circular progress indicator] Does not work on API 22 only


Massi-X:
**Description:** Inflate and put an indeterminate circular spinning indicator programmatically in a layout and try to start the app. For the record I've set it as the actionview for a menu, but I bet it's the same for every other case.

**Expected behavior:** The progress indicator should be visible and spinning. This is not the case on API 22, where the progress indicator doesn't spin at all. From my tests it works on API 19,21,23,24,29 and doesn't work on API 22 only.

**Source code:**
loading.xml:
```
<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
    xmlns:app="http://schemas.android.com/apk/res-auto"
    android:layout_width="wrap_content"
    android:layout_height="wrap_content">

<com.google.android.material.progressindicator.CircularProgressIndicator
        android:layout_width="20dp"
        android:layout_height="20dp"
        android:layout_margin="10dp"
        android:indeterminate="true"
        app:indicatorColor="@color/zero_one_state"
        app:indicatorSize="20dp"
        app:trackThickness="3dp" />
</LinearLayout>
```
Main class:
```
@Override
public boolean onCreateOptionsMenu(Menu menu) { 
    menu.findItem(R.id.item).setActionView(R.layout.loading);
}
```

**Android API version:** 22

**Material Library version:** 1.3.0-rc01 - 1.3.0 stable - 1.4.0-alpha02 - 1.4.0 stable - ... - 1.6.0 alpha01 - 1.7.0-alpha01

**Device:** Emulator x86 or x86_64

Massi-X:
Updated to reflect the new release

kroegerama:
Same error on 1.4.0-alpha02. I guess @pekingme is the one who did the most work on this component. Are there any plans to fix this?

khukuh123:
I have the same issue on API 22 and I'm using version 1.4.0

sebkur:
I can reproduce this with 1.5.0 stable and 1.7.0-alpha01

PR #2675, opened 2022-04-29
[ProgressIndicator] Fix indeterminate progress indicator not working on API 22

sebkur:
Resolves https://github.com/material-components/material-components-android/issues/2028

#### Don't forget:


sebkur:
I was able to reproduce #228 on the catalog app. It seems the check to start the animator needs to be done for API 22, to not just API 21 and below. I also checked API 23 and a few more recent versions, which seems to be fine. This has already changed once in the past (https://github.com/material-components/material-components-android/commit/86b70d7bf773) but it seems the API version used in that check needs to be incremented once more to make this work consistently on all API levels.

sebkur:
So this change fixes the bug that can be observed in the catalog app. However, while the catalog app works fine for API level 23, in the use case I'm coming from, the indicator has issues on API level 22 and 23. So it's possible this really needs to be changed to `VERSION_CODES.M`

PR #2679, opened 2022-05-02
[Catalog] Fix DatePicker when the "Last 2 weeks" option is selected

pubiqq:


dsn5ft:
@pubiqq thanks for the fix!

PR #2685, opened 2022-05-04
[i18n][Catalog] Use regular strings instead of @android:string/* which can have translation issues

pubiqq:
Follow-up of https://github.com/material-components/material-components-android/pull/2656


NTPaa (approved):
vv

Issue #2687, opened 2022-05-05
[MaterialDatePicker] The keyboard doesn't close when switching to calendar mode


pubiqq:
https://user-images.githubusercontent.com/82187521/166897517-713726db-b454-46c6-ae3e-541a3d0aa3fd.mp4

**Material Library version:** 1.6.0

pubiqq:
@leticiarossi, it may be worth changing the `Issue: Catalog` label to `Widget: DatePicker`. I don't think this issue was specific only to the `:catalog` app.

PR #2688, opened 2022-05-05
[MaterialDatePicker] Hide the keyboard when switching to calendar mode

pubiqq:
Resolves https://github.com/material-components/material-components-android/issues/2687


leticiarossi:
thanks for the PR! it looks good, I'll just move the `setUpKeyboardBehavior` method to the base class so both subclasses can use it, but I can do it when I import your code!

Issue #1700, opened 2020-09-10
[Catalog App: Bottom App Bar] BottomSheet is Too Light


manabu-nakamura:
**Description:** https://github.com/material-components/material-components-android/tree/master/catalog/java/io/material/catalog/bottomappbar
The BottomSheet (#3b3b3b) is too light (the elevation of BottomSheet is 32dp totally).

![Screenshot_20210908-145455](https://user-images.githubusercontent.com/71050561/132454477-7cd5c134-ce41-4dd2-a847-239b020092cf.png)

**Expected behavior:** The BottomSheet should be #2d2d2d (the elevation of BottomSheet should be 16dp).
https://material.io/design/environment/elevation.html
https://material.io/design/color/dark-theme.html

**Source code:** https://github.com/material-components/material-components-android/blob/master/catalog/java/io/material/catalog/bottomappbar/res/layout/cat_bottomappbar_fragment.xml#L49
(original)
```
  <FrameLayout
      android:id="@+id/bottom_drawer"
      android:layout_width="match_parent"
      android:layout_height="wrap_content"
      android:elevation="16dp"                                                                <=
      app:behavior_hideable="true"
      app:layout_behavior="@string/bottom_sheet_behavior">

    <com.google.android.material.navigation.NavigationView
        android:id="@+id/navigation_view"
        android:layout_width="match_parent"
        android:layout_height="wrap_content"
        app:menu="@menu/demo_primary"/>
  </FrameLayout>
```
(revised)
```
  <FrameLayout
      android:id="@+id/bottom_drawer"
      android:layout_width="match_parent"
      android:layout_height="wrap_content"
      app:behavior_hideable="true"
      app:layout_behavior="@string/bottom_sheet_behavior">

    <com.google.android.material.navigation.NavigationView
        android:id="@+id/navigation_view"
        android:layout_width="match_parent"
        android:layout_height="wrap_content"
        app:menu="@menu/demo_primary"/>
  </FrameLayout>
```

**Android API version:** 23

**Material Library version:** 1.4.0

**Device:** Nexus 7 (2013)

drchen:
Since it's related to the old Material themes/styles I'll close it for now. Feel free to reopen it if it's still an issue with Material 3.

manabu-nakamura:
I don't think that the elevation of the bottom sheet is 16dp.
https://github.com/material-components/material-components-android/blob/master/docs/components/BottomSheet.md

drchen:
[...]

PR #2692, opened 2022-05-07
[Catalog App: Bottom App Bar] BottomSheet is Too Light

manabu-nakamura:
#1700


catalog/java/io/material/catalog/bottomappbar/res/layout/cat_bottomappbar_fragment.xml
```diff
@@ -42,19 +42,12 @@
       app:layout_anchor="@id/bar"
       app:srcCompat="@drawable/ic_add_24px"/>
 
-  <FrameLayout
-      android:id="@+id/bottom_drawer"
-      android:layout_width="match_parent"
-      android:layout_height="wrap_content"
-      android:elevation="16dp"
```
leticiarossi, line 49: I think instead of deleting the frame layout you can just delete the elevation here and set `style="@style/Widget.Material3.BottomSheet"`, similar to how the bottom sheet demo does. could you try that?

manabu-nakamura: > I think instead of deleting the frame layout you can just delete the elevation here and set `style="@style/Widget.Material3.BottomSheet"`, similar to how the bottom sheet demo does. could you try that?

Thank you for your comment. I tried (https://github.com/material-components/material-components-android/pull/2692/commits/d45cbfe497f3369eaf3603b7b973abd581887fae).

What is the role of the FrameLayout?
(without FrameLayout)
When the state of the bottom sheet is STATE_HALF_EXPANDED, the corners of the bottom sheet are rounded.
Screenshot_20220510_122914
(with FrameLayout)
When the state of the bottom sheet is STATE_HALF_EXPANDED, the corners of the bottom sheet are not rounded.
Screenshot_20220510_123029

leticiarossi (approved):

catalog/java/io/material/catalog/bottomappbar/res/layout/cat_bottomappbar_fragment.xml
manabu-nakamura: Thank you very much for your explanation. I understand. I was surprised when I first saw the bottom sheet without the FrameLayout at https://material.io/components/navigation-drawer/android#bottom-navigation-drawer (https://m3.material.io/components/navigation-drawer/implementation/android#bottom-navigation-drawer).

Issue #2716, opened 2022-05-24
[Catalog] Card selection doesn't work in `CardSelectionModeActivity`


pubiqq:
https://user-images.githubusercontent.com/82187521/170089842-c8a9fd10-7a77-41d4-bea3-75f26e1afc30.mp4

**Material Library version:** 1.7.0-alpha03
**Device:** Pixel 4a 5G

The problem is also present when interacting using TalkBack.

PR #2717, opened 2022-05-24
[Catalog] Fix CardSelectionModeActivity sample

pubiqq:
Resolves https://github.com/material-components/material-components-android/issues/2716

### DIFF
--- a/catalog/java/io/material/catalog/card/SelectableCardsAdapter.java
+++ b/catalog/java/io/material/catalog/card/SelectableCardsAdapter.java
@@ -31,7 +31,6 @@ import androidx.recyclerview.selection.ItemKeyProvider;
 import androidx.recyclerview.selection.SelectionTracker;
 import com.google.android.material.card.MaterialCardView;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 
 /** An Adapter that works with a collection of selectable card items */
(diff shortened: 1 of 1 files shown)


catalog/java/io/material/catalog/card/SelectableCardsAdapter.java
```diff
@@ -99,13 +99,6 @@ private void bind(Item item, int position) {
       if (selectionTracker != null) {
         bindSelectedState();
       }
-      // Set an OnLongClickListener for accessibility
-      materialCardView.setOnLongClickListener(
-          v -> {
-            selectionTracker.setItemsSelected(
```
drchen, line 105: I think the right fix should be adding an if check instead:

if (materialCardView.isChecked()) {
  selectionTracker.xxxxxx....
  return true;
}
return false;

pubiqq: `SelectionTracker` manages the list item selections on its own, it shouldn't need additional `onClickListener`.

drchen:
The original commit that added this solution/workaround: https://github.com/material-components/material-components-android/commit/a49fa6ab36276a762d46d9f716b4ef8ff5947907

If I understand correctly, `SelectionTracker` is using `RecyclerView.OnItemTouchListener` to grab click/long click events. Therefore if the accessibility framework directly send click/long click events, `SelectionTracker` won't be able to catch them, and that's why we need the workaround at the beginning and we need to keep the behavior.

drchen:
Paul, can you take a look?

pubiqq:
boop

pubiqq:
any news?

pubiqq:
any news?

paulfthomas:
This just reverts the original change that was made to make (attempt...) the card selection accessible, would be great to find a proper fix instead of just removing it.

pubiqq:
A proper fix can only be provided by `androidx.recyclerview:recyclerview-selection` developers. 
The original solution is correct from a logical point of view, but doesn't work only with Talkback due to some problem with `recyclerview-selection`. The current solution is a hack and doesn't work correctly at all (neither in normal mode nor in Talkback). That's why I think the current change needs to be reverted.

pubiqq:
bump

pubiqq:
bump

pubiqq:
bump

pubiqq:
bump

PR #2759, opened 2022-06-14
[MaterialSwitch] Move public.xml to the correct directory

AshutoshSundresh:
## public.xml must be included within res-public/values and not res-public

This fixes the following error when compiling the Material components from source in the AOSP build system:
>> materialswitch/res-public/public.xml: error: invalid configuration 'public'.


drchen (approved):
Oh oops. Thanks for helping fix this!

PR #2780, opened 2022-06-26
[MaterialDividerItemDecoration] Fix decorator when lastItemDecorated false and last item contains a view gone and changes the state using notifyItemChanged

### LINKED ISSUE #2771: [MaterialDividerItemDecoration] The decorator in the last item with isLastItemDecorated false is blinking in a dropdown view showing and hiding the decorator very fast
https://github.com/material-components/material-components-android/blob/3fdfb6e230767138849edf858c13da1bf1d3f90b/lib/java/com/google/android/material/divider/MaterialDividerItemDecoration.java#L321

I used this divider in the last item with a dropdown.  When the itemView is clicked then shows a textView from gone to visible state and it is working ok (the divider is not showed) but when the view collapses the textView (from visible to gone) then the childcount is increased one time and decreased very quick, showing and hiding the divider very fast.

Maybe using the itemCount from adapter and not the childCount could fix the problem.  I tried this approach using a custom ItemDecorator overriding the onDrawOver and it is working, but i dont know if is the best solution.

Update: The blink appears when is called the method notifyItemChanged(pos, item)

mnlmato:
Fix #2771

### DIFF
--- a/lib/java/com/google/android/material/divider/MaterialDividerItemDecoration.java
+++ b/lib/java/com/google/android/material/divider/MaterialDividerItemDecoration.java
@@ -24,12 +24,15 @@ import android.graphics.Canvas;
 import android.graphics.Rect;
 import android.graphics.drawable.Drawable;
 import android.graphics.drawable.ShapeDrawable;
+
 import androidx.recyclerview.widget.LinearLayoutManager;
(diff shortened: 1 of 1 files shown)


mnlmato:
Problems with some commits during the fork.

PR #2781, opened 2022-06-26
[MaterialDividerItemDecoration] When lastItemDecorated false then creates a blink with the last view contains a dropdown from visible to gone

mnlmato:
solves #2771


mnlmato:
I need to fix the local email to pass the CLA

Issue #2703, opened 2022-05-13
[ShapeableImageView] Corners color bug with MaterialContainerTransform


dummyco:
**Description:**

When using `ShapeableImageView` together with `MaterialContainerTransform`, `sharedElementReturnTransition` results in black `ShapeableImageView` corners coloring.

**Expected behavior:**

`ShapeableImageView` transition corners coloring should not be impacted.

**Source code:**

```
        binding.shapeableImageView.shapeAppearanceModel = binding.shapeableImageView.shapeAppearanceModel.toBuilder()
            .setAllCornerSizes(
                resources.getDimensionPixelOffset(R.dimen.corner_radius)
                    .toFloat()
            )
            .build()
```

```
        sharedElementEnterTransition = MaterialContainerTransform().also {
            it.drawingViewId = R.id.nav_host_fragment_content_main
            it.duration = 2000
        }

        sharedElementReturnTransition = MaterialContainerTransform().also {
            it.drawingViewId = R.id.nav_host_fragment_content_main
            it.duration = 2000
        }
```

```
        binding.cardView.setOnClickListener {
            val extras =
                FragmentNavigatorExtras(binding.cardView to "test_transition")

            findNavController().navigate(
                R.id.action_FirstFragment_to_SecondFragment,
                null,
                null,
                extras
            )
        }
```

**Recording:**

https://user-images.githubusercontent.com/3989453/168390598-8e248031-2d9a-41fd-a669-6ec458ba1e09.mp4

**Minimal sample app repro:**
https://github.com/dummyco/shapeable-image-view-transition-corners-issue

**Android API version:** 32

**Material Library version:** `1.7.0-alpha01`, `1.5.0`, `1.6.0`

**Device:** `Pixel 6 Pro`, `OnePlus 7T Pro`, `Pixel 4 API 31 Emulator`

Issue #1707, opened 2020-09-11
[Motion] Unloaded ImageView when returning from MaterialContainerTransform


thibseisel:
**Description:** 
I'm using `MaterialContainerTransform` to transition from _Fragment A_, to a _Fragment B_ that contains a `RecyclerView` whose items have `ShapeableImageView`s. Note: _Fragment A_ is part of a `ViewPager2`, hosted by _Fragment Z_.

Transitionning from A to B works fine: transition is postponed until B's `RecyclerView` display is ready.
But when returning from B, we notice that images (that are loaded with Glide) are reset and blank.

![material_container_bug](https://user-images.githubusercontent.com/18646439/92931461-c6971e00-f443-11ea-8495-b0498723c382.gif)

**Expected behavior:**
Images loaded into `ImageView`s should not be reset ; those should be preserved in the return transition.
Also, why is `SheapeableImageView` showing black rounded corners ? This strange behavior of `ShapeableImageView` is also noticeable on Android Studio's layout preview (corners are then light grey).

**Source code:** 

In Fragment Z:
```kotlin
override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
  super.onViewCreated(view, savedInstanceState)
  postponeEnterTransition(1000, TimeUnit.MILLISECONDS)
  // Configure ViewPager adapter...
}
```

In Fragment A:
```kotlin
override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
  super.onViewCreated(view, savedInstanceState)
  // Some view initialization code [...]
  viewModel.liveData.observe(viewLifecycleOwner) {
    adapter.submitList(it)
    val fragmentZ = requireParentFragment()
    fragmentZ.requireView().parent?.doOnPreDraw {
      fragmentZ.startPostponedEnterTransition()
    }
  }
}

private fun displayFragmentB() {
  // Root item view from the clicked element in RecyclerView. Has transitionName set in onBindViewHolder
  val itemView = ...
  val sharedElementExtras = FragmentNavigatortExtras(itemView to "some_item_id")
  requireParentFragment().apply {
    exitTransition = Hold().apply {
      duration = 300
      addTarget(R.id.fragment_z)
    }
    // Reset transition set by other navigation events
    reenterTransition = null
  }

  findNavController().navigate(R.id.to_fragment_b, sharedElementExtras)
}
```

In Fragment B:
```kotlin
override fun onCreate(savedInstanceState: Bundle?) {
[...]

PR #2792, opened 2022-06-28
[ShapeableImageView] Don't change layer type when view is detached from the window

pubiqq:
- _Perhaps_ partially resolves https://github.com/material-components/material-components-android/issues/2703
- Partially resolves the issue with corners in https://github.com/material-components/material-components-android/issues/1707

### DIFF
--- a/lib/java/com/google/android/material/imageview/ShapeableImageView.java
+++ b/lib/java/com/google/android/material/imageview/ShapeableImageView.java
@@ -108,6 +108,8 @@ public class ShapeableImageView extends AppCompatImageView implements Shapeable 
         context.obtainStyledAttributes(
             attrs, R.styleable.ShapeableImageView, defStyle, DEF_STYLE_RES);
 
+    setLayerType(LAYER_TYPE_HARDWARE, null);
+
     strokeColor =
         MaterialResources.getColorStateList(
             context, attributes, R.styleable.ShapeableImageView_strokeColor);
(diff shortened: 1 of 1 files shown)


drchen:
Can you explain why this may (partially) fixes the issues?

pubiqq:
When the exit transition starts, the layer type is set to `LAYER_TYPE_NONE` (inside `onDetachedFromWindow`). In this case, [`clearPaint`](https://github.com/material-components/material-components-android/blob/7939c171dffbe2687d3e12bf00a973d79a839d27/lib/java/com/google/android/material/imageview/ShapeableImageView.java#L69) (`Paint` with `Xfermode = PorterDuff.Mode.DST_OUT`) stops working correctly. 
The patch just deletes the layer type setup from `onDetachedFromWindow`, leaving the layer type set to `LAYER_TYPE_HARDWARE`.

About _perhaps_ and _partially_:
-  I wrote that PR **_perhaps_** fixes #2703 because I didn't check it on [repro app](https://github.com/dummyco/shapeable-image-view-transition-corners-issue). However, I did it on my internal project, and the fix worked there.
- I wrote that PR only **_partially_** fix the issues because it seems that using `LAYER_TYPE_HARDWARE` or `LAYER_TYPE_SOFTWARE` is a _necessary_, but _not sufficient_ condition to work properly. Unfortunately, I don't have an _acceptable_ solution that would completely solve the issue with corners during the transitions yet, most likely I will create a separate issue about it a little later.

drchen (approved):

drchen:
Awesome, thanks! I'm not familiar with how this is working. I'll send it for an internal review and see what they will say. : )

PR #2801, opened 2022-07-04
feat: Update dependencies

MartLintz:


drchen:
Can you sign the CLA and elaborate the reason behind the change?

FYI we will be extra careful regarding updating dependencies because this will basically affect all of our clients.
In general we will only update our dependencies on eligible demands.

PR #2804, opened 2022-07-06
[MaterialDividerItemDecoration] Fix divider count

pubiqq:
Reverts 03c4353937c18ccf13855fc27dc4e6d36ef0eecf

[At the moment](https://github.com/material-components/material-components-android/tree/4e52469ec43e18aaeac6ae20efd4c4ceab66ebaf), `dividerCount` is based on the number of items _in the adapter_, although the enumeration is based on the items that are _directly inside `RecyclerView`_, and this leads to incorrect behavior.

In particular, `setLastItemDecorated(false)` no longer works if there are many items in `RecyclerView`.

The PR rolls back the commit that brought the above issues.

### DIFF
--- a/lib/java/com/google/android/material/divider/MaterialDividerItemDecoration.java
+++ b/lib/java/com/google/android/material/divider/MaterialDividerItemDecoration.java
@@ -317,18 +317,16 @@ public class MaterialDividerItemDecoration extends ItemDecoration {
     left += isRtl ? insetEnd : insetStart;
     right -= isRtl ? insetStart : insetEnd;
 
-    int itemCount = parent.getAdapter() == null ? 0 : parent.getAdapter().getItemCount();
-    int dividerCount = lastItemDecorated ? itemCount : itemCount - 1;
+    int childCount = parent.getChildCount();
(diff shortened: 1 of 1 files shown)


pubiqq:
After 03c4353937c18ccf13855fc27dc4e6d36ef0eecf, `setLastItemDecorated(false)` has no effect if there are many items in `RecyclerView`.

pubiqq:
The reason for revert is given in the PR description.
As for your case, I can't say anything yet. Could you attach a sample app that reproduces your issue?

mnlmato:
I need to create the sample app from scratch, but it is only an adapter with an update method passing by param the position and the new item to insert, using the notifyItemChange.  And the item has the atrrbiutes text as string and the isSelected boolean.  If false, the the extra view is gone else is visible.

drchen:
Hi @pubiqq thanks for the finding and the PR!

Let me create a rollback internally instead for better tracking. : )

drchen (approved):

drchen:
Close the issue as the original PR is rolled back.

Issue #2846, opened 2022-07-21
[Badging] Crash after Process death when horizontalOffset is set but verticalOffset is not


Sh4dowSoul:
**Description:** BadgeState.java performs a wrong null check which can cause crashes when opening an app after a process death. Currently storedState.horizontalOffsetWithoutText is checked, but storedState.verticalOffsetWithoutText is accessed. This results in a crash if horizontalOffsetWithoutText is set but verticalOffsetWithoutText is not.

**Expected behavior:** verticalOffsetWithoutText is checked for a null value before accessing it.

**Source code:** [BadgeState.java](https://github.com/material-components/material-components-android/blob/07938f572fa476364452c283ef7c9ee2037d9ccc/lib/java/com/google/android/material/badge/BadgeState.java#L165)
```java
currentState.verticalOffsetWithoutText =
        storedState.horizontalOffsetWithoutText == null
            ? a.getDimensionPixelOffset(R.styleable.Badge_verticalOffset, 0)
            : storedState.verticalOffsetWithoutText;
 ```

**Stacktrace**

> fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.Integer.intValue()' on a null object reference
>        at com.google.android.material.badge.BadgeState.<init>(BadgeState.java:166)
>        at com.google.android.material.badge.BadgeDrawable.<init>(BadgeDrawable.java:260)
>        at com.google.android.material.badge.BadgeDrawable.createFromSavedState(BadgeDrawable.java:185)
>        at com.google.android.material.badge.BadgeUtils.createBadgeDrawablesFromSavedStates(BadgeUtils.java:297)
>        at com.google.android.material.navigation.NavigationBarPresenter.onRestoreInstanceState(NavigationBarPresenter.java:128)
>        at androidx.appcompat.view.menu.MenuBuilder.dispatchRestoreInstanceState(MenuBuilder.java:357)
>        at androidx.appcompat.view.menu.MenuBuilder.restorePresenterStates(MenuBuilder.java:369)
>        at com.google.android.material.navigation.NavigationBarView.onRestoreInstanceState(NavigationBarView.java:932)
> 

**Android API version:** Does not depend on Android API Version

**Material Library version:** Tested on 1.6.1, but issue persists in master.

**Device:** Is not device dependent

PR #2848, opened 2022-07-21
[Badging] Fix NPE in BadgeState

Sh4dowSoul:
Fix wrong null check in BadgeState.java which can result in crashes after process death.

closes #2846 

- [x] Identify the component the PR relates to in brackets in the title.
- [x] Link to GitHub issues it solves.
- [x] Sign the CLA bot. You can do this once the pull request is opened.

PR #2849, opened 2022-07-21
[Badge] Fix verticalOffsetWithoutText null check in BadgeState

pubiqq:
Fixes https://github.com/material-components/material-components-android/issues/2846


drchen:
Hi thanks for providing the fix. Someone already created #2848 though. Let's merge that one instead. : )

pubiqq:
👌

PS Nice timing :)
Untitled1

PR #2872, opened 2022-07-28
[BottomSheetBehavior] Add method to calculate slide offset

### LINKED ISSUE #2871: [BottomSheetBehavior] Expose slide offset calculation
I have several runtime transitions dependent on the offset of my bottom sheet. However, these transitions need to be initialized on startup, something that can't be provided by `BottomSheetCallback` as it does not fire on startup. Since making that callback fire on startup would break compatibility, instead I want a way to get the slide offset so I can initialize the components myself.

A method (`calculateSlideOffset`) that returns the current slide offset of the bottom sheet. If the sheet is not laid out, it can just be 0.

**Describe alternatives you've considered**
The only other option is either reflecting for the private attributes required to find the slide offset and calculating it myself (Something I dislike), or making the callback fire on startup (Which again, would not be backwards compatible)

OxygenCobalt:
Add a `calculateSlideOffset` method to `BottomSheetBehavior`, allowing the current slide offset to be obtained at runtime.

I have several runtime transitions dependent on the offset of my bottom sheet. However, these transitions need to be initialized on startup, something that can't be provided by BottomSheetCallback as it does not fire on startup. Since making that callback fire on startup would break compatibility, instead I want a way to get the slide offset so I can initialize the components myself.

This PR resolves that by adding a method (`calculateSlideOffset`) that returns the current slide offset of the bottom sheet. If the sheet is not laid out, it will return 0.

Resolves #2871.

### DIFF
--- a/lib/java/com/google/android/material/bottomsheet/BottomSheetBehavior.java
+++ b/lib/java/com/google/android/material/bottomsheet/BottomSheetBehavior.java
@@ -1075,6 +1075,10 @@ public class BottomSheetBehavior<V extends View> extends CoordinatorLayout.Behav
    * value corresponds to an expanded state.
    */
   public float calculateSlideOffset() {
+    if (viewRef == null) {
+      return Float.MIN_VALUE;
+    }
+
     View bottomSheet = viewRef.get();
     if (bottomSheet != null) {
       return calculateSlideOffset(bottomSheet.getTop());
--- a/lib/java/com/google/android/material/bottomsheet/BottomSheetBehavior.java
+++ b/lib/java/com/google/android/material/bottomsheet/BottomSheetBehavior.java
@@ -1068,7 +1068,7 @@ public class BottomSheetBehavior<V extends View> extends CoordinatorLayout.Behav
   /**
    * Calculates the current offset of the bottom sheet.
    *
-   * This method should be called when the child view is laid out. If not, 0 will be returned.
+   * This method should be called when the child view is laid out. If not, {@code Float.MIN_VALUE} will be returned.
    *
    * @return The offset of the bottom sheet, as a float value between -1 and 1. A negative value
    * corresponds to a hidden state, a zero value corresponds to a collapsed state, and a positive
@@ -1080,7 +1080,7 @@ public class BottomSheetBehavior<V extends View> extends CoordinatorLayout.Behav
       return calculateSlideOffset(bottomSheet.getTop());
     }
 
-    return 0;
+    return Float.MIN_VALUE;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


lib/java/com/google/android/material/bottomsheet/BottomSheetBehavior.java
```diff
@@ -1065,6 +1065,29 @@ public int getExpandedOffset() {
...
+   * Calculates the current offset of the bottom sheet.
+   *
+   * This method should be called when the child view is laid out.
+   *
+   * @return The offset of this bottom sheet within [-1,1] range. Offset increases
+   * as this bottom sheet is moving upward. From 0 to 1 the sheet is between collapsed and
+   * expanded states and from -1 to 0 it is between hidden and collapsed states. Returns
+   * {@code Float.MIN_VALUE} if the bottom sheet is not laid out.
+   */
+  public float calculateSlideOffset() {
```
drchen, line 1078: getCurrentSlideOffset()?

drchen (commented):
Can you add unit test for the new API as well?
https://github.com/material-components/material-components-android/blob/master/tests/javatests/com/google/android/material/bottomsheet/BottomSheetBehaviorTest.java

lib/java/com/google/android/material/bottomsheet/BottomSheetBehavior.java
OxygenCobalt: The calculation does work, so I feel like calculate is more appropriate.

OxygenCobalt:
I'm not too familiar with android unit testing @drchen. 

1. How do I run the tests? There's no specific option in android studio for me to run them.

1. Is a test like this appropriate? I gleamed other implementations to create it, however I'm not sure how correct or rigorous it is.
```java

  @Test
  @SmallTest
  public void testCalculateSlideOffset() {
    activityTestRule.runOnUiThread(
        () -> {
          BottomSheetBehavior<?> behavior = getBehavior();
          behavior.setState(BottomSheetBehavior.STATE_EXPANDED);
          assertThat(behavior.calculateSlideOffset(), is(1f));
          behavior.setState(BottomSheetBehavior.STATE_HALF_EXPANDED);
          assertThat(behavior.calculateSlideOffset(), is(0.5f));
          behavior.setState(BottomSheetBehavior.STATE_COLLAPSED);
          assertThat(behavior.calculateSlideOffset(), is(0f));
          behavior.setState(BottomSheetBehavior.STATE_HIDDEN);
          assertThat(behavior.calculateSlideOffset(), is(-1f));
        });
  }
```

drchen:
You can probably just run `./gradlew test` to run all unit tests? (Sorry I'm not entirely familiar with how to run a single test from gradle but it shouldn't take too long to run all of them.)

OxygenCobalt:
Yeah, I was able to run the tests @drchen. The tests seemed to pass, so I pushed the change.

lib/java/com/google/android/material/bottomsheet/BottomSheetBehavior.java
drchen: I guess it's more about our API convention. But feel free to leave it as-is. I'll see what our team will say in the internal review. : )

drchen (approved):

Niknaym (approved):

PR #2876, opened 2022-07-29
[BottomSheetBehavior] Check for shouldSkipHalfExpandedWhenDragging during nested scrolling

OxygenCobalt:
Make `BottomSheetBehavior` check the experimental `shouldSkipHalfExpandedWhenDragging` flag before setting the state to `HALF_EXPANDED`.

The code changes the following behaviors in `onStopNestedScroll`:
1. In the case of `dy > 0`, the state will be set to `STATE_EXPANDED` instead of `STATE_HALF_EXPANDED` when the flag is enabled.
2. In the case of `dy == 0`, the state will be set to `STATE_EXPANDED` when above the half-expanded offset, and `STATE_COLLAPSED` when below it when the flag is enabled.
3. In the state of `dy < 0`, the state will be set to `STATE_COLLAPSED` instead of `STATE_HALF_EXPANDED` when the flag is enabled.

### DIFF
--- a/lib/java/com/google/android/material/bottomsheet/BottomSheetBehavior.java
+++ b/lib/java/com/google/android/material/bottomsheet/BottomSheetBehavior.java
@@ -768,10 +768,14 @@ public class BottomSheetBehavior<V extends View> extends CoordinatorLayout.Behav
         targetState = STATE_EXPANDED;
       } else {
         int currentTop = child.getTop();
-        if (currentTop > halfExpandedOffset) {
-          targetState = STATE_HALF_EXPANDED;
-        } else {
+        if (currentTop < halfExpandedOffset) {
           targetState = STATE_EXPANDED;
+        } else {
+          if (shouldSkipHalfExpandedStateWhenDragging()) {
+            targetState = STATE_EXPANDED;
+          } else {
+            targetState = STATE_HALF_EXPANDED;
+          }
         }
       }
     } else if (hideable && shouldHide(child, getYVelocity())) {
@@ -796,10 +800,14 @@ public class BottomSheetBehavior<V extends View> extends CoordinatorLayout.Behav
             }
           }
         } else {
-          if (Math.abs(currentTop - halfExpandedOffset) < Math.abs(currentTop - collapsedOffset)) {
-            targetState = STATE_HALF_EXPANDED;
-          } else {
+          if (shouldSkipHalfExpandedStateWhenDragging()) {
             targetState = STATE_COLLAPSED;
+          } else {
+            if (Math.abs(currentTop - halfExpandedOffset) < Math.abs(currentTop - collapsedOffset)) {
+              targetState = STATE_HALF_EXPANDED;
+            } else {
+              targetState = STATE_COLLAPSED;
+            }
           }
         }
       }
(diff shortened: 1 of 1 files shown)


lib/java/com/google/android/material/bottomsheet/BottomSheetBehavior.java
```diff
@@ -796,10 +800,14 @@ public void onStopNestedScroll(
             }
           }
         } else {
-          if (Math.abs(currentTop - halfExpandedOffset) < Math.abs(currentTop - collapsedOffset)) {
-            targetState = STATE_HALF_EXPANDED;
-          } else {
+          if (shouldSkipHalfExpandedStateWhenDragging()) {
```
drchen, line 803: if (shouldSkipHalfExpandedStateWhenDragging() || Math.abs(currentTop - halfExpandedOffset) >= Math.abs(currentTop - collapsedOffset)) {

to avoid the below nested if

lib/java/com/google/android/material/bottomsheet/BottomSheetBehavior.java
```diff
@@ -809,10 +817,14 @@ public void onStopNestedScroll(
       } else {
         // Settle to nearest height.
         int currentTop = child.getTop();
-        if (Math.abs(currentTop - halfExpandedOffset) < Math.abs(currentTop - collapsedOffset)) {
-          targetState = STATE_HALF_EXPANDED;
-        } else {
+        if (shouldSkipHalfExpandedStateWhenDragging()) {
```
drchen, line 820: ditto

lib/java/com/google/android/material/bottomsheet/BottomSheetBehavior.java
OxygenCobalt: Completed in 7564954

lib/java/com/google/android/material/bottomsheet/BottomSheetBehavior.java
OxygenCobalt: Completed in 7564954.

OxygenCobalt:
Okay, simplified those if statements @drchen.

drchen (approved):

OxygenCobalt:
Completely forgot about this PR. Any updates on this too? @drchen

drchen:
We found some problems during the internal review. 

I need to take more time to understand the relevant logic and figure out what's the cause. : )

OxygenCobalt:
Could you elaborate on the issue @drchen? I'm using this fix in my app, so it might affect me. I could try looking into it further.

drchen:
My teammate patched and tested the PR and said:

"""
...the scroll behavior is almost completely broken for this case -- it doesn't expand upwards when you try to drag it up. it looks like this is because it only behaves properly when shouldExpandOnUpwardsDrag() is also set to true. wdyt about removing that flag? i can't imagine a scenario where we'd want it set to false. also, that shouldExpandOnUpwardsDrag() only takes effect when used with shouldSkipHalfExpandedStateWhenDragging(), which is pretty unexpected. i think we should delete that flag and just add another shouldSkipHalfExpandedStateWhenDragging() check next to the fitToContents offset check in onViewReleased().
"""

Haven't really got time to check it yet.

OxygenCobalt:
Oh yeah, I already know about that bug @drchen. It was present before my fix, and I'm really not sure what the point behind it was.

I could go look into it and try to remove that function. Could I append it to this PR or should I go and make a new one?

OxygenCobalt:
Any updates here @drchen? Should I make another PR to remove `shouldExpandOnUpwardsDrag` so this can get unblocked?

drchen:
Hi the change is pretty difficult to land the change due to all the corner cases we found during running internal tests.

I'll need a less busy timespan to go back working on this. Sorry for the long delay.

OxygenCobalt:
Sorry for taking awhile to respond @hellosagar. I don't really mind if you copy this changeset into your own vendored bottom sheet.

Issue #2786, opened 2022-06-27
[BottomSheetBehavior] Allow programmatically setting `STATE_HIDDEN` even when non-hideable


OxygenCobalt:
**Is your feature request related to a problem? Please describe.**
My app has a Bottom Sheet that I do not want to be hidden by the user, as it contains important information. However, it still needs to be hidden programmatically when there is nothing to show. However, if I disable hiding with `isHideable`, I can no longer change the state programmatically due to [these lines in `BottomSheetBehavior`](https://github.com/material-components/material-components-android/blob/cd37b9170c149de80ca5729035b5a2ad33afc9dd/lib/java/com/google/android/material/bottomsheet/BottomSheetBehavior.java#L1227):

```
if (!hideable && state == STATE_HIDDEN) {
  Log.w(TAG, "Cannot set state: " + state);
  return;
}
```

**Describe the solution you'd like**
The removal of the `isHideable` check when setting the state to `STATE_HIDDEN`. Allow it to be set programmatically, and change `isHideable` to only apply to user input.

**Describe alternatives you've considered**
I could modify `isHideable` during run-time as such:
```kotlin
if (hide) {
    behavior.isHideable = true
    behavior.state = BottomSheetBehavior.STATE_HIDDEN
} else {
    behavior.isHideable = false
    behavior.state = BottomSheetBehavior.STATE_COLLAPSED
}
```

But I worry about the implications this might have for state consistency.

drchen:
As for the state consistency - the alternative you mentioned is actually better, in my personal opinion. 

I'll close the issue for now since I don't think it's a good idea to alter a released behavior globally when there's a working alternative which makes sense logically.

OxygenCobalt:
Okay, no problem.

OxygenCobalt:
Sorry to revive this, but I've actually encountered layout issues with the workaround @drchen. For some reason setting those flags to true early on enough in the layout process causes the bottom sheet to flicker to the expanded state and then animate back to the collapsed state. Sadly, I can't seem to reproduce this in a test app, so I can't file any meaningful issue or debug it. All I know is that not mutating `isHideable` and just keeping it at `true` fixes the issue.

[...]

PR #2880, opened 2022-08-01
[BottomSheetBehavior] Add internal method to disable hiding gestures

OxygenCobalt:
Some users of `BottomSheetBehavior` need to enable `isHideable` so that they can set a state programmatically, but still want to disable hiding in the UI, as the hiding action would represent something too destructive (Removing the currently playing song, for example). This was discussed in #2786 and a workaround proposed, but the workaround did not necessarily work in all situations, and it was largely a band-aid around the larger issue that `isHideable` simultaneously represents the ability for the bottom sheet to hide programmatically, and the ability to hide the bottom sheet.

This PR adds a new internal method called `shouldEnableHidingGestures` (a la `shouldSkipHalfExpandedStateWhileDragging`) that allows hiding gestures to be disabled even when `isHideable` is enabled. This preserves the behavior of `isHideable` while also allowing a escape hatch for those who need the hidden state internally, but don't want to expose such to the user.

Also see #2786.

### DIFF
--- a/lib/java/com/google/android/material/bottomsheet/BottomSheetBehavior.java
+++ b/lib/java/com/google/android/material/bottomsheet/BottomSheetBehavior.java
@@ -725,7 +725,7 @@ public class BottomSheetBehavior<V extends View> extends CoordinatorLayout.Behav
       }
     } else if (dy < 0) { // Downward
       if (!target.canScrollVertically(-1)) {
-        if (newTop <= collapsedOffset || (hideable && shouldEnableHidingGestures())) {
+        if (newTop <= collapsedOffset || (hideable && isHideableWhenDragging())) {
           if (!draggable) {
             // Prevent dragging
             return;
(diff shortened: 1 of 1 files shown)


lib/java/com/google/android/material/bottomsheet/BottomSheetBehavior.java
```diff
@@ -1867,6 +1872,16 @@ public boolean shouldSkipSmoothAnimation() {
     return true;
   }
 
+  /**
+   * Checks whether hiding gestures should be enabled while {@code isHideable} is set to true.
+   *
+   * @hide
+   */
+  @RestrictTo(LIBRARY_GROUP)
+  public boolean shouldEnableHidingGestures() {
```
drchen, line 1881: isHideableWhenDragging()? To be consistent with isHideable() and shouldSkipHalfExpandedStateWhenDragging()?

OxygenCobalt: Sure.

OxygenCobalt:
Hey, can you re-review now @drchen? I made the change you requested.

drchen (approved):

drchen:
LGTM. I need to send it for internal review and see what other folks think though. Stay tuned.

OxygenCobalt:
Any update on the review process for this one? @drchen

drchen:
Hi we are having some debate over the method naming so it takes a while. (Plus I'm super busy these couple of weeks..)

I hope we can get this submitted today.

PR #2888, opened 2022-08-03
[SwipeDismissBehavior]fix SwipeDissmissBehavior not work when down touch event consumed by …

### LINKED ISSUE #1333: [Snackbar] Custom view can't be dragged
I have created a custom view extending from the BaseTransientBottomBar class. I can't drag to dismiss this view, so no animation is visible for dragging. 

When I'm trying to drag the view, the message below is shown in my logcat console :
"Ignoring pointerId=0 because ACTION_DOWN was not received for this pointer before ACTION_MOVE. It likely happened because  ViewDragHelper did not receive all the events in the event stream."

class CustomSnackBar (parent: ViewGroup, content: CustomView) : BaseTransientBottomBar< CustomView >(parent, content, content){

.....
}

val behavior = Behavior().apply {
                setSwipeDirection(SwipeDismissBehavior.SWIPE_DIRECTION_END_TO_START)
                setDragDismissDistance(0.3f)
            }

customSnackBar.setDuration(Snackbar.LENGTH_INDEFINITE).setBehavior(behavior)

**Android target and compiled SdkVersion:**  : 28

**Material Library version:**  tested with 1.1.0,  1.2.0-alpha05, 1.2.0-alpha07 (the same problem)

kingwang666:
Fix #1333
This bug is caused by two reasons
1.  Recieving a ACTION_CANCEL event from the CoordinatorLayout when the SwipeDismissBehavior calls requestDisallowInterceptTouchEvent in onViewCaptured.
2. ACTION_DOWN consumed by other view, causing the Behavior not to call back onTouchEvent

### DIFF
--- a/lib/java/com/google/android/material/behavior/SwipeDismissBehavior.java
+++ b/lib/java/com/google/android/material/behavior/SwipeDismissBehavior.java
@@ -89,6 +89,8 @@ public class SwipeDismissBehavior<V extends View> extends CoordinatorLayout.Beha
   private float sensitivity = 0f;
   private boolean sensitivitySet;
 
+  private boolean ignoreCancelEvent = false;
+
   int swipeDirection = SWIPE_DIRECTION_ANY;
   float dragDismissThreshold = DEFAULT_DRAG_DISMISS_THRESHOLD;
   float alphaStartSwipeDistance = DEFAULT_ALPHA_START_DISTANCE;
@@ -212,6 +214,12 @@ public class SwipeDismissBehavior<V extends View> extends CoordinatorLayout.Beha
   @Override
   public boolean onTouchEvent(CoordinatorLayout parent, V child, MotionEvent event) {
     if (viewDragHelper != null) {
+      // Ignore ACTION_CANCEL event from the CoordinatorLayout when the SwipeDismissBehavior calls
+      // requestDisallowInterceptTouchEvent in onViewCaptured
+      if (ignoreCancelEvent && event.getActionMasked() == MotionEvent.ACTION_CANCEL) {
+        ignoreCancelEvent = false;
+        return true;
+      }
       viewDragHelper.processTouchEvent(event);
       return true;
     }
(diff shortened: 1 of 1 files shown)


kingwang666:
这是来自QQ邮箱的自动回复邮件。邮件已收到

lib/java/com/google/android/material/behavior/SwipeDismissBehavior.java
```diff
@@ -89,6 +89,8 @@
   private float sensitivity = 0f;
   private boolean sensitivitySet;
 
+  private boolean ignoreCancelEvent = false;
```
pekingme, line 92: nit: Could you move this to L88?

pekingme (approved):
Thanks for the PR! Left a minor comment.

lib/java/com/google/android/material/behavior/SwipeDismissBehavior.java
kingwang666: ok~ @pekingme

pekingme:
I approved cuz' I didn't see the error mentioned in the issue with this change. But when I took another look, I cannot repro the error messages with the current snapshot. It seems it got fixed at some point (the issue has a bit history now). Could you confirm? If so, I'll close this PR and the issue.

kingwang666:
这是来自QQ邮箱的自动回复邮件。邮件已收到

kingwang666:
> I approved cuz' I didn't see the error mentioned in the issue with this change. But when I took another look, I cannot repro the error messages with the current snapshot. It seems it got fixed at some point (the issue has a bit history now). Could you confirm? If so, I'll close this PR and the issue.

@pekingme Hello, I tested and it still exists. You can use `CardSwipeDismissFragment` to test it. 
set `OnClickListener` to `cardContentLayout`.  The `ACTION_DOWN` consumed by `cardContentLayout`

````java
cardContentLayout.setOnClickListener(new View.OnClickListener() {
  @Override
  public void onClick(View v) {
    
  }
});
````

The last master code.   

https://user-images.githubusercontent.com/16848272/192931246-1ab78794-032f-4611-8c8f-b88b8b491acf.mp4 

After set onClickListener.  


https://user-images.githubusercontent.com/16848272/192931539-263317a9-0515-4ff6-afd0-7bedf1b6a651.mp4  

After fixed. 


https://user-images.githubusercontent.com/16848272/192931656-2b074c1f-7bb5-4e17-942e-bb3744f9e4e5.mp4

pekingme:
Thanks for the recording! Yes, I can reproduce the same behavior with the catalog built **externally**. Was testing with the catalog built internally. It seems the fix works well with `androidx.coordinatorlayout:coordinatorlayout:1.1.0`. But we use a more recent version. If you update `androidx.coordinatorlayout:coordinatorlayout` to `1.2.0`, the fix won't work anymore. 

I don't see any regression with this fix internally. Let me test this out a bit more. Meanwhile I'm looking for a fix that works with the `1.2.0` and above now. Let me know if you find anything. TIA

pekingme:
Ok, I have updated the fix (internally) to make it works for both 1.1.0 and later version. Should be merged soon.

PR #2898, opened 2022-08-08
[Catalog] Replace android.support dependencies with androidx

pubiqq:


drchen:
Thanks for noticing this! We needed to fix it from an internal path due to some script settings, though. But still thanks for the PR!

PR #2915, opened 2022-08-17
[ShapeAppearance] Fix rotation angle calculation

pubiqq:


drchen (approved):
oh my....  Thanks for catching this!

pekingme:
Wow! Good job!

Issue #543, opened 2019-08-28
[BottomNavigationView] Allow to set more then 5 menu items


palatin:
Hello, I understand that for correct displaying, limitation of items in bottom navigation view were set. But why is it hardcoded and final? For example, for tablet in landscape mode I want to have 7 items. Can you please allow to override max menu items.

antonkovalyov351:
Up

rmasarovic:
Hi i have same problem and only one way how to do it is fork code...

drchen:
We have design concerns around allowing more items than 5. If you really need that, you can extend BottomNavigationView and override getMaxItemCount() to provide a different value than 5.

NoorTech98:
same problem in bottom navigation view 
solved the solution item code

PR #2923, opened 2022-08-19
[BottomNavigation] Fix hardcoded number of children

anyercastillo:
Fixes a bug that prevents developers to define an arbitrary number of maximum item in `BottomNavigationView`.

**The problem.**
 `BottomNavigationMenuView` was using the static constant `BottomNavigationView.MAX_ITEM_COUNT` as part of its logic. So overriding  `BottomNavigationView.getMaxItemCount()` with a value greater than `BottomNavigationView.MAX_ITEM_COUNT` causes the exception `ArrayIndexOutOfBondsException` 

Related to https://github.com/material-components/material-components-android/issues/543


akwappnet (approved):

imhappi:
lgtm 👍

Issue #1578, opened 2020-08-06
[RangeSlider] theming of valueLabel applies only to unselected label


NoyaD9:
**Description:**  After theming a label using the following attributes:

```
  <com.google.android.material.slider.RangeSlider
      ...
      app:labelStyle="@style/Filters.Style.Criteria.SliderSectionItem.SliderLabel"
      />
```

in styles.xml:

```
  <style name="Filters.Style.Criteria.SliderSectionItem.SliderLabel" parent="Widget.MaterialComponents.Tooltip">
    <item name="backgroundTint">?cosmaColorFloatingButtonSecondaryVariantBackground</item>
    <item name="android:textAppearance">@style/Filters.TextAppearance.Criteria.SliderSectionItem.SliderLabel</item>
  </style>

  <style name="Filters.TextAppearance.Criteria.SliderSectionItem.SliderLabel" parent="TextAppearance.MaterialComponents.Tooltip">
    <item name="android:textColor">?colorSurface</item> 
    <item name="fontFamily">@font/cosma_make_it_sans</item>
    <item name="android:fontFamily">@font/cosma_make_it_sans</item>
  </style>
```
I see only the currently unselected item of the range being styled properly:

![sliderbug](https://user-images.githubusercontent.com/25151005/89536834-bfec0a00-d7f8-11ea-8904-4454ea0ce519.png)

**Expected behavior:** both value label are having the same style (black with white text)

**Source code:**  see above

**Android API version:** 28

**Material Library version:** 1.2.0

**Device:** Samsung Galaxy S10

NoyaD9:
FYI: I was also able to reproduce it on an Emulator with API 29 and by hardcoding the colors instead of using the attributes

```
  <style name="Filters.Style.Criteria.SliderSectionItem.SliderLabel" parent="Widget.MaterialComponents.Tooltip">
    <item name="backgroundTint">@color/cosma_charcoal</item>
    <item name="android:textAppearance">@style/Filters.TextAppearance.Criteria.SliderSectionItem.SliderLabel</item>
  </style>

  <style name="Filters.TextAppearance.Criteria.SliderSectionItem.SliderLabel" parent="TextAppearance.MaterialComponents.Tooltip">
    <item name="android:textColor">@color/cosma_white</item>
    <item name="fontFamily">@font/cosma_make_it_sans</item>
    <item name="android:fontFamily">@font/cosma_make_it_sans</item>
  </style>
```

cketcham:
[...]

PR #2931, opened 2022-08-22
[Slider] Do not capture inflation parameters for reuse

pubiqq:
Fixes https://github.com/material-components/material-components-android/issues/1578

### DIFF
--- a/lib/java/com/google/android/material/slider/BaseSlider.java
+++ b/lib/java/com/google/android/material/slider/BaseSlider.java
@@ -246,11 +246,7 @@ abstract class BaseSlider<
   private final AccessibilityManager accessibilityManager;
   private AccessibilityEventSender accessibilityEventSender;
 
-  private interface TooltipDrawableFactory {
-    TooltipDrawable createTooltipDrawable();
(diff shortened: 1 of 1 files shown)


drchen (commented):
It looks like there's no behavioral change (which is also funny..). Am I missing anything because you said it probably fixes an issue?

pubiqq:
I was just able to reproduce https://github.com/material-components/material-components-android/issues/1578 with [this example](https://github.com/material-components/material-components-android/issues/1578#issuecomment-1209048542). And yes, it looks like the PR really resolves the issue.

drchen:
That's pretty bizarre to me... can you explain why the behavior can be different?

pubiqq:
`attrs` is not just an `AttributeSet`, it's an `XmlResourceParser`. And when view inflation completes, the parser closes, and you can no longer get values from it. In other words, trying to preserve `attrs` is useless, after inflation it will not give correct values.

drchen:
Interesting! That makes sense. : )

drchen (approved):

PR #2934, opened 2022-08-25
[Internal] Set the outline to a path whenever possible

pubiqq:


lib/java/com/google/android/material/drawable/DrawableUtils.java
```diff
@@ -277,4 +279,31 @@ public static int[] getUncheckedState(@NonNull int[] state) {
...
+      return true;
+    }
+
+    if (VERSION.SDK_INT >= VERSION_CODES.Q) {
+      // As of Android Q, the restriction that the path must be convex is removed.
+      outline.setConvexPath(path);
+      return true;
+    }
+
+    if (VERSION.SDK_INT >= VERSION_CODES.LOLLIPOP) {
```
drchen, line 298: It think it can be just written as:

```
if (VERSION.SDK_INT >= VERSION_CODES.LOLLIPOP && path.isConvex()) {
  outline.setConvexPath(path);
  return true;
}

return false;

Comment (pubiqq) [reply lib/java/com/google/android/material/drawable/DrawableUtils.java:298]: It seems to me that such a structure is easier for other developers to understand, even though it's a bit longer:

```java
if (VERSION.SDK_INT >= VERSION_CODES.R) {
    // Impl for R+
    return Api30Impl.setOutlineToPath(...);
}

if (VERSION.SDK_INT >= VERSION_CODES.Q) {
    // Impl for [Q, R)
    return Api29Impl.setOutlineToPath(...);
}

if (VERSION.SDK_INT >= VERSION_CODES.LOLLIPOP) {
    // Impl for [L, Q)
    return Api21Impl.setOutlineToPath(...);
}

return false;
```

But if you want, I can rewrite it.

drchen: Well we are not using Impl pattern here anyways. : )

To me it's more straightforward to just make all first three ifs do the actual work and return true, and to let the default case return false.

pubiqq: I was referring to a technique where outer `if` statements are used to check the version:

```java
if (VERSION.SDK_INT >= VERSION_CODES.R) {
    // Impl for R+
    return ...
}

if (VERSION.SDK_INT >= VERSION_CODES.Q) {
    // Impl for [Q, R)
    return ...
}

if (VERSION.SDK_INT >= VERSION_CODES.LOLLIPOP) {
    // Impl for [L, Q)
    return ...
}

// Fallback for K-
return false;
```

This approach is more common and perhaps easier for developers to perceive.

pubiqq: Although, you know, nvm.

drchen (approved):

lib/java/com/google/android/material/drawable/DrawableUtils.java
drchen: I understand your point. : )

But I think we usually prefer less nested depth and less return statements. In this case I think this makes the code much cleaner. : )

Issue #2950, opened 2022-09-01
[Slider] Thumb halo is active for no reason (labeled slider, API 19)


pubiqq:
<table>
  <tr>
    <th>SliderLabelBehaviorDemoFragment<br>on API 19</th>
    <th>SliderLabelBehaviorDemoFragment<br>on API 29</th>
   </tr> 
   <tr>
    <td><img src="https://user-images.githubusercontent.com/82187521/187992230-c8fe2310-061c-4303-a5dc-db0e33df6566.png" width="324px"></td>
    <td><img src="https://user-images.githubusercontent.com/82187521/187992273-bb2e721e-9b12-47f1-b018-de5e30d308be.png" width="324px"></td>
    </tr>
</table>

**Material Library version:** 1.7.0-rc01

drchen:
Interesting. Does this happen with 1.6?

pubiqq:
Yes, it does.

drchen:
Thanks! Just wanted to make sure it's not a regression from 1.6.

PR #2951, opened 2022-09-01
[Slider] Fix compat halo visibility

pubiqq:
Fixes https://github.com/material-components/material-components-android/issues/2950


drchen (approved):
Thanks for the fix!!

PR #2952, opened 2022-09-01
[MaterialTimePicker] Fix divider alignment on the clock display

pubiqq:
And we continue to pick up the [lost fixes](https://github.com/material-components/material-components-android/pull/2921) 😄

The divider (colon) should be aligned as intended by font designer. Do not try to align it yourself, the hardcoded indentation may be incorrect with a different font.

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
   </tr> 
   <tr>
    <td></td>
    <td></td>
    </tr>
</table>


paulfthomas:
Hi @pubiqq, thank you for your great contributions, they are really appreciated!

Sorry that some of your other and valid suggestions were not all included previously, it is easier/faster for us to review and merge one change at a time.

Regarding this alignment issue, as per my initial testing, a lot of our screenshot tests are impacted as the divider appears slightly lower. I will reach out to our design team to see what we can do.

pubiqq:
> Regarding this alignment issue, as per my initial testing, a lot of our screenshot tests are impacted as the divider appears slightly lower.

Yes, I understand that, but it happens because font designer intended it that way:



But ok, let's see what your design team says first.

McCallumDillon:
Chiming in here on the design side– _**agreed**_, the proposed "after" with dividing colon height matching that of the adjacent input fields and also aligning with spec, guidelines, and design kit implementation.

PR #2960, opened 2022-09-06
[MaterialSwitch] Add support for setting thumb icon size

pubiqq:


lib/java/com/google/android/material/materialswitch/res/values/attrs.xml
```diff
@@ -46,6 +46,8 @@
            result to valid color values. Saturate(S + D) -->
       <enum name="add" value="16" />
     </attr>
+    <!-- Size that will be applied to the thumb icon drawable. -->
```
afohrman, line 49: I know this naming pattern is used throughout this file, but it's wordy, especially for something as straightforward as size. You can replace "that will be applied to" with "of" or "used for".

lib/java/com/google/android/material/drawable/DrawableUtils.java
```diff
@@ -205,44 +228,51 @@ public static Drawable compositeTwoLayeredDrawable(
     }
     LayerDrawable drawable =
         new LayerDrawable(new Drawable[] {bottomLayerDrawable, topLayerDrawable});
+
+    final int topLayerWidth = topLayerDrawable.getIntrinsicWidth();
```
afohrman, line 232: For clarity, I suggest renaming to topLayerPreviousWidth. Same for the other three new width/height variables.

lib/java/com/google/android/material/drawable/DrawableUtils.java
```diff
@@ -205,44 +228,51 @@ public static Drawable compositeTwoLayeredDrawable(
...
-      // If there's no intrinsic width or height, keep bottom layer's size.
-      topLayerNewWidth = bottomLayerDrawable.getIntrinsicWidth();
-      topLayerNewHeight = bottomLayerDrawable.getIntrinsicHeight();
-    } else if (topLayerDrawable.getIntrinsicWidth() <= bottomLayerDrawable.getIntrinsicWidth()
-        && topLayerDrawable.getIntrinsicHeight() <= bottomLayerDrawable.getIntrinsicHeight()) {
-      // If the top layer is smaller than the bottom layer in both its width and height, keep top
-      // layer's size.
-      topLayerNewWidth = topLayerDrawable.getIntrinsicWidth();
-      topLayerNewHeight = topLayerDrawable.getIntrinsicHeight();
+    if (topLayerDesiredWidth != -1 && topLayerDesiredHeight != -1) {
```
afohrman, line 239: Is there currently a case where topLayerDesiredWidth will be -1? Looks like the default set in MaterialSwitch's constructor is 0, which is technically a valid size. What case is this -1 check trying to handle?

afohrman (changes requested):
Thanks for the PR! Looks good, see comments within.

lib/java/com/google/android/material/drawable/DrawableUtils.java
pubiqq: I think `***PreviousWidth` is even worse, but I think I get what you mean. How about `topLayerOriginalWidth`?

lib/java/com/google/android/material/drawable/DrawableUtils.java
pubiqq: `compositeTwoLayeredDrawable` is also used in `MaterialCheckBox`.

lib/java/com/google/android/material/drawable/DrawableUtils.java
afohrman: `topLayerOriginalWidth` sounds great to me.

lib/java/com/google/android/material/drawable/DrawableUtils.java
afohrman: 👍

lib/java/com/google/android/material/materialswitch/res/values/attrs.xml
afohrman: Did you forget to merge the new changes?

pubiqq: No, I wanted to get all my questions answered first 😄

lib/java/com/google/android/material/drawable/DrawableUtils.java
pubiqq: Good.

lib/java/com/google/android/material/materialswitch/res/values/attrs.xml
afohrman: Ah got it :)

afohrman (approved):
LGTM, thanks for the PR! I'll start merging this in.

afohrman:
Update: While preparing to merge internally, I noticed this regression:

Before:
image
After:
image

I'm working on debugging the source of the issue (it's possible it's just a test issue) and you're welcome to help track it down as well, but until then this PR is on hold.

pubiqq:
Could you give me some information on how to reproduce the issue?

pubiqq:
any news?

Mids9 (approved):

pubiqq:
mb now?

pubiqq:
bump

pubiqq:
bump

pubiqq:
bump

pubiqq:
bump

pubiqq:
bump

lib/java/com/google/android/material/drawable/DrawableUtils.java
```diff
@@ -207,44 +230,52 @@ public static Drawable compositeTwoLayeredDrawable(
...
```
drchen, line 241: I think we can just adjust width & height separately like:

topLayerNewWidth = topLayerDesiredWidth == -1 ? topLayerOriginalWidth : topLayerDesiredWidth
topLayerNewHeight = topLayerDesiredHeight == -1 ? topLayerOriginalHeight : topLayerDesiredHeight

And then fall back to the original logic.

lib/java/com/google/android/material/drawable/DrawableUtils.java
```diff
@@ -207,44 +230,52 @@ public static Drawable compositeTwoLayeredDrawable(
     }
     LayerDrawable drawable =
         new LayerDrawable(new Drawable[] {bottomLayerDrawable, topLayerDrawable});
+
+    final int topLayerOriginalWidth = topLayerDrawable.getIntrinsicWidth();
```
drchen, line 234: nit)  name those variables as:

topLayerIntrinsicWidth, etc.

drchen: I'll check why the tests are broken. Can you adjust the PR according to the comment first?

lib/java/com/google/android/material/drawable/DrawableUtils.java
pubiqq: I suppose we could. But I decided to fix some edge cases, that's why there are a little more changes than expected.

lib/java/com/google/android/material/drawable/DrawableUtils.java
pubiqq: > nit) name those variables as...

Done

lib/java/com/google/android/material/drawable/DrawableUtils.java
drchen: I guess a more important question here is if we want to handle width and height separately?

Like, can we only specify a desired width but not a desired height?

lib/java/com/google/android/material/drawable/DrawableUtils.java
```diff
@@ -207,46 +241,70 @@ public static Drawable compositeTwoLayeredDrawable(
...
+      if (topLayerNewWidth != -1 && topLayerNewHeight != -1) {
+        drawable.setLayerSize(1, topLayerNewWidth, topLayerNewHeight);
+      }
       drawable.setLayerGravity(1, Gravity.CENTER);
     } else {
-      int horizontalInset = (bottomLayerDrawable.getIntrinsicWidth() - topLayerNewWidth) / 2;
-      int verticalInset = (bottomLayerDrawable.getIntrinsicHeight() - topLayerNewHeight) / 2;
-      drawable.setLayerInset(1, horizontalInset, verticalInset, horizontalInset, verticalInset);
+      //noinspection StatementWithEmptyBody
+      if (topLayerNewWidth != -1 && topLayerNewHeight != -1) {
```
drchen, line 293: Please flatten the nested ifs here. Something like:

if (topLayerNewWidth == -1 
    || topLayerNewHeight == -1 
    || bottomLayerIntrinsicWidth == -1
    || bottomLayerIntrinsicHeight == -1) {
  Log.w(...)
} else {
 ...
}
  
}

lib/java/com/google/android/material/drawable/DrawableUtils.java
pubiqq: Ah, got it. Well, I can add this, why not.

pubiqq:
done btw

lib/java/com/google/android/material/checkbox/MaterialCheckBox.java
```diff
@@ -732,7 +732,7 @@ private void refreshButtonDrawable() {
     updateButtonTints();
 
     super.setButtonDrawable(
-        DrawableUtils.compositeTwoLayeredDrawable(buttonDrawable, buttonIconDrawable));
```
drchen, line 735: Please keep the original method name. I believe it's being verbose for a reason.

lib/java/com/google/android/material/drawable/DrawableUtils.java
```diff
@@ -190,63 +194,124 @@ private static Drawable createTintableMutatedDrawableIfNeeded(
...
-   * @param bottomLayerDrawable the drawable to be on the first layer (bottom)
-   * @param topLayerDrawable the drawable to be on the second layer (top)
+   * @param bottomLayerDrawable the drawable to be on the bottom layer
+   * @param topLayerDrawable the drawable to be on the top layer
    */
   @Nullable
-  public static Drawable compositeTwoLayeredDrawable(
+  public static Drawable compositeDrawables(
       @Nullable Drawable bottomLayerDrawable, @Nullable Drawable topLayerDrawable) {
+
```
drchen, line 208: Why don't you call the overloading method with desired width/height as -1? 

Also it's confusing to have overloading methods with inconsistent behavior. What's the purpose of this newly introduced method?

drchen, line 208: Why don't you call the overloading method with desired width/height as -1? 

Also it's confusing to have overloading methods with inconsistent behavior. What's the purpose of this newly introduced method?

lib/java/com/google/android/material/checkbox/MaterialCheckBox.java
pubiqq: The method doesn't guarantee it always returns layered drawable, so, technically, the original name is incorrect, that's why I've changed it.

lib/java/com/google/android/material/drawable/DrawableUtils.java
pubiqq: Compositing with intrinsic size (without custom one) works on any API.
Compositing with custom size works only on API 23+.

The previous method was only half working, and I had to split it into two fully working methods.

pubiqq: see above

pubiqq: duplicate I suppose?

lib/java/com/google/android/material/checkbox/MaterialCheckBox.java
drchen: I don't think the original method name implies that....

I think "two layered drawable" refers to the input values, not the output value. 

Another thing is restricted APIs are still technically "public". There can still be clients affected by this change. Let avoid unnecessary changes like this as much as possible.

lib/java/com/google/android/material/materialswitch/MaterialSwitch.java
```diff
@@ -369,8 +404,23 @@ private void refreshThumbDrawable() {
...
     updateDrawableTints();
 
-    super.setThumbDrawable(
-        DrawableUtils.compositeTwoLayeredDrawable(thumbDrawable, thumbIconDrawable));
+    final Drawable thumbWithIconDrawable;
+    if (VERSION.SDK_INT >= VERSION_CODES.M) {
+      thumbWithIconDrawable = DrawableUtils.compositeDrawables(
+          thumbDrawable, thumbIconDrawable, thumbIconSize, thumbIconSize);
+    } else {
+      if (thumbIconDrawable != null
```
drchen, line 412: The default value of thumbIconSize is 0. In other words, this condition will be triggered by most of the default settings, which is wrong. 

I think it is fine to just ignore the value without leaving logs.

lib/java/com/google/android/material/drawable/DrawableUtils.java
```diff
@@ -190,63 +194,124 @@ private static Drawable createTintableMutatedDrawableIfNeeded(
...
+    if (topLayerDrawable == null) {
+      return bottomLayerDrawable;
+    }
+
+    final int topLayerIntrinsicWidth = topLayerDrawable.getIntrinsicWidth();
+    final int topLayerIntrinsicHeight = topLayerDrawable.getIntrinsicHeight();
+    final int bottomLayerIntrinsicWidth = bottomLayerDrawable.getIntrinsicWidth();
+    final int bottomLayerIntrinsicHeight = bottomLayerDrawable.getIntrinsicHeight();
+
+    int topLayerNewWidth =
```
drchen, line 273: Declare int constants instead of using magic number -1.

lib/java/com/google/android/material/drawable/DrawableUtils.java
```diff
@@ -190,63 +194,124 @@ private static Drawable createTintableMutatedDrawableIfNeeded(
...
+        topLayerDesiredWidth != -1 ? topLayerDesiredWidth : topLayerIntrinsicWidth;
+    int topLayerNewHeight =
+        topLayerDesiredHeight != -1 ? topLayerDesiredHeight : topLayerIntrinsicHeight;
+
+    // If there's no intrinsic width or height, keep bottom layer's size.
```
drchen, line 278: As I mentioned before, let's keep the original logic as much as possible. 

We should only override topLayerIntrinsicWidth / topLayerIntrinsicHeight with desired width / height in this change and keep other logic intact in this class. 

Unless - if we think it's meaningful to always respect desired width / height - even if they are larger than the bottom layer. In that case we can rethink a more clear structure to do that.

lib/java/com/google/android/material/checkbox/MaterialCheckBox.java
pubiqq: If you are right, the name is even more meaningless and incorrect. First, the method already takes two arguments, no need to indicate it also in the name. Second, it's not guaranteed that both drawings must be layered. And even if it were so, it would be enough to declare the input parameters types as `LayerDrawable` and this "name refinement" would again be redundant.

About `@RestrictTo`: Restricted APIs are `public` only because there is no equivalent of `internal` in Java. And if the client explicitly ignored the restrictions of some API, it should definitely have been prepared for the potential consequences.

I think it's better to fix the name instead of leaving it only because someone might use this "only for Material internal purposes" API.

lib/java/com/google/android/material/materialswitch/MaterialSwitch.java
pubiqq: The condition will be triggered only for API < 23 and only for drawables that are not `thumbIconSize`x`thumbIconSize`. Logs are necessary here: user should be informed about why the custom icon size has not been applied.

As for size, default `thumbIconSize` for M3 theme - 16dp (`@dimen/mtrl_switch_thumb_icon_size`). The default value in `MaterialSwitch` constructor will be changed to intrinsic in the next rebase.

lib/java/com/google/android/material/drawable/DrawableUtils.java
pubiqq: Android doesn't seem to have `Drawable.INTRINSIC_SIZE` or something...

Are you proposing to create something like that?
```java
@RestrictTo(Scope.LIBRARY_GROUP)
public final class DrawableUtils {

  public static final int INTRINSIC_SIZE = -1;

  ...
}
```

lib/java/com/google/android/material/drawable/DrawableUtils.java
pubiqq: > As I mentioned before, let's keep the original logic as much as possible.
>
> We should only override topLayerIntrinsicWidth / topLayerIntrinsicHeight with desired width / height in this change and keep other logic intact in this class.

It doesn't work. On pre-M, desired sizes will be ignored because pre-M `LayerDrawable` doesn't support scaling, and you will end up with a drawable that is both incorrectly sized and incorrectly centered.

lib/java/com/google/android/material/drawable/DrawableUtils.java
drchen: Yes, that's what I mean. And maybe call it UNSPECIFIED_SIZE?

lib/java/com/google/android/material/drawable/DrawableUtils.java
drchen: oh ok. I see what you mean. Maybe let me download the PR and play with it by myself for a little bit.

lib/java/com/google/android/material/materialswitch/MaterialSwitch.java
drchen: Thanks for the explanation. I'm wondering if we can just restrict the attribute to API 23+ so we don't have to worry about this here. Let me check that with the team.

lib/java/com/google/android/material/drawable/DrawableUtils.java
pubiqq: Well, technically, "intrinsic size" cannot be called unspecified. And I think `use INTRINSIC_SIZE to use the intrinsic size` a bit clearer than `use UNSPECIFIED_SIZE to use the intrinsic size`. But if you insist, I can change it.

pubiqq:
bump

pubiqq:
bump

PR #2974, opened 2022-09-16
[MaterialButtonToggleGroup] handle setEnable childs

BluestormDNA:
This PR aims to to propagate MaterialButtonToggleGroup setEnabled to it's MaterialButton children

Closes 2962
Closes 1813

### DIFF
--- a/lib/java/com/google/android/material/button/MaterialButtonToggleGroup.java
+++ b/lib/java/com/google/android/material/button/MaterialButtonToggleGroup.java
@@ -171,6 +171,7 @@ public class MaterialButtonToggleGroup extends LinearLayout {
   private boolean skipCheckedStateTracker = false;
   private boolean singleSelection;
   private boolean selectionRequired;
+  private boolean isEnabled;
 
   @IdRes
   private final int defaultCheckId;
@@ -201,6 +202,8 @@ public class MaterialButtonToggleGroup extends LinearLayout {
     selectionRequired =
         attributes.getBoolean(R.styleable.MaterialButtonToggleGroup_selectionRequired, false);
     setChildrenDrawingOrderEnabled(true);
+    isEnabled =
+        attributes.getBoolean(R.styleable.MaterialButtonToggleGroup_android_enabled, true);
     attributes.recycle();
 
     ViewCompat.setImportantForAccessibility(this, ViewCompat.IMPORTANT_FOR_ACCESSIBILITY_YES);
@@ -248,6 +251,9 @@ public class MaterialButtonToggleGroup extends LinearLayout {
             shapeAppearanceModel.getBottomLeftCornerSize(),
             shapeAppearanceModel.getTopRightCornerSize(),
             shapeAppearanceModel.getBottomRightCornerSize()));
+
+    // Enable children based on the MaterialButtonToggleGroup own isEnabled
+    buttonChild.setEnabled(isEnabled);
 
     ViewCompat.setAccessibilityDelegate(
         buttonChild,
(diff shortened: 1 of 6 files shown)


lib/java/com/google/android/material/button/MaterialButtonToggleGroup.java
```diff
@@ -249,6 +252,9 @@ public void addView(View child, int index, ViewGroup.LayoutParams params) {
             shapeAppearanceModel.getTopRightCornerSize(),
             shapeAppearanceModel.getBottomRightCornerSize()));
 
+    // Enable children based on the MaterialButtonToggleGroup own isEnabled
+    buttonChild.setEnabled(isEnabled);
```
pekingme, line 256: In the [spec](https://lh3.googleusercontent.com/wLirQErMH92A2qj58hHihYhhHG9tNNsOlSqMLjoLb1RwufR7sh0b3mIw-X7DoHOGXT75Rb0FMhCUKsJA7ycfq1qNpeY1NAmYVHXcV_eEcrWi=s0), we want to change individual child button state independently. This will make a previously disabled child button enabled, when the group goes enabled -> disabled -> enabled. But I'm still thinking if we should address it here or as a separate issue. Do you have any thoughts how to make it work like that?

pekingme (commented):
Thanks for the PR!

lib/java/com/google/android/material/button/MaterialButtonToggleGroup.java
BluestormDNA: I think the MaterialButtonToggle enable or disable enabling all or disabling all is pretty consistent and what one should expect. 
See the issues already open for that (I myself made this PR because i too one day disabled and enabled it expecting this too).
My thought is that Individual child state can still be handled individually. Maybe a comment with a warning should be added along the lines of  "it operates on all childs and if you need a more fine grained aproach please operate on each button individually"
A more cumbersome approach could be added to disable all and on reenable only reenable the ones that where previously enabled but it is even needed? Would that even make sense or be the thing that someone would expect from setEnable(true)?
If you want to explore that (i think it should be other PR) we could explore an overload of the method too.
something like setEnable(bool enable, bool preservePreviousState) but it opens a lot of other considerations and as I think it's complicating a thing that shouldn't be.

lib/java/com/google/android/material/button/MaterialButtonToggleGroup.java
```diff
@@ -769,6 +775,22 @@ void onButtonCheckedStateChanged(@NonNull MaterialButton button, boolean isCheck
       checkInternal(button.getId(), isChecked);
   }
 
+  /**
+   * setEnable this {@link MaterialButtonToggleGroup} and all it's {@link MaterialButton} children
```
pekingme, line 779: nit: Enables this {@link MaterialButtonToggleGroup} and all its {@link MaterialButton} children

lib/java/com/google/android/material/button/res/values/attrs.xml
```diff
@@ -112,6 +112,8 @@
     <attr name="selectionRequired"/>
     <!-- The id of the child button that should be checked by default within this button group. -->
     <attr name="checkedButton" format="reference"/>
+    <!-- Whether this MaterialButtonToggleGroup and all it's children are enabled -->
```
pekingme, line 115: nit: it's -> its

lib/java/com/google/android/material/button/MaterialButtonToggleGroup.java
pekingme: Yea, I was considering just let this in and deal with that later. This PR still resolves the linked issues. We can take a look at the state restoring later as needed.

lib/java/com/google/android/material/button/MaterialButtonToggleGroup.java
BluestormDNA: Done

lib/java/com/google/android/material/button/res/values/attrs.xml
BluestormDNA: Done

pekingme (approved):
Thanks, will pull this in for internal reviews.

Issue #2942, opened 2022-08-30
[Slider] Label breaks if it's too close to the edge of the screen


pubiqq:
<img src="https://user-images.githubusercontent.com/82187521/187322121-2cc1c0a6-cd1c-44a9-a4ac-1acb001672e6.png" />

**Material Library version:** 1.7.0-beta01
**Device:** Pixel 4 API 31

PR #3027, opened 2022-10-15
Bugfix/2869 label scroll

### LINKED ISSUE #2869: [Slider] The label is scrolled when it's in the scroll view and LABEL_VISIBLE
**Description:** 
This issue comes when the slider is placed inside ScrollView and `labelBehavior` is set as `LABEL_VISIBLE`. When the user scrolled down, the label is scrolled too.

WhatsApp Video 2022-07-27 at 1 07 47 PM

**Expected behavior:** The Expected behavior is the label is not scrolled and is still in the slider position

**Android API version:** Android 10 (API level 29) and Android 11 (API level 30) 

**Material Library version:** Tested on 1.6.0 and 1.7.0-alpha03

**Device:** Tested on Google Pixel 3A (Emulator) and Xiaomi Redmi 10S

PaulKlauser:
Fixes #2869 

See commit messages. Not sure if the commit that adds to the sample is wanted, let me know. 

Also ran into a related issue, #2942, once the label is drawn to the canvas. My solution was to remove the Tooltip padding entirely, however I could see why that might be undesirable in the case that the Tooltip is still being drawn over the content overlay. Let me know, that could also easily be parameterized based on label behavior. I do think we at least don't want that Tooltip padding if it's drawn to the canvas, since the view will already have padding set on it.

### DIFF
--- a/lib/java/com/google/android/material/slider/BaseSlider.java
+++ b/lib/java/com/google/android/material/slider/BaseSlider.java
@@ -68,6 +68,8 @@ import android.view.ViewParent;
 import android.view.accessibility.AccessibilityEvent;
 import android.view.accessibility.AccessibilityManager;
 import android.widget.SeekBar;
+import android.widget.TextView;
+
 import androidx.annotation.ColorInt;
 import androidx.annotation.ColorRes;
 import androidx.annotation.DimenRes;
@@ -759,6 +761,7 @@ abstract class BaseSlider<
       // if more thumbs are added.
       TooltipDrawable tooltipDrawable =
           TooltipDrawable.createFromAttributes(getContext(), null, 0, labelStyle);
+      tooltipDrawable.setLayoutMargin(0);
       labels.add(tooltipDrawable);
       if (ViewCompat.isAttachedToWindow(this)) {
         attachLabelToContentView(tooltipDrawable);
@@ -1726,7 +1729,7 @@ abstract class BaseSlider<
 
     // Draw labels if there is an active thumb or the labels are always visible.
     if ((activeThumbIdx != -1 || shouldAlwaysShowLabel()) && isEnabled()) {
-      ensureLabelsAdded();
+      ensureLabelsAdded(canvas);
     } else {
       ensureLabelsRemoved();
     }
@@ -2193,7 +2196,7 @@ abstract class BaseSlider<
     }
   }
 
-  private void ensureLabelsAdded() {
+  private void ensureLabelsAdded(Canvas canvas) {
     if (labelBehavior == LABEL_GONE) {
       // If the label shouldn't be drawn we can skip this.
       return;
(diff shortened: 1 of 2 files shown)


raajkumars:
Thanks for the PR. Please upload a  video recording of the issue with the fix applied.

PaulKlauser:
@raajkumars Here you go!
[device-2022-11-02-172623.webm](https://user-images.githubusercontent.com/6461398/199605696-2bfd12e6-bf44-4202-a1a7-9f13e73fe250.webm)

lib/java/com/google/android/material/slider/BaseSlider.java
```diff
@@ -759,6 +761,7 @@ private void createLabelPool() {
       // if more thumbs are added.
       TooltipDrawable tooltipDrawable =
           TooltipDrawable.createFromAttributes(getContext(), null, 0, labelStyle);
+      tooltipDrawable.setLayoutMargin(0);
```
leticiarossi, line 764: would you like to create a separate PR to fix the label breaking issue (or change this one)? for that we can change the layout margin to 2dp instead of 0dp (so it doesn't directly touch the screen edges but it also doesn't break the label), and the change can be done in the tooltip style [here](https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/tooltip/res/values/styles.xml#L37)!

lib/java/com/google/android/material/slider/BaseSlider.java
```diff
@@ -2248,13 +2251,19 @@ private void setValueForLabel(TooltipDrawable label, float value) {
...
+      Rect rect = new Rect(label.getBounds());
+      DescendantOffsetUtils.offsetDescendantRect(ViewUtils.getContentView(this), this, rect);
+      label.setBounds(rect);
+    }
 
-    ViewUtils.getContentViewOverlay(this).add(label);
+    if (labelBehavior != LABEL_VISIBLE) {
+      ViewUtils.getContentViewOverlay(this).add(label);
+    } else {
+      label.draw(canvas);
```
leticiarossi, line 2265: so for this, the problem we have is that is if we don't draw on the view overlay, we can not draw out of the view bounds, unless we set clipChildren to false on the container (which we can't do from this class). Ideally the solution would possibly be to hide the label while the container is being scrolled

PaulKlauser: Yea that's a good observation. I think drawing the label within the view makes more sense in this case. I don't think users would want to draw the label outside of the view if it's persistent, since in that case it would likely be covering up an adjacent view.

drchen:
Hi, thank you so much for contributing the PR!

After discussing internally, we feel the best fix we can offer here is hiding the tooltip entirely during scrolling. (And show it again after scrolling.)

I'll close this PR as a no-go. Any suggestions / PRs will be very welcome!

PR #3031, opened 2022-10-18
[Buttons] Fixes text measurement to allow Spanned text

wesalvaro:
Changes to use `Layout`'s text measurements so that we can leverage the existing measurements that take into account `Spanned` text (as opposed to `TextPaint` measures that does not handles `Span`s).


drchen:
It's merged in https://github.com/material-components/material-components-android/commit/3b333ee8343dc0bde20d110a39ccc4654977eb23.

PR #3042, opened 2022-10-21
[MaterialDatePicker] Change hint formatting from m/d/yy → mm/dd/yyyy

### LINKED ISSUE #3039: [Material Date Picker] When using INPUT_MODE_TEXT, the date shows as "m/d/yy", but should show up as "mm/dd/yy"
**Description:** 
Hey all!  Thanks for reading my issue <3 

image

I'm in the US, so my local is set to US.  It's very odd for users to have to enter in their date in this format. Traditionally it's mm/dd/yyyy.

#1180 (point 2) was a dupe of #774 and both @dsn5ft and @drchen closed these as completed but I still don't see relevant commits or anyway to change my format to mm/dd/yyyy.

**Expected behavior:** 

(copied from material.io website)
image

**Minimal sample app repro:** 
This is reproducible on the [material components app](https://github.com/material-components/material-components-android.git)

**Android API version:** Android API version here

**Material Library version:** This occurs on both Material 2 and Material3 components.

**Device:** All devices

nachtien:
Closes #3039

The hint now shows as:
image

Also, the error suggestion is adjusted accordingly
image

### DIFF
--- a/lib/java/com/google/android/material/datepicker/UtcDates.java
+++ b/lib/java/com/google/android/material/datepicker/UtcDates.java
@@ -170,6 +170,24 @@ class UtcDates {
       formatHint = formatHint.replace("y", "yyyy");
     }
 
+    // Format year to always be displayed as 4 chars when 2 chars are used in localized pattern.
+    // Example: (fr-FR) dd/MM/y -> dd/MM/yyyy
+    if (formatHint.replaceAll("[^y]", "").length() == 2) {
+      formatHint = formatHint.replace("yy", "yyyy");
+    }
+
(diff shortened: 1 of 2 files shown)


nachtien:
The user can still enter in a m/d/yy format. I chose to keep it this way as we display the calculated date at the top and thus won't cause any harm forcing a user to enter a format as mm/dd/yyyy.

hunterstich:
Thanks for the PR @nachtien,

I'm not sure we want to force this change by directly updating the pattern. I'm looking at this internally and leaning towards something similar to what chanakin mentioned about [allowing custom `SimpleDateFormats`](https://github.com/material-components/material-components-android/issues/3039#issuecomment-1286154933)

PR #3053, opened 2022-10-25
[Internal] Deprecate public MathUtils

pubiqq:
Utility methods for internal use should not be public (by analogy with [AnimationUtils](https://github.com/material-components/material-components-android/blob/ba70cd72d92501d556c3a4dbf01b1446217e4627/lib/java/com/google/android/material/animation/AnimationUtils.java), [ColorUtils](https://github.com/material-components/material-components-android/blob/ba70cd72d92501d556c3a4dbf01b1446217e4627/lib/java/com/google/android/material/color/utilities/ColorUtils.java), [ContextUtils](https://github.com/material-components/material-components-android/blob/ba70cd72d92501d556c3a4dbf01b1446217e4627/lib/java/com/google/android/material/internal/ContextUtils.java) and so on).
In order not to break binary compatibility, the public `MathUtils` class is preserved, but marked as `@Deprecated`.

### DIFF
--- a/lib/java/com/google/android/material/circularreveal/CircularRevealHelper.java
+++ b/lib/java/com/google/android/material/circularreveal/CircularRevealHelper.java
@@ -15,7 +15,7 @@
  */
 package com.google.android.material.circularreveal;
 
-import static com.google.android.material.math.MathUtils.DEFAULT_EPSILON;
+import static com.google.android.material.internal.MathUtils.DEFAULT_EPSILON;
 
 import android.graphics.Bitmap;
 import android.graphics.Bitmap.Config;
@@ -37,7 +37,7 @@ import androidx.annotation.IntDef;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import com.google.android.material.circularreveal.CircularRevealWidget.RevealInfo;
-import com.google.android.material.math.MathUtils;
+import com.google.android.material.internal.MathUtils;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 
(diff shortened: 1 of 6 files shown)


DanielJorn (approved):

raajkumars:
Thanks for the PR. But can you please update the description section to provide some explanation as to why this change is required.

pubiqq:
Ok, done.

drchen:
Hi thanks for the PR but unfortunately I don't think we can make this change since it's an API breakage. 

I agree MathUtils is better to be private but given now it's public in released versions, we cannot hide it.

pubiqq:
I was sure you had removed methods from `MathUtils` before, but it turns out this was before the first stable MDC was released 😅.
Could you reopen my PR so that I can change my proposal a bit?

drchen:
Apparently.... I cannot reopen a closed PR either....  (That's a surprise!)
Would you mind to create a new PR? Sorry for the inconvenience.

(And yes, basically we can't change any public APIs after **beta** versions are released.)

pubiqq:
Could you try again?

pubiqq:
New changes are not picked up when the PR is closed, so I reverted the branch to its original state.

pubiqq:
Ok, what about this proposal?

pubiqq:
bump

pubiqq:
bump

pubiqq:
bump

pubiqq:
bump

pubiqq:
bump

pubiqq:
bump

drchen:
Hey, still I don't see a strong reason to duplicate the code just to deprecate the public API...

If we are worried about exposing more unneeded API in the future, the common practice will be delegating the logic from the deprecated API to the internal API. But it really seems like an over-kill to me in the MathUtils case...

pubiqq:
This change is intended so that you can make any changes to `MathUtils` without fear that they will break something in the public API.

I just didn't want to mix deprecation of the public class and changes to the restricted class one in the same PR, so I decided to request approval for deprecation first.

drchen:
@dsn5ft Dan, can you help make a decision here?

dsn5ft:
This doesn't really feel worth it to me. At this point it's not difficult for us to maintain the existing MathUtils, since we basically never change it. So I think duplicating / deprecating is more confusing than just leaving it as is.

Also we do have some other public util classes.

dsn5ft:
Will close the PR for now but feel free to reopen or add more comments.

Issue #3058, opened 2022-11-02
[Menu][BottomAppBar][BottomNavigationView] background color is slightly different


manabu-nakamura:
**Description:** The background color of menu (and action bar) is slightly different from the background color of bottom app bar (and navigation bar) while the specs seem to be the same.
(Menu)
https://m3.material.io/components/menus/specs:
```
Element      | Design attribute   | Role         | Token
5. Container | Color              | Surface      | md.sys.color.surface
             | Surface tint color | Surface tint | md.sys.color.surface-tint
             | Elevation          | Level 2      | md.sys.elevation.level2
```
![Screenshot_20221102_140603](https://user-images.githubusercontent.com/71050561/199409281-9403034d-09a6-463c-8ba0-e4c6d1a3c7eb.png)
The background color of menu is **#f6ede8**.
(Action Bar)
![Screenshot_20221114_112020](https://user-images.githubusercontent.com/71050561/201563690-6cb5cb8a-b93b-4b7a-be6e-a87c4af44c6e.png)
The background color of action bar is **#f6ede8**.
(Bottom App Bar, Navigation Bar)
https://m3.material.io/components/bottom-app-bar/specs:
https://m3.material.io/components/navigation-bar/specs:
```
Element      | Design attribute | Role         | Token or value
1. Container | Color            | Surface      | md.sys.color.surface
             | Elevation        | Level 2      | md.sys.elevation.level2
             | Surface tint     | Surface tint | md.sys.color.surface-tint
```
![Screenshot_20221102_140841](https://user-images.githubusercontent.com/71050561/199409228-6204eb1e-6006-490b-a890-486a62639086.png)
![Screenshot_20221114_114246](https://user-images.githubusercontent.com/71050561/201564362-128fd62b-8ae6-468f-a498-8fc2abfab4be.png)
The background color of bottom app bar (and navigation bar) is **#f5ece7**.
(Top App Bar)
https://github.com/material-components/material-components-android/issues/2664

**Expected behavior:** The background color of menu (and action bar) is the same as the background color of bottom app bar (and navigation bar).

**Source code:**
https://github.com/material-components/material-components-android/blob/master/catalog/java/io/material/catalog/menu/MenuMainDemoFragment.java#L89
https://github.com/material-components/material-components-android/blob/master/catalog/java/io/material/catalog/menu/res/menu/popup_menu.xml
[...]

PR #3062, opened 2022-11-04
[ElevationOverlayProvider] overlayAlphaFraction should be rounded

manabu-nakamura:
#3058


drchen (approved):
The change itself looks good to me but I'm not sure if we are able to land it due to it may introduce a lot of internal test breakages. Will update here if we decide it's not worth to merge it now given we are going to deprecate this very soon.

manabu-nakamura:
Please concentrate on developing the new mechanism.

Issue #3095, opened 2022-11-21
[TabLayout] Incorrect tab textAppearance when scrolling ViewPager


pubiqq:
To reproduce the bug, set different values for `tabTextAppearance` and `tabSelectedTextAppearance`:

https://user-images.githubusercontent.com/82187521/203131257-0e1d83fc-3a26-42de-9447-187f6f8bf6e4.mp4

**Material Library version:** 1.8.0-alpha02

PR #3096, opened 2022-11-21
[TabLayout] Fix tab selection state check

pubiqq:
Fixes https://github.com/material-components/material-components-android/issues/3095


drchen (approved):
Good catch!! Thanks for the fix!

drchen:
Hi, so the PR actually doesn't work - not the PR's problem but the issue of the selected state management of Tabs. 

@imhappi is working on this to solve the Tab issue. Let me close the PR for now.

pubiqq:
In fairness, the PR worked with `1.8.0-alpha02`, but stopped working with the new TabLayout changes. However, these new changes already resolve https://github.com/material-components/material-components-android/issues/3095, so this fix is really no longer needed.

PR #3098, opened 2022-11-22
[TabLayout] Fix confusing documentation

pubiqq:


lib/java/com/google/android/material/tabs/TabLayout.java
```diff
@@ -719,8 +719,6 @@ public void setSelectedTabIndicatorHeight(int height) {
    * Set the scroll position of the tabs. This is useful for when the tabs are being displayed as
    * part of a scrolling container such as {@link androidx.viewpager.widget.ViewPager}.
    *
-   * <p>Calling this method does not update the selected tab, it is only used for drawing purposes.
```
drchen, line 722: I think we should still keep this part:

"<p>Note that calling this method does not update the selected tab."

lib/java/com/google/android/material/tabs/TabLayout.java
```diff
@@ -734,8 +732,6 @@ public void setScrollPosition(int position, float positionOffset, boolean update
...
```
drchen, line 737: ditto

lib/java/com/google/android/material/tabs/TabLayout.java
pubiqq: >"Note that calling this method does not update the selected tab."

But [it does](https://github.com/material-components/material-components-android/blob/b7cfffcdd9d57bc6e9e0564dc0f3a10e84e7efec/lib/java/com/google/android/material/tabs/TabLayout.java#L1908).

drchen: I guess it means the tab selection won't be changed? Probably rephrase the sentence then?

pubiqq: Maybe another time. I'll put this line back then.

drchen: Hey sorry I didn't take a close look at the code but now I get what you mean. 

The method is actually updating the selected tab and yeah the javadoc is definitely improper. 
Sorry that can you still remove those two lines? And I'll send it for our internal review.

Thanks for your contribution to the codebase, as always. : )

lib/java/com/google/android/material/tabs/TabLayout.java
```diff
@@ -716,36 +716,40 @@ public void setSelectedTabIndicatorHeight(int height) {
...
+   * @param position Position of the tab to scroll.
    * @param positionOffset Value from [0, 1) indicating the offset from {@code position}.
-   * @param updateSelectedText Whether to update the text's selected state.
+   * @param updateTabSelection Whether to draw the tab at the given position + positionOffset
+   *     as selected.
    * @see #setScrollPosition(int, float, boolean, boolean)
    */
-  public void setScrollPosition(int position, float positionOffset, boolean updateSelectedText) {
-    setScrollPosition(position, positionOffset, updateSelectedText, true);
+  public void setScrollPosition(int position, float positionOffset, boolean updateTabSelection) {
```
imhappi, line 731: Can you revert this change (and the comment) and all of the changes for `updateTabSelection` back to `updateSelectedText`?

I don't think `updateTabSelection` is a good name,  since the `updateSelectedTab` parameter only does determine whether or not to update the selected state of the textview of the tab and text appearance. `updateTabSelection` implies that the tab is being selected, which `setScrollPosition` does not do

imhappi (changes requested):
LGTM, but I think the `updateTabSelection` changes should be reverted

lib/java/com/google/android/material/tabs/TabLayout.java
pubiqq: Ok, done.

P.S. Well, technically, the comment is not entirely improper, it's just ambiguous. 
I could rephrase it like "Note that calling this method with `updateTabSelection = true` does not actually select a tab at the requested position, but only _draws it as selected_", but I'm not sure this comment is needed here at all.

lib/java/com/google/android/material/tabs/TabLayout.java
pubiqq: This parameter determines not just whether to update the text view, but whether to update _the whole tab_ (two tabs, to be more precise), including text view, icon, background, and custom view (if set). Perhaps `updateTabSelection` is not the best name, but it's definitely better and more correct than `updateSelectedText`, imo.

imhappi: Hmm I do agree that it is a better name, but still misleading since a selected tab means something already in TabLayout.

Maybe `updateTabAppearance` or something to indicate? It's also confusing since it doesn't include the indicator haha but maybe that's just something we clarify in the comments

lib/java/com/google/android/material/tabs/TabLayout.java
pubiqq: I decided not to remove the comment, but just changed it a little. It still needs some editing, but I believe it's better to leave this task to your reviewers.

lib/java/com/google/android/material/tabs/TabLayout.java
pubiqq: `updateTabAppearance` is an acceptable name, but it's too vague I think. How about `updateSelectedTabView` + some comments? (see my last change)

imhappi (approved):
LGTM! Thanks for the changes

Issue #3122, opened 2022-12-01
[Catalog][Preferences] edge-to-edge preference does not apply to preferences bottom sheet


manabu-nakamura:
**Is your feature request related to a problem? Please describe.**
The edge-to-edge preference does not apply to the preferences bottom sheet.
![Screenshot_20221201_104932](https://user-images.githubusercontent.com/71050561/204947945-4c69b9c8-5f2c-4a6b-9ebd-625678df4231.png)

**Describe the solution you'd like**
The edge-to-edge preference applies to the preferences bottom sheet.
![Screenshot_20221201_105151](https://user-images.githubusercontent.com/71050561/204947970-ec6cccf9-5450-4a5b-bb65-8fa78b80f4b5.png)

PR #3123, opened 2022-12-01
edge-to-edge preference applies to preferences bottom sheet

manabu-nakamura:
#3122


leticiarossi (approved):
Thank you for the PR!

Issue #3133, opened 2022-12-08
[Catalog] Applied edge-to-edge to modal side sheet


manabu-nakamura:
**Is your feature request related to a problem? Please describe.**
The edge-to-edge preference does not apply to the modal side sheet.
![Screenshot_20221208_122759](https://user-images.githubusercontent.com/71050561/206351126-480a718b-f202-48b3-a3f6-13dcb82ce6c4.png)

**Describe the solution you'd like**
The edge-to-edge preference applies to the modal side sheet.
![Screenshot_20221208_124002](https://user-images.githubusercontent.com/71050561/206351176-287ef269-9fdb-4148-9372-a30cf3eae574.png)

afohrman:
Closed in https://github.com/material-components/material-components-android/pull/3155.

PR #3134, opened 2022-12-08
Applied edge-to-edge to modal side sheet

manabu-nakamura:
#3133


imhappi:
Hi! Could you confirm if this change fixes the issue? In my testing, it doesn't look like it does..

manabu-nakamura:
Without this change, the system navigation bar (the bottom of the screen) would be colored.
https://user-images.githubusercontent.com/71050561/208216620-28ea8fe1-4702-49d7-89ba-92384d981a68.mp4
Screenshot_20221208_122759

PR #3145, opened 2022-12-18
[MaterialDatePicker] Adding an optional 'Clear' button

### LINKED ISSUE #1332: [Feature Request] Add Neutral Button to MaterialDatePicker
Use case: There are times when one wants to clear a selected date (i.e. I have a task with a due date, I want to use the MaterialDatePicker to display the current date, pick a new date OR clear the selected date.

The first two are currently supported, the 3rd requires either the ability to extend the class and override the onCreateView (not possible because the class is final) or adding support for a Neutral button.

DvTonder:
This change adds an optional 'Clear' button that can be used to remove the selected date with the option to calling a clear listener and dismissing the dialog when clicked.

closes #1332 

Note: This is the same code as previously submitted as a pull request but it has been rebased against the latest codebase as of 18 December 2022

### DIFF
--- a/jitpack.yml
+++ b/jitpack.yml
@@ -1,0 +1,2 @@
+jdk:
+  - openjdk11
--- a/lib/java/com/google/android/material/datepicker/DateSelector.java
+++ b/lib/java/com/google/android/material/datepicker/DateSelector.java
@@ -67,6 +67,11 @@ public interface DateSelector<S> extends Parcelable {
   void setSelection(@NonNull S selection);
 
   /**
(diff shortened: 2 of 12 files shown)


DvTonder:
I have updated the full screen date picker to support the clear function

drchen (dismissed):

DvTonder:
It was a bit messy with a force push but I managed to rebase to the HEAD.  Should be good now

drchen:
It seems working now. Thank you! : )

Issue #3194, opened 2023-01-27
[Snackbar] `isAnchorViewLayoutListenerEnabled` is not working if anchor changed immediately


aleksandra-krzemien:
**Description:** 
As mentioned [here](https://github.com/material-components/material-components-android/issues/2744#issuecomment-1301495088), fix for the `isAnchorViewLayoutListenerEnabled` issue with infinite loop, caused a different error which causes the snackbar not be shown above the anchor view if the anchor position is changed right after screen is loaded.

**Expected behavior:** 
If `isAnchorViewLayoutListenerEnabled` is set to `true` it should always update snackbar position, so it's shown above the anchor view.

**Source code:** 
`recalculateAndUpdateMargins` does not trigger `updateMargins` if `newBottomMarginAnchorView` has already the same value, but because of the missing parent in the first run, the margins never get updated.

Combination of:
```Java
private void recalculateAndUpdateMargins() {
  int newBottomMarginAnchorView = calculateBottomMarginForAnchorView();
  if (newBottomMarginAnchorView == extraBottomMarginAnchorView) {
    return;
  }
  extraBottomMarginAnchorView = newBottomMarginAnchorView;
  updateMargins();
}

private void updateMargins() {
    LayoutParams layoutParams = view.getLayoutParams();
    if (!(layoutParams instanceof MarginLayoutParams) || view.originalMargins == null) {
      Log.w(TAG, "Unable to update margins because layout params are not MarginLayoutParams");
      return;
    }
    if (view.getParent() == null) {
      // Parent will set layout params to view again. Wait for addView() is done to update layout
      // params, in case we save the already updated margins as the original margins.
      return;
    }
...
}

final void showView() {
    ...
      this.view.addToTargetParent(targetParent);
      recalculateAndUpdateMargins();
    ...
}
```

**Minimal sample app repro:** 
https://github.com/aleksandra-krzemien/nested-nav-graphs-example/tree/krzemien/snackbar-anchor-listener
This is a multipurpose repo, but on this branch you can run the app, and tap on `to first fragment` button. The snackbar should be shown above the button, and it's not. All the code needed for repro is in `FirstFragment.snackbarExample()`.
<img src="https://user-images.githubusercontent.com/101054541/215221880-2d53dc64-aeae-41d1-9dc9-3d6cbb133570.gif" width="250"/>

**Android API version:** 28

**Material Library version:** 1.8.0

**Device:** emulator

PR #3211, opened 2023-02-03
[Snackbar] Fix the snackbar position when the anchor view layout listener is enabled

pubiqq:
Fixes https://github.com/material-components/material-components-android/issues/3194

### DIFF
--- a/lib/java/com/google/android/material/snackbar/BaseTransientBottomBar.java
+++ b/lib/java/com/google/android/material/snackbar/BaseTransientBottomBar.java
@@ -324,6 +324,7 @@ public abstract class BaseTransientBottomBar<B extends BaseTransientBottomBar<B>
   private int extraRightMarginWindowInset;
   private int extraBottomMarginGestureInset;
   private int extraBottomMarginAnchorView;
+  private boolean extraBottomMarginAnchorViewChanged = true;
 
   private boolean pendingShowingView;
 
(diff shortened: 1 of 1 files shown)


drchen:
Hi, thanks for the PR. I actually have a fix using a different approach on its way. : )

Let's pause the review here and see if my fix works.

pubiqq:
Okay, no problem. 

P. S. If your solution is similar to this:

<details>
  <summary>Code</summary>
  
```java
private boolean extraBottomMarginAnchorViewChanged;

// ...

private void recalculateAndUpdateMargins() {
  int newBottomMarginAnchorView = calculateBottomMarginForAnchorView();
  if (newBottomMarginAnchorView != extraBottomMarginAnchorView) {
    extraBottomMarginAnchorViewChanged = true;
    extraBottomMarginAnchorView = newBottomMarginAnchorView;
  }

  if (extraBottomMarginAnchorViewChanged) {
    updateMargins();
  }
}

// ...

private void updateMargins() {
  // ...
  
  marginParams.bottomMargin = view.originalMargins.bottom + extraBottomMargin;
  marginParams.leftMargin = view.originalMargins.left + extraLeftMarginWindowInset;
  marginParams.rightMargin = view.originalMargins.right + extraRightMarginWindowInset;
  marginParams.topMargin = view.originalMargins.top;
  
  view.requestLayout();

  extraBottomMarginAnchorViewChanged = false;
  
  // ...
}
```
  
</details>

then I would still recommend taking mine.

drchen:
Not really. But you reminded me a very good point regarding the originalMargins can also be changed. : )

drchen:
Ah - actually. After re-reading it, I think your fix is better than mine. : )

Let me pull your PR for internal review instead.

drchen (approved):

pubiqq:
Oh, I almost forgot: this change slightly aggravates the existing problem with `gestureInsetBottomIgnored`, but I intentionally did nothing about it, because in the next PR I'm going to eliminate the problem completely.
However, if you really need to, I can also add a regression fix (only regression, not the whole problem) to this PR.

drchen:
Ah. Yes please, a quick workaround for now will be good. Just FYI any regression may cause a lot of internal breakages that make the change hard to land itself. : )   (I do see some but haven't got time to check yet. So if you already have some idea that will be very helpful.)

pubiqq:
I believe this is not the case. However, ok, I can fix the issue in another way.

pubiqq:
I believe this is not the case. But ok, I can fix the issue in another way.

PR #3221, opened 2023-02-07
[Internal] Add compat method to get the screen size

pubiqq:


drchen (approved):
That's awesome! Thanks for the PR!

drchen:
No problem. The only difference I saw is adding a private constructor to the WindowUtils class. Is that all the change you made? (I already added that during the internal review process. : ) )

pubiqq:
Yep. If there's anything else you need, don't hesitate to ask me.

Issue #1119, opened 2020-03-17
[Toolbar] Navigation icon color red on Android 5.1.1


wbervoets:
**Description:** 

`<com.google.android.material.appbar.AppBarLayout xmlns:android="http://schemas.android.com/apk/res/android"
    android:id="@+id/appBarLayout"
    android:layout_width="match_parent"
    android:layout_height="wrap_content">
    <com.google.android.material.appbar.MaterialToolbar
        android:id="@+id/toolBar"
        android:layout_width="match_parent"
        android:layout_height="wrap_content"
        style="@style/Widget.MaterialComponents.Toolbar.Surface" />

</com.google.android.material.appbar.AppBarLayout>` 

I'm using the style @style/Widget.MaterialComponents.Toolbar.Surface 
I expect it to use my colorSurface and colorOnSurface attributes from my theme for the background and the tinting of the hamburger navigation icon

My colorSurface color is white / my colorOnSurface color is black in Day (light) mode

This issue is that on Android 5.1.1 devices the hamburer icon is tinted red. The background color of the toolbar is white and the title of the toolbar is also black as expected. The only red color I'm using is for errors (colorError). I do not understand where this red color comes from

My app theme parent is Theme.MaterialComponents.DayNight.NoActionBar 

I do not see this bug when using Widget.MaterialComponents.Toolbar.PrimarySurface on Android 5.1.1 but then it uses my primary color / primary on color which in this case I do not want to use

**Expected behavior:** 

I would expect the navigation icon to be black too.

**Android API version:** 

API 21

**Material Library version:** Material Android Library version you are using here (e.g., 1.1.0-alpha07)

1.1.0 or any 1.2 alpha

**Device:**

LGE Nexus 4 running Android 5.1.1

aL05flas:
Hello! I also encountered this problem and after three days of research I found out that the dark theme uses the resource to colorControlNormal across ColorStateList with alpha and color+theme attribute, what work since Android M+, otherwise there will be an error and red color. The material design library currently does not represent the correct implementation of resources for Android < M.

killvetrov:
I've also done some research on this issue with repro project in #1820.

ArcherEmiya05:
Is this fix in Material 3? I tried to provide `colorOnSurfaceVariant` for action menu but it seems not working. It works when using `colorControlNormal`. Thanks

drchen:
I think this is fixed with Material 3. Since we deprecated Material 2, let me close this issue as obsolete.

Issue #1820, opened 2020-10-24
[Theming] On SDK <23 DrawerArrowDrawable incorrectly resolves color under ThemeOverlay.MaterialComponents.Toolbar.Surface


killvetrov:
**Description:** I noticed this issue by using Jetpack Navigation. `NavigationUI.setupWithNavController(toolbar, navController)` calls `Toolbar.setNavigationIcon(new DrawerArrowDrawable())`. This DrawerArrowDrawable sets its color by style attribute `color`. It works correctly in majority of cases on all API levels, but I encountered this issue when using Toolbar with @style/Widget.MaterialComponents.Toolbar.Surface, which sets color to `@color/material_on_surface_emphasis_medium`. For some reason when DrawerArrowDrawable gets this color on SDK <23, it resolves to red: 0.6 alpha is respected, but colorOnSurface is not.

![Screenshot_1603540384](https://user-images.githubusercontent.com/7795552/97081080-be1c3080-1608-11eb-8c1f-179da71ba77e.png)

**Expected behavior:** DrawerArrowDrawable resolves color correctly (same as ?attr/homeAsUpIndicator) on all API levels.

**Source code:** Relevant code snippets below.
 
```java
// from androidx.appcompat.graphics.drawable.DrawerArrowDrawable.java

final TypedArray a = context.getTheme().obtainStyledAttributes(null,
                R.styleable.DrawerArrowToggle, R.attr.drawerArrowStyle,
                R.style.Base_Widget_AppCompat_DrawerArrowToggle);
setColor(a.getColor(R.styleable.DrawerArrowToggle_color, 0));
```
```xml
<!-- from XML of AppCompat and Material Components -->

<style name="Widget.AppCompat.DrawerArrowToggle" parent="Base.Widget.AppCompat.DrawerArrowToggle">
    <item name="color">?attr/colorControlNormal</item>
</style>

<style name="Widget.MaterialComponents.Toolbar.Surface">
    ...
    item name="android:theme">@style/ThemeOverlay.MaterialComponents.Toolbar.Surface</item>
</style>

<style name="ThemeOverlay.MaterialComponents.Toolbar.Surface" parent="">
    <item name="colorControlNormal">@color/material_on_surface_emphasis_medium</item>
    ...
</style>

<!-- material_on_surface_emphasis_medium.xml -->
<selector xmlns:android="http://schemas.android.com/apk/res/android">
[...]

PR #3267, opened 2023-02-24
[AppBar] Add theme overlays supported on API < 23

pubiqq:
Resolves https://github.com/material-components/material-components-android/issues/1119 and https://github.com/material-components/material-components-android/issues/1820.

### DIFF
--- a/lib/java/com/google/android/material/appbar/res/values-v23/styles.xml
+++ b/lib/java/com/google/android/material/appbar/res/values-v23/styles.xml
@@ -15,6 +15,22 @@
   ~ limitations under the License.
 -->
 <resources>
+  <style name="ThemeOverlay.MaterialComponents.Toolbar.Surface" parent="Base.ThemeOverlay.MaterialComponents.Toolbar.Surface">
+    <item name="colorControlNormal">@color/material_on_surface_emphasis_medium</item>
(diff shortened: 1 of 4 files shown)


pubiqq:
bump

pubiqq:
bump

lib/java/com/google/android/material/appbar/res/values-v23/styles.xml
```diff
@@ -15,6 +15,29 @@
   ~ limitations under the License.
 -->
 <resources>
+  <style name="ThemeOverlay.MaterialComponents.Toolbar.Surface" parent="">
```
drchen, line 18: Let's follow the standard Base.ThemeOverlay -> ThemeOverlay pattern for these theme overlays to avoid duplicate attributes.

drchen (approved):

drchen:
HI, after doing some tests I found this actually breaks the M2 style on pre-M devices.

In the happy flow we are using compat way to load CSLs so the alpha value will be applied correctly. 
The change actually causes worse breakages. 

I agree it's not ideal because we cannot control how non-Material components will load the color (like `?attr/colorControlNormal`) and this may cause incorrect colors being shown.

However, since we stopped supporting M2 and a simple solution like this will cause worse breakages, I'll close this PR as a no-go.

PR #2978, opened 2022-09-19
[Badge] Add text support

pubiqq:
Resolves https://github.com/material-components/material-components-android/issues/1217

PR #3321, opened 2023-03-30
[Badge] Fix badge issues

pubiqq:
This PR contains changes that are part of [another PR](https://github.com/material-components/material-components-android/pull/2978), but were lost when merged into the main branch.

**Changes:**
- Don't redraw the badge unless its content has changed.
- Don't ignore `maxCharacterCount` if the badge text is set.
- Don't distort `contentDescriptionForText` state when copying.
- `BadgeDrawable#clearNumber()` should clear the number from the state, even if the drawable also contains text.
- The badge should frame any text, even single-digit numbers.
- Fix badge shape update when only text is set.
- Fix inaccurate comments.
- Fix table formatting in BadgeDrawable.md.

### DIFF
--- a/docs/components/BadgeDrawable.md
+++ b/docs/components/BadgeDrawable.md
@@ -91,12 +91,12 @@ center, use `setHorizontalOffset(int)` or `setVerticalOffset(int)`
 ### `BadgeDrawable` Attributes
 
 | Feature               | Relevant attributes                                                                                                                                      |
-|-----------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------|
-| Color                 | `app:backgroundColor` <br> `app:badgeTextColor`                                                                                                                                     |
-| Width                 | `app:badgeWidth` <br> `app:badgeWithTextWidth`                                                                                                                                 |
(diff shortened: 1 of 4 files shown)


pubiqq:
P. S. I hope I won't have to wait another six months 🙃.

imhappi:
Thanks for the PR! Will take a look

As a note, we discussed with design and they do not want to truncate string text; will be removing that from the PR, but will get everything else in

pubiqq:
>As a note, we discussed with design and they do not want to truncate string text; will be removing that from the PR, but will get everything else in

No no no, `maxCharacterCount` _must_ truncate a text. The property literally sets the "[maximum number of characters a badge supports displaying by default](https://github.com/material-components/material-components-android/blob/a0d0b53472751ffafeacd829d5433afde7546db9/lib/java/com/google/android/material/badge/BadgeState.java#L64-L65)", it's even reflected in its name. 

If your design team wants to truncate only numbers, we most likely need to add a new property, something like `maxNumber` or `maxDigitCount`. If you like, I can add such a property in this PR, or in a separate one.

imhappi:
Ah you are correct; I've made the necessary changes to add new property to truncate only-numbers, and will be updating the m3 style.
Hopefully this will be submitted soon!

Issue #3329, opened 2023-04-03
[Carousel] Item masking is broken on API 21


pubiqq:
https://user-images.githubusercontent.com/82187521/229579371-f13fb845-dc18-4ec0-b8f9-b90c11aea028.mp4

<br>

**Material Library version:** 1.10.0-alpha01

**Device:** API 21 Emulator

PR #3330, opened 2023-04-03
[Carousel] Fix item masking for API 21

pubiqq:
Fixes https://github.com/material-components/material-components-android/issues/3329


hunterstich:
Thanks, @pubiqq!

Issue #3322, opened 2023-03-31
[SideSheet] - 1.10.0-alpha01 - NPE: shapeAppearanceModel not initialized before layout (with workaround)


fat-tire:
Runtime NPE when trying to layout a SideSheet (1.10-alpha01):

```
java.lang.NullPointerException: Attempt to invoke virtual method 'com.google.android.material.shape.ShapeAppearanceModel$Builder com.google.android.material.shape.ShapeAppearanceModel.toBuilder()' on a null object reference
      	at com.google.android.material.sidesheet.SideSheetBehavior.setSheetEdge(SideSheetBehavior.java:163)
     	at com.google.android.material.sidesheet.SideSheetBehavior.setSheetEdge(SideSheetBehavior.java:158)
  	at com.google.android.material.sidesheet.SideSheetBehavior.onLayoutChild(SideSheetBehavior.java:352)
     	at androidx.coordinatorlayout.widget.CoordinatorLayout.onLayout(CoordinatorLayout.java:918)
```

Line 163 is where it crashes:
```kotlin
      ShapeAppearanceModel.Builder builder = shapeAppearanceModel.toBuilder();
```

Turns out `shapeApparanceModel` is null.  Hence the crash.

`shapeApparanceModel` should have been set on line 134 earlier:
```kotlin
    if (a.hasValue(R.styleable.SideSheetBehavior_Layout_shapeAppearance)) {
      this.shapeAppearanceModel =
          ShapeAppearanceModel.builder(context, attrs, 0, DEF_STYLE_RES).build();
    }
```

But the assignment is skipped as the test w/ `a` fails.  Let's look at where `a` is set:

Line 128:
```kotlin
    TypedArray a = context.obtainStyledAttributes(attrs, R.styleable.SideSheetBehavior_Layout);
```

We need to make sure `a.hasValue(R.styleable.SideSheetBehavior_Layout_shapeAppearance`).

Here's my workaround to set `a` so that it won't fail the test on 134.  In `app/src/main/res/values/themes.xml` (or `style.xml`), add something like this to the app's main `<style>`:

```xml
        <item name="shapeAppearance">@style/Widget.Material3.SideSheet</item>
```

`Widget.Material3.SideSheet.Modal` also works.  For example:
```xml
<?xml version="1.0" encoding="utf-8"?>
<resources xmlns:tools="http://schemas.android.com/tools">
    <!-- the theme applied to the application or activity -->
[...]

PR #3332, opened 2023-04-04
[SideSheet] Fix NPE when shapeAppearance is not set

pubiqq:
Fixes https://github.com/material-components/material-components-android/issues/3322


pekingme:
We have a similar fix rolling out. Closing this PR for duplication.

Issue #3335, opened 2023-04-04
[Tabs] Incorrect tab indicator size


pubiqq:
<table>
  <tr>
    <th>Actual behavior<br>(API 17 Emulator)</th>
    <th>Expected behavior<br>(API 19 Emulator)</th>
   </tr> 
   <tr>
    <td><img src="https://user-images.githubusercontent.com/82187521/229940360-167e71e8-a4e9-4abe-bb0b-05f2959b7dcd.png" width="324px"></td>
    <td><img src="https://user-images.githubusercontent.com/82187521/229940433-4fac7a37-5373-4df4-94c4-9688563c4c61.png" width="324px"></td>
    </tr>
</table>

**Material Library version:** 1.10.0-alpha01

**Device:** API 17 Emulator

PR #3336, opened 2023-04-04
[Tabs] Fix tab indicator size

pubiqq:
Fixes https://github.com/material-components/material-components-android/issues/3335

### DIFF
--- a/lib/java/com/google/android/material/tabs/TabLayout.java
+++ b/lib/java/com/google/android/material/tabs/TabLayout.java
@@ -3132,10 +3132,12 @@ public class TabLayout extends HorizontalScrollView {
       int left = 0;
       int right = 0;
 
-      for (View view : new View[] {textView, iconView, customView}) {
-        if (view != null && view.getVisibility() == View.VISIBLE) {
-          left = initialized ? Math.min(left, view.getLeft()) : view.getLeft();
-          right = initialized ? Math.max(right, view.getRight()) : view.getRight();
+      for (int i = 0; i < getChildCount(); i++) {
+        final View child = getChildAt(i);
+
+        if (child.getVisibility() != View.GONE) {
+          left = initialized ? Math.min(left, child.getLeft()) : child.getLeft();
+          right = initialized ? Math.max(right, child.getRight()) : child.getRight();
           initialized = true;
         }
       }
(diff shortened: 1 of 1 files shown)


drchen:
Thanks for the PR! Can you explain what's the root cause of the issue and how the PR fixes it?

pubiqq:
`textView` and `iconView` are not direct children of `TabView` on API 17 and below (see [inflateAndAddDefaultTextView()](https://github.com/material-components/material-components-android/blob/a0d0b53472751ffafeacd829d5433afde7546db9/lib/java/com/google/android/material/tabs/TabLayout.java#L2871) and [inflateAndAddDefaultIconView()](https://github.com/material-components/material-components-android/blob/a0d0b53472751ffafeacd829d5433afde7546db9/lib/java/com/google/android/material/tabs/TabLayout.java#L2858)), so `getLeft()`/`getRight()`/`getTop()`/`getBottom()` methods cannot be used to determine the position of these views relative to `TabView`. This PR fixes the issue by replacing incorrect views with direct `TabView` children.

Also, in the PR, `view.getVisibility() == View.VISIBLE` is replaced with `view.getVisibility() != View.GONE`, because invisible elements also take up space, and therefore should be taken into account when determining the content size.

pubiqq:
I don't think it's worth doing.

`TabView` can have only one of the three view hierarchy configurations:
```kotlin
// #1 For custom view
TabView { CustomView() }

// #2 For icon and/or text (API 18+)
TabView { Icon(), Text() }

// #3 For icon and/or text (API 17-)
TabView { 
    FrameLayout(wrap_content, wrap_content) { Icon() }, 
    FrameLayout(wrap_content, wrap_content) { Text() } 
}
```
The "PR" way and the "conservative" way will work the same in these cases, but the "PR" way will have a simpler implementation, it's more logically correct, and it's consistent with the behavior of a tab with a custom view set.

drchen:
The behavior is the same in the current happy flow, yes. : )

However clients can actually add any random view to a tab view, and the implementation of tab view might also changed in the future as well. 

If you insist, I can bring this to the internal review and see how the team thinks.

pubiqq:
> However clients can actually add any random view to a tab view, and the implementation of tab view might also changed in the future as well.

This is unlikely, but even if so, the added views will become part of the `TabView` content and logically should be accounted for in the `getContentWidth()` and `getContentHeight()` methods (just as it happens with the custom view). So changing behavior in such an unlikely case is more of a fix than a breakdown.

pubiqq:
bump

pubiqq:
bump

pubiqq:
bump

pubiqq:
bump

pubiqq:
bump

pubiqq:
any news?

drchen:
Dan, can you provide your thoughts here?

dsn5ft:
The fix seems reasonable to me, but I think we don't need it anymore now that the minSdkVersion is 19 for our library (and all androidx libraries).

Closing the PR but feel free to comment or reopen if you feel it is still necessary.

PR #3347, opened 2023-04-13
[Internal] Guard against NPE in ManufacturerUtils

mshafrir-stripe:
In non-Robolectric JVM tests, `Build.MANUFACTURER` is null.

To guard against a potential NPE, return empty string instead of null.

Fixes https://github.com/cashapp/paparazzi/issues/380
#### Don't forget:

### DIFF
--- a/lib/java/com/google/android/material/internal/ManufacturerUtils.java
+++ b/lib/java/com/google/android/material/internal/ManufacturerUtils.java
@@ -15,35 +15,45 @@
  */
 package com.google.android.material.internal;
 
-import com.google.android.material.R;
-
 import android.os.Build;
+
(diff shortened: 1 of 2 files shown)


drchen (approved):
Thanks for the fix! (Not sure if we still need the util class though, will check that during the internal review.)

mshafrir-stripe:
> Thanks for the fix! (Not sure if we still need the util class though, will check that during the internal review.)

Thanks for the review. Is there a way for me to merge this, or will you merge it after it passes internal review?

drchen:
It has to be merged from our internal review process. 

It should not take too long as the change is straightforward. : )

PR #3356, opened 2023-04-20
[Badge] Fix badge issues

pubiqq:
This PR returns the missing diff between [the commit in the master branch](https://github.com/material-components/material-components-android/commit/4d50aa41e714b4b0e0ec4a40672431d4d4f7d6d9) and [the proposed PR](https://github.com/material-components/material-components-android/pull/3321):
- No need to invalidate the badge when clearing an already clear value (as it was in [1.8.0](https://github.com/material-components/material-components-android/blob/1.8.0/lib/java/com/google/android/material/badge/BadgeDrawable.java#L494-L497) and before).
- Fix table formatting in BadgeDrawable.md.
- Some small fixes.

### DIFF
--- a/docs/components/BadgeDrawable.md
+++ b/docs/components/BadgeDrawable.md
@@ -92,12 +92,12 @@ center, use `setHorizontalOffset(int)` or `setVerticalOffset(int)`
 
 | Feature               | Relevant attributes                                                                                                                                      |
 |-----------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------|
-| Color                 | `app:backgroundColor` <br> `app:badgeTextColor`                                                                                                                                      |
+| Color                 | `app:backgroundColor` <br> `app:badgeTextColor`                                                                                                          |
 | Width                 | `app:badgeWidth` <br> `app:badgeWithTextWidth`                                                                                                           |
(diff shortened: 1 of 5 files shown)


imhappi:
Thanks for the PR as always @pubiqq !

Please note that the suffix name was changed from `mtrl_exceed_max_badge_text_suffix` to `m3_exceed_max_badge_text_suffix` intentionally, as we use the `mtrl` prefix for values that reflect the m2 style, and `m3` accordingly for m3.

Will submit the change to stop invalidating the badge when it is already clear.

pubiqq:
> Please note that the suffix name was changed from mtrl_exceed_max_badge_text_suffix to m3_exceed_max_badge_text_suffix intentionally, as we use the mtrl prefix for values that reflect the m2 style, and m3 accordingly for m3.

`*_exceed_max_badge_text_suffix` is used in both m2 and m3 styles, so, I think, it's not correct to use the `m3` prefix for it.

imhappi:
While used for both m2 and m3 styles, this value was introduced in m3 and reflects our current m3 style which is why it uses the `m3` prefix.

pubiqq:
But it doesn't reflect the m3 style specifically, it's not even used in m3 by default.
Anyway, it's not that crucial to me, so do as you see fit.

Issue #2885, opened 2022-08-02
[BottomSheetBehavior] Giving touch priority to a nested bottom sheet


OxygenCobalt:
**Is your feature request related to a problem? Please describe.**
My app uses two bottom sheets, one nested into the other. Overall, there are few issues with it, except for the fact that the container bottom sheet intercepts most touch and nested scrolling events to it, preventing me from swiping up the nested sheet if there is a scrolling view within it, and preventing me from swiping down the nested sheet normally once it is expanded.

**Describe the solution you'd like**
An attribute like `behavior_nestedBottomSheetId` that allows me to specify the ID of a nested bottom sheet within the layout. The parent bottom sheet would then check for that ID, see if the touch/nested scrolling event actually applies to the nested sheet, and then delegate to the nested sheet instead of consuming the 

**Describe alternatives you've considered**
I can hack around this with the following:

```kotlin
override fun onInterceptTouchEvent(
    parent: CoordinatorLayout,
    child: V,
    event: MotionEvent
): Boolean = super.onInterceptTouchEvent(parent, child, event) && state != STATE_EXPANDED // Do not consume nested scrolling events when expanded so they are delegated to the nested sheet
```

And I could give touch priority to normal dragging by implementing something like this in a callback:
```kotlin
containerBottomSheetBehavior.isDraggable = nestedBottomSheetBehavior.state == BottomSheetBehavior.STATE_COLLAPSED // if nested sheet is expanded, do not intercept dragging downwards
```

But I don't really feel like these will work correctly in all cases compared to a native solution for handling touch events.

**Additional context**
I can understand if you reject this, since I think nesting sheets like this is not really expected by the library. If you want me to file more particular bug reports for the issues I encountered, that is also okay.

PR #3379, opened 2023-05-05
[BottomSheetBehavior] Fix issue with nested scrolling child views

### LINKED ISSUE #3377: [BottomSheetBehavior] Nested scrolling child view does not scroll when its inside nested scrolling parent
**Description:** When a nested scrolling child view is inside a nested scrolling parent view and the child view changes the scroll position, the touch events get intercepted by the bottom sheet behavior. For example, if I put a ViewPager2 in the bottom sheet dialog and put a vertically scrollable RecyclerView inside the ViewPager2, and navigate to the second page, the RecyclerView becomes unscrollable.

**Expected behavior:** Nested scrolling child view must be scrollable and should not be intercepted by bottom sheet behavior.

**Minimal sample app repro:** https://github.com/UdaraWanasinghe/bottom-sheet-nested-scrolling-test

**Android API version:** All affected.

**Material Library version:** 1.8.0

UdaraWanasinghe:
Fixes #3377

**Root cause**
BottomSheetBehavior stores the reference of the nestedScrollingChild view but it only sets this reference during the layout pass. When this nested scrolling child view is inside another scrollable view and the scroll position changes, the BottomSheetBehavior doesn't see this change and it compares if the touched point is inside the nestedScrollingChild with the old view reference. This results in touch events not propagating to the descendent views.

**How PR fixes it**
Instead of setting nested scrolling child reference in the layout pass, I changed to set it during the touch-down event. Additionally, it will be checked if the touch-down point is inside the nested scrolling child view.

### DIFF
--- a/lib/java/com/google/android/material/internal/res/layout/design_navigation_item_subheader.xml
+++ b/lib/java/com/google/android/material/internal/res/layout/design_navigation_item_subheader.xml
@@ -21,5 +21,6 @@
     android:paddingEnd="?attr/listPreferredItemPaddingEnd"
     android:gravity="center_vertical|start"
     android:maxLines="1"
+    android:textAlignment="viewStart"
     android:textAppearance="@style/TextAppearance.AppCompat.Body2"
     android:textColor="?android:textColorSecondary"/>


OxygenCobalt:
Would this also resolve #2885? They seem like very similar classes of issues.

UdaraWanasinghe:
@OxygenCobalt This fix does not resolve your issue. You only have to get the coordinator layout of the nested bottom sheet and set `isNestedScrollingEnabled` to true.

zLordZachy (commented):
It look good why it is still not accepted?

PR #3414, opened 2023-05-19
[MaterialDatePicker] Allow client app to access user selected inputMode

### LINKED ISSUE #3397: [MaterialDatePicker] Expose user selected input mode
I would like to manage the user preferred mode of using the MaterialDatePicker. So whenever the dialogue is closed, I would store the input mode it was used in, which would allow to open it with the same mode next time. MaterialTimePicker has this affordance by means of method getInputMode.

MaterialDatePicker should update its internal variable inputMode, when the headerToggleButton is toggled, and provide read access to it.

If there is interest in adding this, I'll happily provide a pull request.

mtotschnig:
closes #3397


mtotschnig:
I just discovered that with my suggested change, the selected input mode is lost upon device rotation. I am adding a commit to fix this.

PR #3450, opened 2023-06-12
[Carousel] Update mask size when MaskableFrameLayout size is changed

pubiqq:


pubiqq:
bump

imhappi:
@hunterstich  could you take a look to see if this was already addressed in your other change?

hunterstich:
Hi @pubiqq,

Thanks for the fix! I actually have a change that's in review currently that addresses this issue along with some catalog issues in a slightly different way. Do you mind if I close this and use the change that's coming through internally?

pubiqq:
👌

pubiqq:
Looks like https://github.com/material-components/material-components-android/commit/14023d2c8589a515ae0d80f75f301077aa324f08 was rolled back by https://github.com/material-components/material-components-android/commit/64b066a0c46e78f2c41403dd566fbc1ce8125b66. Would you mind reopening and re-reviewing my PR?

dsn5ft:
@hunterstich @imhappi any status update here?

imhappi:
I took a look at the PR, heads up that maskXPercentage is only set in `setMaskXPercentage` which is deprecated and not being used by any of our code. I've updated the PR to only update the mask rect with the mask x percentage if it was set in case someone is using the deprecated method, will send it out for review.

The mask rect actually gets directly set by the CarouselLayoutManager; we may have to make some other changes to let the CLM recalculate its keylines/masks in the `onLayoutChildren` method which should get called if the MaskableFrameLayout size is changed so I'll re-open https://github.com/material-components/material-components-android/issues/3490 but close this PR once it gets reviewed

pubiqq:
I created this PR when `setMaskXPercentage` was not deprecated yet. If you think it doesn't need fixing, you can just close the PR.

PR #3481, opened 2023-07-21
[Internal][Gradle] Update gradle wrapper to 7.6.2

TacoTheDank:
Currently they are almost 7 years old. That is quite ancient. It would be good for them to be kept completely up-to-date. This may slightly improve building times on local machines. See [here](https://docs.gradle.org/7.6.2/userguide/gradle_wrapper.html#:~:text=Using%20the%20wrapper%20task%20ensures%20that%20any%20optimizations%20made%20to%20the%20Wrapper%20shell%20script%20or%20batch%20file%20with%20that%20specific%20Gradle%20version%20are%20applied%20to%20the%20project.%20As%20usual%2C%20you%20should%20commit%20the%20changes%20to%20the%20Wrapper%20files%20to%20version%20control.) for why updating these is recommended.

The command I ran was `./gradlew wrapper --gradle-version 7.6.2 --distribution-type all` in Git Bash, which automatically updates all four files to whatever version specified. Since the JAR is already 7.6.2, it didn't change, leaving the three here.

The `networkTimeout` option is just a [new thing from Gradle](https://docs.gradle.org/7.6.2/userguide/gradle_wrapper.html#:~:text=The%20network%20timeout%20to%20use%20when%20downloading%20the%20gradle%20distribution%2C%20in%20ms.%20The%20default%20value%20is%2010000.).


paulfthomas (approved):
Thanks @TacoTheDank

PR #3516, opened 2023-08-10
[Theme] Fix swapped error colors for dynamic contrast theme in light mode

patzly:
Commit 862a7e1 which introduced dynamic contrast had a copy/paste mistake, the color values of onError and errorContainer in [L223 and L224](https://github.com/material-components/material-components-android/blob/09a058ddd9e69c8e36fba0537bf02b2f22b261f9/lib/java/com/google/android/material/theme/res/values/themes_overlay.xml#L223C5-L224C81) were swapped in light mode so an error container has a white background with 1.11.0-alpha01.


pekingme (approved):
Nice! Thanks for fixing this bug.

patzly:
Thanks for reviewing, glad I could help!

Issue #3545, opened 2023-08-25
[Catalog] status bar is not transparent in TopAppBarScrollingTransparentStatusDemoActivity (Scrolling Demo (transparent status bar))


manabu-nakamura:
**Description:** The status bar is not transparent in TopAppBarScrollingTransparentStatusDemoActivity (Scrolling Demo (transparent status bar)).
![Screenshot_20230825_192940](https://github.com/material-components/material-components-android/assets/71050561/a7ff4c9c-fbeb-4ed9-b158-e343f14009d8)

**Expected behavior:** The status bar is transparent in TopAppBarScrollingTransparentStatusDemoActivity (Scrolling Demo (transparent status bar)).
![Screenshot_20230825_193008](https://github.com/material-components/material-components-android/assets/71050561/43909e4a-976b-41d0-8e14-c66b0e314944)

**Minimal sample app repro:** Catalog app

**Android API version:** 34

**Material Library version:** 1.11.0-alpha02

**Device:** Emulator 32.1.14 (API 32)

PR #3546, opened 2023-08-25
Make status bar transparent

manabu-nakamura:
#3545


imhappi:
Going to close this since actually the transparent status bar scrolling demo and scrolling demo layouts were swapped. This change is in review; I'll accept the other edge-to-edge PR

Issue #3547, opened 2023-08-25
[Catalog] Applied edge-to-edge to demos


manabu-nakamura:
**Is your feature request related to a problem? Please describe.**
The edge-to-edge preference does not apply to TopAppBarScrollingTransparentStatusDemoActivity and TopAppBarActionBarDemoActivity.
![Screenshot_20230825_190410](https://github.com/material-components/material-components-android/assets/71050561/a407043c-683f-4bf1-87eb-b96122397208)
![Screenshot_20230825_190425](https://github.com/material-components/material-components-android/assets/71050561/fb1a727f-3e05-423a-9633-00be8fd85f05)

**Describe the solution you'd like**
The edge-to-edge preference applies to TopAppBarScrollingTransparentStatusDemoActivity and TopAppBarActionBarDemoActivity.
![Screenshot_20230825_191302](https://github.com/material-components/material-components-android/assets/71050561/fd7280d3-9ce3-4d5b-b61d-a7a12ce14b73)
![Screenshot_20230825_191320](https://github.com/material-components/material-components-android/assets/71050561/f8e06e72-8293-4ab0-88a2-e99c6e9d8ab4)
Additionally, make the status bar transparent (#3545):
![Screenshot_20230825_204900](https://github.com/material-components/material-components-android/assets/71050561/5fd319fb-be35-4a78-8518-9b290082670d)

PR #3551, opened 2023-08-29
Applied edge-to-edge to demos

manabu-nakamura:
I think this is better than https://github.com/material-components/material-components-android/commit/aa5b5bc157d935cf616ce8d172d157d008308038.
https://github.com/material-components/material-components-android/commit/ebfb3dbc6ce2cee9f4c57c991c53f09c0eb7568a
#3547


leticiarossi:
Hi there, so we initially had the top app bar demos not be edge to edge due to glitches with the status bar in action mode (see https://github.com/material-components/material-components-android/commit/ebfb3dbc6ce2cee9f4c57c991c53f09c0eb7568a). I'll revert your other change and close this PR, apologies about that!

Issue #3573, opened 2023-09-14
[Catalog][BottomAppBar][BottomSheet][Predictive Back] When bottom sheet is collapsed, it will not detach from left and right edges of screen


manabu-nakamura:
**Description:** When the bottom sheet is collapsed, the back callback is **disabled** and the bottom sheet will **not** detach from the left and right edges of the screen.
https://m3.material.io/components/bottom-sheets/guidelines#3d7735e2-73ea-4f3e-bd42-e70161fc1085
https://github.com/material-components/material-components-android/blob/master/docs/components/BottomSheet.md#standard-non-modal-bottom-sheets
![Screenshot_20230913_172011](https://github.com/material-components/material-components-android/assets/71050561/c10ffbb4-4553-4a5e-8a17-db49436145f4)

**Expected behavior:** When the bottom sheet is collapsed, the back callback is **enabled** and the bottom sheet will detach from the left and right edges of the screen.

**Minimal sample app repro:** Catalog app

**Android API version:** 34

**Material Library version:** 1.11.0-alpha02

**Device:** Emulator 32.1.15 (API 34)

manabu-nakamura:
This is not resolved.

dsn5ft:
I don't think this is really an issue. The bottom sheet in the bottom app bar demo is marked as hideable, so when it "collapses" the bottom sheet is removed from the screen and the callback is disabled anyway.

The main issue of the bottom sheet not detaching predictively after rotation was fixed by https://github.com/material-components/material-components-android/commit/f9102c745ad8b84dc16a4914665f2bb40aa4f5a2 because now the expanded state is correctly restored.

If there is still a functional issue with the demo can you record a video and provide some more explanation?

dsn5ft:
Oh interesting, I didn't realize it can actually be collapsed without being hidden. I guess that's only possible in landscape for some reason?

manabu-nakamura:
I think it's because the bottom sheet height is low.

I think this logic[^1] is for the persistent (not hideable) bottom sheet, not for the hideable bottom sheet.
[^1]: https://github.com/material-components/material-components-android/blob/master/docs/components/BottomSheet.md#standard-non-modal-bottom-sheets

PR #3574, opened 2023-09-14
[Catalog] When bottom sheet is collapsed, back callback is enabled

manabu-nakamura:
#3573


dsn5ft:
Thanks for these issues and PRs @manabu-nakamura.

I'm thinking of handling this a bit differently, by updating the state text and back callback enabled to be correct on rotation, instead of hiding/collapsing the bottom sheet on rotation. Will sync out the fix within the next few days.

Issue #3579, opened 2023-09-14
[Carousel] Part of the next item is visible when FullScreenCarouselStrategy is set


pubiqq:
<img src="https://user-images.githubusercontent.com/82187521/268045005-8640e4a6-012d-4856-86a6-5758cf9026aa.png" width="270px" />

**Material Library version:** 1.11.0-alpha02
**Device:** Pixel 4a (5G)

Issue #3580, opened 2023-09-14
[Carousel] A single item is incorrectly displayed in RTL mode


pubiqq:
<img src="https://github.com/material-components/material-components-android/assets/82187521/8eed70c6-abb9-410c-abd2-ed37a116ad0e" width="270px" />

**Material Library version:** 1.11.0-alpha02
**Device:** Pixel 4a (5G)

hunterstich:
Hey @pubiqq!

Thanks for filing this! I've actually been working on a fix for https://github.com/material-components/material-components-android/issues/3554, which I think should solve this issue as well. Should be completed soon and we can double check once the change lands!

PR #3581, opened 2023-09-14
[Carousel] Reduce the number of truncations in intermediate calculations

pubiqq:
Fixes https://github.com/material-components/material-components-android/issues/3579 and https://github.com/material-components/material-components-android/issues/3580.


hunterstich:
This is great, @pubiqq! Thank you so much. This has been reviewed and should be synced out today!

PR #3588, opened 2023-09-18
[MaterialTimePicker] Fix the dialog buttons getting clipped at the bottom

### LINKED ISSUE #3584: [MaterialTimePicker] Buttons at the bottom are getting cut off
**Description:** `MaterialTimePicker` is getting its buttons cut off
at the bottom in the `1.11.x` releases, this looks correct in `1.10.0-beta01`
so it may be a regression.

I suspect this might improve if those two buttons are also constrained
to `parent` at their bottom edge.

**Expected behavior:** The expected behaviour is that those buttons
would not be clipped. Here are two screenshots - one from `1.10` and
one from `1.11`, notice how the dialog's height is less than it was in `1.10`
and the buttons are clipped:

1.10|1.11
-|-
Screenshot_1694795666|Screenshot_1694795666 copy

**Source code:** `material_timepicker_dialog.xml`; This may be due to
commit d10201dc88f919421b3bc4d8d7aa432de685c892 which updated the layout

**Minimal sample app repro:** n/a

**Android API version:** 33

**Material Library version:** 1.11.0-alpha02

**Device:** Pixel 6 emulator

mck182:
The "Cancel" and "Ok" buttons in the dialog are being clipped at the bottom
when the theme's `minTouchTargetSize` is not the default `48dp`.

This is because those two buttons are constrained only to the top of
the "input mode" button and have no bottom constraint. The "input mode"
button is the only one constrained to `parent` on the bottom edge, which
means that if this button gets smaller than those other two buttons - like
when the `minTouchTargetSize` is `0dp` - the bottom edge of the dialog
will move up, but because those other two buttons are only constrained
at the top edge, they stay in place and get clipped.

So this adds a bottom constraint to the "Cancel" and "Ok" buttons, which
fixes the clipping issue. I've also changed the end margin from `8dp` to
`12dp` to match the start margin of the "input mode" button, but I'm not
too sure about that one, it just seemed nicer to have equal margins on both
sides of the dialog.

closes #3584

Here are the screenshots with `minTouchTargetSize` set to `0dp`:

Before|After
-|-
Screenshot_1695055330|Screenshot_1695055255
Before with layout bounds|After with layout bounds
Screenshot_1695051473|Screenshot_1695051555

### DIFF
--- a/lib/java/com/google/android/material/timepicker/res/layout/material_timepicker_dialog.xml
+++ b/lib/java/com/google/android/material/timepicker/res/layout/material_timepicker_dialog.xml
@@ -83,7 +83,8 @@
     android:minWidth="72dp"
     android:text="@string/mtrl_timepicker_cancel"
     app:layout_constraintEnd_toStartOf="@id/material_timepicker_ok_button"
-    app:layout_constraintTop_toTopOf="@id/material_timepicker_mode_button" />
+    app:layout_constraintTop_toTopOf="@id/material_timepicker_mode_button"
+    app:layout_constraintBottom_toBottomOf="parent" />
 
   <Button
     android:id="@+id/material_timepicker_ok_button"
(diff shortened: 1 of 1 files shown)


liu-wanshun (approved):

nikclayton:
Ping.

alopix:
Curious if this fix is onna get reviewed again? It seems that it has been two years and I was actually gonna open a PR myself to fix this as I encountered the issue in an Android app I use, but looks like this could have been included in a release already?

PR #3592, opened 2023-09-20
[Catalog] When bottom sheet is collapsed, back callback is enabled

manabu-nakamura:
#3573


dsn5ft:
I don't think this is really an issue. The bottom sheet in the bottom app bar demo is marked as hideable, so when it "collapses" the bottom sheet is removed from the screen and the callback is disabled anyway.

The main issue of the bottom sheet not detaching predictively after rotation was fixed by f9102c7 because now the expanded state is correctly restored.

If there is still a functional issue with the demo can you record a video and provide some more explanation?

Issue #3610, opened 2023-09-30
[Catalog][Navigation Drawer] After restarting activity and pushing back button, opened navigation drawer will not be closed but activity will be closed


manabu-nakamura:
#3593

**Description:** After restarting the activity (e.g. rotating the device) and pushing the back button, the opened navigation drawer will not be closed but the activity will be closed.
![Screenshot_20230930_154601](https://github.com/material-components/material-components-android/assets/71050561/67ab0a94-e592-4498-86be-d220fd3e3003)

**Expected behavior:** After restarting the activity and pushing the back button, the activity will not be closed but the opened navigation drawer will be closed.

We don't have to use the OnBackPressedCallback if the DrawerLayout itself uses the OnBackPressedCallback...

**Minimal sample app repro:** Catalog app (NavigationDrawerDemoActivity)

**Android API version:** 34

**Material Library version:** 1.11.0-alpha03

**Device:** Emulator 32.1.15 (API 32, 33, and 34)

dsn5ft:
Hi @manabu-nakamura I did make a change in the library to address this type of issue within NavigationView (https://github.com/material-components/material-components-android/commit/bccbd4f778f59a263ae627be771d5d4409ce4dd2).

Can you sync to head and see if your PR is still necessary? I guess it might be for older API levels, but that's more of an issue with DrawerLayout itself. I also want to make sure your PR doesn't conflict with my change by overriding the predictive back behavior in any way.

manabu-nakamura:
Tnanks to https://github.com/material-components/material-components-android/commit/bccbd4f778f59a263ae627be771d5d4409ce4dd2, without this PR, after restarting the activity and pushing the back button, the activity will not be closed but the opened navigation drawer will be closed correctly in Android 13+. The opened navigation drawer will not be closed but the activity will be closed incorrectly in Android 12-.

PR #3611, opened 2023-09-30
[Catalog][Navigation Drawer] Update navigation drawer state handling to be more accurate in demo

manabu-nakamura:
#3610


manabu-nakamura:
Is the code for the predictive back in drawerlayout:1.2.0 unneccesary?
The catalog app doesn't use drawerlayout:1.2.0 but uses drawerlayout:1.1.1.
https://developer.android.com/jetpack/androidx/releases/drawerlayout#1.2.0

Issue #3635, opened 2023-10-16
[Navigation Drawer][Side Sheet][Bottom Sheet][Date Picker][Time Picker][Dialog] Unify scrim color


manabu-nakamura:
**Is your feature request related to a problem? Please describe.**
The scrim color of the navigation drawer, the side sheet, the bottom sheet, the date picker, and the time picker is darker than the scrim color of the dialog. Why? The icons in the system navigation bar are hard to see.
![Screenshot_20230904_143239](https://github.com/material-components/material-components-android/assets/71050561/19c5d936-1ee4-487f-9ca0-ac561dffc828)
![Screenshot_20230904_143854](https://github.com/material-components/material-components-android/assets/71050561/3de7fa7b-235d-479e-b2bc-dd68b9b15ce6)
![Screenshot_20230914_195902](https://github.com/material-components/material-components-android/assets/71050561/4d865814-05ac-4385-9742-d044f741c5eb)
![Screenshot_20231019_095504](https://github.com/material-components/material-components-android/assets/71050561/f5f861aa-31b6-4e25-ae07-e1a415e039eb)
![Screenshot_20231019_095302](https://github.com/material-components/material-components-android/assets/71050561/9b7353eb-8a75-431c-ae7f-86dca446eb46)
![Screenshot_20230905_161807](https://github.com/material-components/material-components-android/assets/71050561/835707ed-bcd5-4434-9d14-c094d59b66a3)
https://android.googlesource.com/platform/frameworks/support/+/refs/heads/androidx-drawerlayout-release/drawerlayout/drawerlayout/src/main/java/androidx/drawerlayout/widget/DrawerLayout.java#172
https://android.googlesource.com/platform/prebuilts/fullsdk/platforms/+/refs/heads/main/android-34/data/res/values/themes_material.xml#53
https://android.googlesource.com/platform/prebuilts/fullsdk/platforms/+/refs/heads/main/android-34/data/res/values/themes_material.xml#424
https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/dialog/res/values/themes_overlay.xml#L38

**Describe the solution you'd like**
Unify the scrim color (lighten the scrim color of the navigation drawer, the side sheet, the bottom sheet, the date picker, and the time picker).
![Screenshot_20230904_143433](https://github.com/material-components/material-components-android/assets/71050561/a82c62bb-a562-4c95-864b-08d8a02ef836)
[...]

PR #3643, opened 2023-10-19
Unify scrim color

manabu-nakamura:
#3635 

androidx/drawerlayout/widget/DrawerLayout.java:
```
-   private static final int DEFAULT_SCRIM_COLOR = 0x99000000;
+   private static final int DEFAULT_SCRIM_COLOR = 0x52000000;
```


pekingme (approved):
Thanks for adding this!

PR #3849, opened 2023-11-14
Remove redundant version checks

MGaetan89:
Now that the minimum SDK version of Material Components is 19, there was a lot of code that could be simplified or removed.

### DIFF
--- a/lib/java/com/google/android/material/badge/BadgeUtils.java
+++ b/lib/java/com/google/android/material/badge/BadgeUtils.java
@@ -47,7 +47,7 @@ import com.google.android.material.internal.ToolbarUtils;
 @ExperimentalBadgeUtils
 public class BadgeUtils {
 
-  public static final boolean USE_COMPAT_PARENT = VERSION.SDK_INT < VERSION_CODES.JELLY_BEAN_MR2;
+  public static final boolean USE_COMPAT_PARENT = false;
 
   private static final String LOG_TAG = "BadgeUtils";
 
--- a/catalog/java/io/material/catalog/application/theme/res/values/themes.xml
+++ b/catalog/java/io/material/catalog/application/theme/res/values/themes.xml
@@ -26,7 +26,7 @@
 
   <!-- A theme which sets the status bar color to transparent -->
   <style name="Theme.Catalog.TransparentStatus">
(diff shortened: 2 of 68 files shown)


lib/java/com/google/android/material/badge/BadgeUtils.java
```diff
@@ -47,7 +47,7 @@
 @ExperimentalBadgeUtils
 public class BadgeUtils {
 
-  public static final boolean USE_COMPAT_PARENT = VERSION.SDK_INT < VERSION_CODES.JELLY_BEAN_MR2;
+  public static final boolean USE_COMPAT_PARENT = false;
```
MGaetan89, line 50: I didn't know how to proceed here: `BadgeUtils` is a public class, and so is this `USE_COMPAT_PARENT` constant.
Can I simply remove it? Or should I deprecate it for removal in a future release?

MGaetan89:
I've fixed the conflict in `CarouselLayoutManagerRtlTest`

MGaetan89:
@hunterstich I've updated the PR and removed some additional legacy code.

leticiarossi:
Thank you for working on this! I removed the change on the Badge file since it'd be best to refactor badge instead of just having a variable set to false, so that should be done in a separate change.

MGaetan89:
> Thank you for working on this! I removed the change on the Badge file since it'd be best to refactor badge instead of just having a variable set to false, so that should be done in a separate change.

Are you open for a PR to remove `BadgeUtils.USE_COMPAT_PARENT`?

leticiarossi:
sure if you'd like to contribute feel free to do so and we'll happily review it! :)

MGaetan89:
I'll prepare a PR then 🙂 
Just to be sure that we're aligned: the goal is to remove `BadgeUtils.USE_COMPAT_PARENT` and every it's used, correct?

leticiarossi:
that's right, we should remove the checks for `BadgeUtils.USE_COMPAT_PARENT` similar to how you removed the checks for API <= 19

MGaetan89:
Ok, https://github.com/material-components/material-components-android/pull/4170 is ready when you have time to have a look 🙂

Issue #3124, opened 2022-12-01
[RangeSlider] Label got stuck when used in BottomSheet


mhettig:
**Description:**
When adding a RangeSlider to a BottomSheet with the labels always visible (= `app:labelBehavior="visible"`) the labels will not move together with the slider when the bottom sheet is dragged. They get stuck:

https://user-images.githubusercontent.com/2598617/205148343-5c6cf634-ac8a-4805-836c-6b8815e8b9a3.mp4

**Expected behavior:**
Labels stay fixed at the top of the RangeSlider and will be in sync with the bottom sheet movement.

**Source code:**
To provide a simple and reproducible example I used the Catalog App from this repository.

There I added the following code snippet to `cat_bottomsheet_fragment.xml`. It was included at the bottom of the BottomSheet so below the last MaterialSwitch.

```
<com.google.android.material.slider.RangeSlider
  android:layout_width="match_parent"
  android:layout_height="wrap_content"
  android:layout_below="@id/cat_bottomsheet_enabled_switch"
  app:labelBehavior="visible" />
```

**Android API version:** 33

**Material Library version:** 1.7.0 - Bug is also visible with  version 1.6.0 when label behavior "visible" was firstly introduced

**Device:** Xiaomi Mi 11 Ultra (with xiaomi.eu firmware)

leticiarossi:
Thank you for reporting, this looks like the same issue as https://github.com/material-components/material-components-android/issues/2869 so will add this to there and close it here as duplicate

PR #3917, opened 2023-12-10
[Slider] Fix slider label not moving

manabu-nakamura:
#3124
https://github.com/material-components/material-components-android/issues/2869#issuecomment-1821123500

### DIFF
--- a/lib/java/com/google/android/material/slider/BaseSlider.java
+++ b/lib/java/com/google/android/material/slider/BaseSlider.java
@@ -361,6 +361,8 @@ abstract class BaseSlider<
           }
         }
       };
+  @NonNull
+  private final ViewTreeObserver.OnGlobalLayoutListener onGlobalLayoutListener = onScrollChangedListener::onScrollChanged;
 
   /**
    * Determines the behavior of the label which can be any of the following.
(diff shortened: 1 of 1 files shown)


lib/java/com/google/android/material/slider/BaseSlider.java
```diff
@@ -361,6 +361,8 @@ abstract class BaseSlider<
           }
         }
       };
+  @NonNull
+  private final ViewTreeObserver.OnGlobalLayoutListener onGlobalLayoutListener = onScrollChangedListener::onScrollChanged;
```
drchen, line 365: This is sort of confusing. Can we declare a method, say, `updateLableVisibility()`, and do something like:

```
private final ViewTreeObserver.OnScrollChangedListener onScrollChangedListener = this::updateLableVisibility;
private final ViewTreeObserver.OnGlobalLayoutListener onGlobalLayoutListener = this::updateLableVisibility;
```

instead?

manabu-nakamura: Thank you very much. I tried. I'm not sure if the name of the method is appropriate.

lib/java/com/google/android/material/slider/BaseSlider.java
```diff
@@ -341,26 +341,9 @@ abstract class BaseSlider<
...
-              ViewUtils.getContentViewOverlay(this).add(label);
-            } else {
-              ViewUtils.getContentViewOverlay(this).remove(label);
-            }
-          }
-        }
-      };
+  private final ViewTreeObserver.OnScrollChangedListener onScrollChangedListener = this::updateLabel;
+  @NonNull
+  private final ViewTreeObserver.OnGlobalLayoutListener onGlobalLayoutListener = this::updateLabel;
```
dsn5ft, line 346: do we need both the global layout listener and the scroll listener or is just the global layout listener enough?

manabu-nakamura: By only the global layout listener, the slider label doesn't move while scrolling (#2869).

Issue #3978, opened 2024-01-08
[RangeSlider] Crash occurs when assigning a large value to `valueTo`.


Park-SM:
**Description:** Crash occurs when assigning a large value to `valueTo`.
To reproduce, set the values ​​as follows.
- `valueFrom` : 1000.0
- `valueTo` : 258896000.0
- `stepSize` : 1000.0

**Expected behavior:** 
<img src="https://github.com/material-components/material-components-android/assets/47319426/4ac9e6eb-5f15-46b0-92f1-c0d6229c535a" width="300" />

**Source code:** 
The cause is the floating point rounding error.
```java
private boolean valueLandsOnTick(float value) {
    // Check that the value is a multiple of stepSize given the offset of valueFrom.
    return isMultipleOfStepSize(value - valueFrom);
}
```

**Minimal sample app repro:** https://github.com/Park-SM/material-components-android/tree/reproduced_branch

**Android API version:** Android API 33

**Material Library version:** 1.9.0

**Device:** Galaxy Z Flip4

paulfthomas:
Fixed in ac77b4cfb5ac15b37240fd455233cbca237e103f

PR #3979, opened 2024-01-08
[RangeSlider] Resolve issues that crash when assigning a large value to `valueTo`.

Park-SM:
Resolve the [issue](https://github.com/material-components/material-components-android/issues/3978) that crash when assigning a large value to `valueTo`.
The cause is the floating point rounding error.
```java
// AS-IS
private boolean valueLandsOnTick(float value) {
    // Check that the value is a multiple of stepSize given the offset of valueFrom.
    return isMultipleOfStepSize(value - valueFrom);
}

// TO-BE
private boolean valueLandsOnTick(float value) {
    // Check that the value is a multiple of stepSize given the offset of valueFrom.
    BigDecimal bigDecimal =
        new BigDecimal(Float.toString(value))
            .subtract(new BigDecimal(Float.toString(valueFrom)), MathContext.DECIMAL64);
    return isMultipleOfStepSizeByBigDecimal(bigDecimal);
}
```

### DIFF
--- a/lib/java/com/google/android/material/slider/BaseSlider.java
+++ b/lib/java/com/google/android/material/slider/BaseSlider.java
@@ -613,7 +613,7 @@ abstract class BaseSlider<
   }
 
   private boolean isMultipleOfStepSize(float value) {
-    return isMultipleOfStepSizeByBigDecimal(new BigDecimal(value));
+    return isMultipleOfStepSizeByBigDecimal(new BigDecimal(Float.toString(value)));
   }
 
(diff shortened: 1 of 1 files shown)


Park-SM:
@drchen Hello.

Can i get why this PR is closed?

Park-SM:
Hi @drchen, Can i get why this PR is closed?

paulfthomas:
The fix is merged https://github.com/material-components/material-components-android/commit/ac77b4cfb5ac15b37240fd455233cbca237e103f

PR #4015, opened 2024-01-23
[Docs] Fix anchors

pubiqq:


raajkumars:
Were these anchors broken or are you just changing them to title case?

pubiqq:
They are still broken, try clicking on the "Design & API Documentation" link [here](https://github.com/material-components/material-components-android/blob/5688027/docs/components/Button.md), for example. 
Title/sentence case doesn't affect the functionality of the links, I changed it just to make the link names match the headers.

raajkumars (approved):
LGTM

leticiarossi:
I'm not sure adding an extra `-` will work either. It'd probably be best to change the `&` to be `and` so we have `Design and API Documentation` and the link can be `#design-and-api-documentation`

pubiqq:
> I'm not sure adding an extra `-` will work either. 

Well, you can compare the [current version](https://github.com/material-components/material-components-android/blob/fb4761cc20599a4ef192f94512c554e0e266b96e/docs/components/Button.md) with the [proposed](https://github.com/pubiqq/material-components-android/blob/0b4097653df971659ed1526e450bc40030f502e3/docs/components/Button.md) one yourself. You don't think I just added hyphens to random places for no reason, right? 😂

Or I can just replace all the ampersands in the headings with "and", as you wish.

leticiarossi:
Ah, I only tested internally where we have the docs and there the current link works and the new one doesn't, but on github I see now it's the opposite. Sometimes there are discrepancies like this with github so changing to `Design and API Documentation` would be the best option then to work on both places!

Side note - of course I didn't think you'd do a change for no reason, but misunderstandings and mistakes can happen from both sides. Let's cultivate a respectful environment here :)

pubiqq:
Ok, deal.

pubiqq:
Done.

leticiarossi (approved):

Issue #4036, opened 2024-02-07
[Catalog][Carousel] catalog app crashes


manabu-nakamura:
**Description:** The catalog app crashes:
1. Select 2-9 from the exposed dropdown menu.
![Screenshot_20240207_215456](https://github.com/material-components/material-components-android/assets/71050561/6c56360b-27b8-4197-ad7e-34bbd3f0a6e3)
2. Rotate the device.
![Screenshot_20240207_215511](https://github.com/material-components/material-components-android/assets/71050561/296662c9-9519-482f-8449-e2e4fbd96594)
3. Scroll the carousel to the end.

**Expected behavior:** The catalog app doesn't crash.

**Minimal sample app repro:** Catalog app (MultiBrowseCarouselDemoFragment, HeroCarouselDemoFragment, and UncontainedCarouselDemoFragment)

**Android API version:** 34

**Material Library version:** 1.12.0-alpha03

**Device:** Emulator 32.1.15 (API 32)

manabu-nakamura:
After rotating the device, the catalog app doesn't crash but the item count is reset to 10.
![Screenshot_20240418_130148](https://github.com/material-components/material-components-android/assets/71050561/7169142b-da3c-4fe2-8339-261dbe9ae245)
#4037

imhappi:
This is WAI. The item count not resetting is a known bug https://github.com/material-components/material-components-android/issues/1464

PR #4037, opened 2024-02-07
[Catalog][Carousel] catalog app doesn't crash

manabu-nakamura:
#4036


paulfthomas:
@imhappi could you take a look at this?

Issue #4070, opened 2024-02-17
[Slider] Trackpath rounded corners draws a circle


Pittvandewitt:
**Description:** At the end of a track path, a rounded corner is drawn. When the track path is at its mimimum size, the rounded corner draws a circle. Instead it should use the `trackInsideCornerSize` for the corners near the slider.

**Expected behavior:** 

https://github.com/material-components/material-components-android/assets/1721405/eda92b60-2d36-4161-bc98-4f8daf05bc5f

**Current behaviour:**

https://github.com/material-components/material-components-android/assets/1721405/4e5a2270-e330-484e-a2ef-bbcc48b94c3e

**Source code:** [The code snippet which is causing this issue](https://github.com/material-components/material-components-android/blob/dd8f6bebdd45c2875fab6574376a604a7d40061a/lib/java/com/google/android/material/slider/BaseSlider.java#L2205)

**Minimal sample app repro:** The current catalog app

**Android API version:** Android API 34

**Material Library version:** 1.12.0-alpha03

**Device:** Pixel 7

PR #4071, opened 2024-02-17
[Slider] Use trackInsideCornerSize for rounded corners near thumb

Pittvandewitt:
closes https://github.com/material-components/material-components-android/issues/4070


pekingme (approved):
Thanks for the PR!

Issue #4090, opened 2024-03-07
[Slider] Incorrect slider label behavior inside ViewPager


pubiqq:
**Description:** When swiping the ViewPager to a page with a slider with `labelBehavior="visible"`, a label appears, but it shouldn't be there until the slider is visible (see video, starts from 00:13):

https://github.com/material-components/material-components-android/assets/82187521/a359ec50-5840-4034-b9c1-3f35b99cdb60

**Material Library version:** 1.12.0-alpha03

PR #4093, opened 2024-03-07
[Slider] Use a uniform way to update labels

pubiqq:
Also fixes https://github.com/material-components/material-components-android/issues/4090


pekingme (approved):
Thanks for the PR! Port for internal reviews.

Issue #4105, opened 2024-03-17
[TabLayout] Can't customize TabView click listener


vvinogra:
**Description:**

I have a business logic in the app to perform a certain validation when the TabView is clicked.
Due to the current TabLayout implementation, I can't prevent tab selection when the tab is clicked. 

**Expected behavior:**

We should have an option to control this behavior.

**Source code:**
[TabLayout.java](https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/tabs/TabLayout.java#L2639-L2652)
```java
    @Override
    public boolean performClick() {
      final boolean handled = super.performClick();

      if (tab != null) {
        if (!handled) {
          playSoundEffect(SoundEffectConstants.CLICK);
        }
        tab.select();
        return true;
      } else {
        return handled;
      }
    }
```

**Minimal sample app repro:**

1. Create Activity/Fragment with TabLayout and a few TabView items.
2. Click on the tab

**Material Library version:** 1.11.0

PR #4107, opened 2024-03-17
[TabLayout] Fix TabView click listener customization

vvinogra:
Closes https://github.com/material-components/material-components-android/issues/4105


lib/java/com/google/android/material/tabs/TabLayout.java
```diff
@@ -2644,7 +2669,9 @@ public boolean performClick() {
         if (!handled) {
           playSoundEffect(SoundEffectConstants.CLICK);
         }
-        tab.select();
+        if (tab.onTabClickListener != null) {
+          tab.onTabClickListener.onClick(tab);
+        }
```
drchen, line 2674: Let's try to avoid overriding the default behavior unless there's a valid use case to do that.

drchen (changes requested):

lib/java/com/google/android/material/tabs/TabLayout.java
vvinogra: @drchen Hey, thanks for the comment.

The main idea is to have an option to override a default behavior.
A usecase is described in the https://github.com/material-components/material-components-android/issues/4105 issue.

tl.dr.
We need to override a default behaviour to add a validation logic when user clicks on the tab before selecting it.
With the existing implementation, I tried selecting the previous tab in case the data is not valid but it doesn't look good because the tab is selected with the animation.

drchen: Ok cool. So I think there are several improvements I'll suggest here:

1. Let's make `onTabClickListener`  be non-null.
2. Let `OnTabClickListener .onClick()` return a boolean to indicate if the click event is handled, so we can simplify the logic like:
```
final boolean handled = super.performClick() || tab.onTabClickListener.onClick(tab);
... rest the same
```
3. Refine the javadoc to make it more clear that how to override the default behavior.

lib/java/com/google/android/material/tabs/TabLayout.java
```diff
@@ -2644,7 +2669,9 @@ public boolean performClick() {
         if (!handled) {
```
drchen, line 2669: I think this should be part of the default onClick behavior as well.

drchen (changes requested):

neronguyenvn:
I have a usecase need to distinguish between the tab is selected by code `.select()` and tap on it manually, how should I do ...

PR #4117, opened 2024-03-26
[Internal] Stop using deprecated androidx.core.view.* methods

pubiqq:


pubiqq:
Done.

pubiqq:
Ok, I've removed the changes from the auto-generated classes, but unfortunately I have no way to check if they're generated correctly because I don't have access to your build system, so you'll have to check it yourself.

pubiqq:
Bump

PR #4119, opened 2024-03-27
[NavigationView] Fix section label alignment

pubiqq:
Fixes https://github.com/material-components/material-components-android/issues/4178

### DIFF
--- a/lib/java/com/google/android/material/internal/res/layout/design_navigation_item_subheader.xml
+++ b/lib/java/com/google/android/material/internal/res/layout/design_navigation_item_subheader.xml
@@ -21,5 +21,6 @@
     android:paddingEnd="?attr/listPreferredItemPaddingEnd"
     android:gravity="center_vertical|start"
     android:maxLines="1"
+    android:textAlignment="viewStart"
     android:textAppearance="@style/TextAppearance.AppCompat.Body2"
     android:textColor="?android:textColorSecondary"/>


lib/java/com/google/android/material/internal/res/layout/design_navigation_item_subheader.xml
```diff
@@ -19,7 +19,8 @@
     android:layout_height="?attr/listPreferredItemHeightSmall"
     android:paddingLeft="?attr/listPreferredItemPaddingLeft"
     android:paddingRight="?attr/listPreferredItemPaddingRight"
-    android:gravity="center_vertical|start"
+    android:gravity="center_vertical"
     android:maxLines="1"
+    android:textAlignment="viewStart"
```
leticiarossi, line 24: There's an error with this:`"Inconsistent alignment specification between textAlignment and gravity attributes: was center_vertical, expected start"`

So we'd need to keep `android:gravity="center_vertical|start"`. However then it's easier to just keep it as is without setting `textAlignment` since `gravity` is already working as intended. 

What is your reason to use textAlignment instead of gravity?

pubiqq: > There's an error with this:`"Inconsistent alignment specification between textAlignment and gravity attributes: was center_vertical, expected start"`

That's the message from the `RtlCompat` lint check, isn't it?

> What is your reason to use textAlignment instead of gravity?

I've added an issue, see description.

leticiarossi: I see, thank you for filing the issue! the message is from the lint check but it's a blocking error, but I confirmed that keeping both `android:gravity="center_vertical|start"` and `android:textAlignment="viewStart"` fixes the issue so we can do that!

pubiqq: Ok, I've rebased the branch with `android:gravity="center_vertical|start"`.

> it's a blocking error

It's a false positive. `RtlCompat` should not be used in the library because it no longer supports API < 17.

PR #4123, opened 2024-03-28
[NavigationView] Use relative listPreferredItemPaddings

pubiqq:


leticiarossi (approved):
we can remove `android:paddingLeft` and `android:paddingRight` since our min sdk is now 19 (doing that internally while syncing this PR!)

PR #4166, opened 2024-05-08
[Internal] Fix drawable compositing on API < 23

pubiqq:


leticiarossi:
Thank you for the PR! Could you explain what was the issue in APIs<23 and how does your change fix it? Why are you removing the `shouldScaleTopLayer` check?

pubiqq:
This "optimization" is just incorrect. If the desired size is `INTRINSIC_SIZE`, it doesn't mean that the image doesn't need to be scaled, because the `topLayerNewWidth` and `topLayerNewHeight` may differ from the desired ones.

I have no idea why `shouldScaleTopLayer` was [added](https://github.com/material-components/material-components-android/commit/a10c5083a9e8ffb58903d9631235fe5f71766010), it wasn't in the [original PR](https://github.com/material-components/material-components-android/pull/3513).

pubiqq:
Ping

PR #4170, opened 2024-05-10
[Badge] Remove `BadgeUtils.USE_COMPAT_PARENT`

MGaetan89:
Now that the min SDK is 19 and that #3849 is merged, the `BadgeUtils.USE_COMPAT_PARENT` constant is always `false`.
This PR removes its usages, as suggested in https://github.com/material-components/material-components-android/pull/3849#issuecomment-2102874260.


lib/java/com/google/android/material/badge/BadgeDrawable.java
```diff
@@ -408,48 +391,6 @@ public FrameLayout getCustomBadgeParent() {
...
-    if ((anchorViewParent != null && anchorViewParent.getId() == R.id.mtrl_anchor_parent)
-        || (customBadgeParentRef != null && customBadgeParentRef.get() == anchorViewParent)) {
-      return;
-    }
-    // Must call this before wrapping the anchor in a FrameLayout.
-    updateAnchorParentToNotClip(anchorView);
-
-    // Create FrameLayout and configure it to wrap the anchor.
-    final FrameLayout frameLayout = new FrameLayout(anchorView.getContext());
-    frameLayout.setId(R.id.mtrl_anchor_parent);
```
MGaetan89, line 426: `R.id.mtrl_anchor_parent` is still used here:
https://github.com/material-components/material-components-android/blob/35f11c97b64f22b9c3994b41b35c955f7a4261ed/lib/java/com/google/android/material/badge/BadgeDrawable.java#L382-L386

Should this method be removed too, or should I leave as it is now?

leticiarossi: It makes sense to also remove it and wherever it's used since that will now always be false!

MGaetan89: Done in https://github.com/material-components/material-components-android/pull/4170/commits/25d7f3536179c1ee88cc781e4be568cecf6239e1

leticiarossi (approved):

PR #4181, opened 2024-05-19
[ProgressIndicator] Fix corner radius calculation

whitipet:
Because `trackCornerRadius` is passed as an int and not a float, rounded corners may not be displayed correctly with a certain combination of screen size, `trackThickness`, and `trackCornerRadius`.
In my case, `view.getResources().getDisplayMetrics().density = 2.25`. So `trackThickness = 4dp = 9px` and `trackCornerRadius = 2dp = 4.5px`. Because of type casting, the 0.5f fraction is lost and the corners are rounded incorrectly.

<details>
  <summary>Comparison</summary>

Before | After
-|-
before  |  after
</details>


pekingme (approved):
LGTM

PR #4192, opened 2024-05-27
[Color] Fix dynamic color support check

pubiqq:


lib/java/com/google/android/material/color/DynamicColors.java
```diff
@@ -401,7 +401,7 @@ public static boolean isDynamicColorAvailable() {
     if (VERSION.SDK_INT < VERSION_CODES.S) {
       return false;
     }
-    if (BuildCompat.isAtLeastT()) {
+    if (VERSION.SDK_INT >= VERSION_CODES.S_V2) {
```
dsn5ft, line 404: Hi, shouldn't this be `>= VERSION_CODES.TIRAMISU` in order to be equivalent to `BuildCompat.isAtLeastT()`? Or `> VERSION_CODES.S_V2`.

Android T is the first version where all devices _should_ support dynamic color by default, so we don't want to return true for S_V2.

pubiqq: > Hi, shouldn't this be `>= VERSION_CODES.TIRAMISU` in order to be equivalent to `BuildCompat.isAtLeastT()`? Or `> VERSION_CODES.S_V2`.

No, that's not a mistake. When the `BuildCompat.isAtLeastT()` check [was written](https://github.com/material-components/material-components-android/commit/b3a7dafe6e9a6f5ed7c83ed5b1898b2024957d09), the library used `androidx.core:core:1.7.0` (transitive dependency from `androidx.appcompat:appcompat:1.4.0`), and in it, the code name "T" meant the version that comes after "S" (see [implementation](https://github.com/androidx/androidx/blob/8b6d694f52b1ccf1fc045845a8c80244bc2e3545/core/core/src/main/java/androidx/core/os/BuildCompat.java#L158-L183)).

pubiqq: > Android T is the first version where all devices should support dynamic color by default, so we don't want to return true for S_V2.

- Dynamic color has been available [since Android 12](https://source.android.com/docs/core/display/dynamic-color#tonal-palettes).
- Dynamic color with monet, wallpaper color extraction, and stuff has been available [since Android 12L](https://source.android.com/docs/core/display/material#android-12-patches).

?

dsn5ft: That's an interesting find re: the initial `BuildCompat.isAtLeastT()` implementation, but tbh I don't think that was intentional on our part.

Dynamic Color was initially introduced in Android 12, but OEMs had to opt-in to generate the system_ colors at runtime based on the wallpaper / user choice, otherwise those system_ colors would just remain at their hardcoded defaults. That's why we have the device / manufacturer allow list. I'm fairly certain that T is the first version where that behavior of updating system_ colors was baked into the Android Framework / OS by default, but I'll have to do some more digging to make sure.

dsn5ft: Regardless, I would say the current behavior of the library (with up-to-date dependencies), both internally and externally is that `BuildCompat.isAtLeastT()` means `VERSION.SDK_INT >= VERSION_CODES.T`. So I think we should follow that as opposed to reverting to the potentially initial / accidental behavior from years ago.

pubiqq: The behavior you require is implemented in the [Android 12 patches](https://source.android.com/docs/core/display/material#android-12-patches). They're included in Android 12L and above.

pubiqq:
Bump

pubiqq:
Bump

pubiqq:
Ping

pubiqq:
Ping

pubiqq:
Ping

PR #4235, opened 2024-07-22
[Internal] Fix lint errors

pubiqq:


pubiqq:
I see you've fixed the problem in [this commit](https://github.com/material-components/material-components-android/commit/4abc9c60f16622a25c815166230e6871b4d0bbef), but I still recommend using my PR as it reuses version checking.

PR #4238, opened 2024-07-22
[NavigationView] Fix pointer icon for navigation items

pubiqq:
Fixes https://github.com/material-components/material-components-android/issues/4314

### DIFF
--- a/lib/java/com/google/android/material/navigation/NavigationBarItemView.java
+++ b/lib/java/com/google/android/material/navigation/NavigationBarItemView.java
@@ -45,6 +45,7 @@ import android.util.TypedValue;
 import android.view.Gravity;
 import android.view.LayoutInflater;
 import android.view.MotionEvent;
+import android.view.PointerIcon;
 import android.view.View;
 import android.view.ViewGroup;
 import android.view.accessibility.AccessibilityNodeInfo;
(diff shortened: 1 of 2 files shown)


lib/java/com/google/android/material/navigation/NavigationBarItemView.java
```diff
@@ -619,13 +619,15 @@ public void setEnabled(boolean enabled) {
...
     largeLabel.setEnabled(enabled);
     icon.setEnabled(enabled);
+  }
 
-    if (enabled) {
-      ViewCompat.setPointerIcon(
-          this, PointerIconCompat.getSystemIcon(getContext(), PointerIconCompat.TYPE_HAND));
-    } else {
-      ViewCompat.setPointerIcon(this, null);
+  @RequiresApi(api = VERSION_CODES.N)
```
drchen, line 624: What about API < 24?

Can you explain why the original approach doesn't work?

pubiqq: > Can you explain why the original approach doesn't work?

Original approach overwrites custom pointer icon (if set). Also it doesn't work correctly if itemView is enabled, but non-clickable.

> What about API < 24?

The pointer icon won't change, as it does when using ViewCompat.

dsn5ft:
See https://github.com/material-components/material-components-android/issues/4314#issuecomment-2379840252

PR #4240, opened 2024-07-24
[Slider] Fix stop indicators color

pubiqq:
[Specs](https://m3.material.io/components/sliders/specs#ac9f6c26-b539-41c2-bf9e-c9daa0d116c2)


pekingme (commented):
I think we have to revert this. The active track is primary; the inactive track is secondaryContainer. To avoid any contrast issues, the tick mark drawn on active track need to be onPrimary; the tick mark drawn on inactive track need to be onSecondaryContainer. This is WAI.

pekingme:
Btw, the tokens / specs are not correct. We are working on updating them.

pubiqq:
Well, even though the current tick colors match the roles now, I hope your designers will reconsider them again, because in light theme inactive ticks are almost black, like they lack tinting.




<details><summary>Prev design</summary>
<p>



</p>
</details>

PR #4242, opened 2024-07-25
BaseSlider Tooltip Stroke width

### LINKED ISSUE #4241: [RangeSlider] Tooltip Stroke visible for two values
**Description:** The issue is when i add two values in the range slider it's showing stroke with default color 

**Expected behavior:** Expecting that i should be able to change the tooltip border as per my need.  20 and 70 have white border 
Screenshot_20240725_132211

**Minimal sample app repro:** Please consider attaching a minimal sample app that reproduces the issue. This will help narrow down the conditions required for reproducing the issue, and it will speed up the bug fix process. You may attach a zip file of the sample app or link to a GitHub repo that contains the sample app.

**Android API version:** Android API version here

**Material Library version:** 1.12.0

**Device:** All

officenik:
Added api to set and get tooltip width size

-  `[RangeSlider] Tooltip Stroke width api added `
- [RangeSlider] Tooltip Stroke visible for two values `closes #4241`

### DIFF
--- a/lib/java/com/google/android/material/slider/BaseSlider.java
+++ b/lib/java/com/google/android/material/slider/BaseSlider.java
@@ -329,6 +329,7 @@ abstract class BaseSlider<
   private boolean forceDrawCompatHalo;
   private boolean isLongPress = false;
   private boolean dirtyConfig;
+  private int tooltipStrokeWidth;
 
   @NonNull private ColorStateList haloColor;
   @NonNull private ColorStateList tickColorActive;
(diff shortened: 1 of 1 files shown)


officenik:
CLA is done

Mika35551 (approved):

lib/java/com/google/android/material/slider/BaseSlider.java
```diff
@@ -827,11 +828,17 @@ private void createLabelPool() {
     }
 
     // Add a stroke if there is more than one label for when they overlap.
-    int strokeWidth = labels.size() == 1 ? 0 : 1;
```
drchen, line 830: This change will break the existing behavior. Please preserve the default behavior when there's only one label, no border will be drawn.

PR #4266, opened 2024-08-20
[Internal] Remove pre-Lollipop checks

pubiqq:


drchen (approved):
Great thanks for the effort!

pubiqq:
Just in case, I want to make it clear that this is only part of the work to remove pre-Lollipop code, as a PR with getting rid of it completely would be too big and too hard for the review, so of course there will be other PRs.

drchen:
Yeah, thanks for splitting it. Some extra efforts are needed on our end to get it through as well, so it might take a few days.

Green369-1 (approved):

Issue #4231, opened 2024-07-19
[TextInputLayout] prefix is overlap with text


Mahes2:
**Description:** 
when using prefix, the prefix text is overlap with the text itself.
![Screenshot 2024-07-19 211946](https://github.com/user-attachments/assets/95ae6a45-6de8-4202-b3cb-46fbbb3a36b0)

**Expected behavior:** 
prefix and actual text must not be overlap

**Source code:** 
 ```
    <com.google.android.material.textfield.TextInputLayout
        android:id="@+id/textInputLayout"
        android:layout_width="match_parent"
        android:layout_height="match_parent"
        app:prefixText="@string/prefix_currency"
        style="@style/Widget.MaterialComponents.TextInputLayout.OutlinedBox"
        android:hint="@string/price">

        <com.google.android.material.textfield.TextInputEditText
            android:layout_width="match_parent"
            android:layout_height="wrap_content"
            android:text="100"
            android:id="@+id/txtVariantPriceEdt" />
    </com.google.android.material.textfield.TextInputLayout>
```

**Android API version:** 
30 (Android 11)

**Material Library version:** Material Android Library version you are using here (e.g., 1.1.0-alpha07)
'com.google.android.material:material:1.12.0'
'com.google.android.material:material:1.13.0-alpha04'

**Device:** Device on which the bug was encountered here
Samsung A50

To help us triage faster, please check to make sure you are using the [latest version](https://github.com/material-components/material-components-android/releases) of the library.

We also happily accept [pull requests](https://github.com/material-components/material-components-android/pulls).

Issue #3993, opened 2024-01-18
[TextInputLayout] startIconDrawable overlaps text


IreneKhramova:
**Description:** startIconDrawable in TextInputLayout in RecyclerView.ViewHolder overlaps text if item does not fit on the screen without scroll (ViewHolder is created later, not when screen is just opened). Text is displayed correctly after scroll to top or after minimizing the app.

**Expected behavior:** startIconDrawable does not overlap text

**Source code:** 
```
<?xml version="1.0" encoding="utf-8"?>
<com.google.android.material.textfield.TextInputLayout xmlns:android="http://schemas.android.com/apk/res/android"
    xmlns:app="http://schemas.android.com/apk/res-auto"
    xmlns:tools="http://schemas.android.com/tools"
    android:id="@+id/textInputLayout"
    android:layout_width="match_parent"
    android:layout_height="wrap_content"
    android:layout_marginHorizontal="16dp"
    android:layout_marginTop="16dp"
    app:layout_constraintBottom_toBottomOf="parent"
    app:layout_constraintEnd_toEndOf="parent"
    app:layout_constraintStart_toStartOf="parent"
    app:layout_constraintTop_toTopOf="parent"
    app:startIconDrawable="@drawable/ic_check">

    <com.google.android.material.textfield.TextInputEditText
        android:id="@+id/editText"
        android:layout_width="match_parent"
        android:layout_height="wrap_content"
        tools:text="Test text" />

</com.google.android.material.textfield.TextInputLayout>
```

**Minimal sample app repro:** https://github.com/IreneKhramova/start-icon-issue

**Android API version:** API 29

**Material Library version:** 1.11.0+

**Device:** Honor 20 pro

https://github.com/material-components/material-components-android/assets/22375303/85085fd0-2016-4c01-9cce-2bfaf1d75a18

Jacks0N23:
Hey, google devs, it's a huge bug!

We cannot update to latest versions unless you fix this. It's ruins some of our screens

Issue #3869, opened 2023-11-24
[TextInputLayout] Wrong hint placement


SamYStudiO:
From android preview textInputLayout hint is misplaced under start icon, This may happen as well in runtime for a view frame

![image](https://github.com/material-components/material-components-android/assets/3678076/180dc50c-1d2b-4b1a-be8d-2dfdb6004df2)

**Material Library version:** 1.11.0-rc01

[TextInputLayout.zip](https://github.com/material-components/material-components-android/files/13460057/TextInputLayout.zip)

SamYStudiO:
Here is a video showing isssue at runtime as well

https://github.com/material-components/material-components-android/assets/3678076/55164290-2026-436b-965c-01d8a41e9e6c

Jacks0N23:
same at #3993 and #4231

PR #4287, opened 2024-09-10
[TextInputLayout] Update text paddings in compound layouts uniformly

pubiqq:
Fixes https://github.com/material-components/material-components-android/issues/4231, most likely fixes https://github.com/material-components/material-components-android/issues/3869 and fixes https://github.com/material-components/material-components-android/issues/3993.

### DIFF
--- a/lib/java/com/google/android/material/textfield/TextInputLayout.java
+++ b/lib/java/com/google/android/material/textfield/TextInputLayout.java
@@ -196,7 +196,7 @@ import java.util.LinkedHashSet;
  * developer guidance</a> and <a href="https://material.io/components/text-fields/overview">design
  * guidelines</a>.
  */
-public class TextInputLayout extends LinearLayout implements OnGlobalLayoutListener {
+public class TextInputLayout extends LinearLayout {
 
   private static final int DEF_STYLE_RES = R.style.Widget_Design_TextInputLayout;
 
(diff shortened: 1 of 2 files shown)


lib/java/com/google/android/material/textfield/TextInputLayout.java
```diff
@@ -718,17 +716,6 @@ public TextInputLayout(@NonNull Context context, @Nullable AttributeSet attrs, i
     setHelperText(helperText);
   }
 
-  @Override
-  public void onGlobalLayout() {
```
leticiarossi, line 722: The OnGlobalLayoutListener was originally added in https://github.com/material-components/material-components-android/commit/93360a5a5d2549c14599507db94423906d5a9757 in order to prevent an onMeasure() infinite loop bug, wouldn't removing OnGlobalLayoutListener reintroduce that bug?

pubiqq: I don't know about the bug, but the situation from the explanation shouldn't happen since `endLayout.measure()` for the visible `endLayout` is guaranteed to be called by `super.onMeasure()`, which is before `updateDummyDrawables()`.

pubiqq:
Ping

pubiqq:
Ping

pubiqq:
Ping

PR #4335, opened 2024-10-08
[MaterialAutoCompleteTextView] incorrect filtering is not applied after rotating device

### LINKED ISSUE #1464: [Exposed Dropdown Menu] Filtering incorrectly applied after rotation
**Description:** It seems that in some cases filtering  is incorrectly applied to `AutocompleteTextView` after rotating the  device which causes that all options except for the selected one disappear from Dropdown Menu.

Steps to reproduce:
1. Open Material Catalog app
2. Go to TextField -> Exposed Dropdown Menu Demo
3. Tap on 4th TextField from the  top and select any  value from the  dropdown menu.
4. Rotate the device to landscape and back to portrait
5. Try to select a value from any of the 4 TextFields

The result is that those Dropdowns are now showing only 1 value instead of all of them.

**Expected behavior:** All TextField Dropdown Menus should show all values after device rotations

**Android API version:** Tested on Android 10 and Android 11 Beta

**Material Library version:** Checked on 1.3.0-alpha01 and 1.2.0-beta01

**Device:** Google Pixel 3 and Emulator

manabu-nakamura:
closes #1464
https://github.com/material-components/material-components-android/issues/1464#issuecomment-1829309736
#4506


manabu-nakamura:
https://github.com/material-components/material-components-android/commit/690bbfc917b5a31f18d6285a684e39e6ef55dd6a

manabu-nakamura:
https://github.com/material-components/material-components-android/commit/76dce7347baa9157b50d0333d611bcaa6ba6d856

Issue #4473, opened 2024-12-16
[NavigationView] Relative item shape insets work incorrectly in RTL


pubiqq:
`NavigationView` with `app:itemShapeInsetEnd="96dp"` in RTL:

<table>
  <tr>
    <th>Actual result</th>
    <th>Expected result</th>
   </tr> 
   <tr>
    <td><img src="https://github.com/user-attachments/assets/dcdaeb40-211a-4667-b9a6-98ed3fba32c9" width="270px"></td>
    <td><img src="https://github.com/user-attachments/assets/f7ac1438-8d83-4d18-8d85-b16cdf0ba928" width="270px"></td>
      </tr>
</table>

**Material Library version:** 1.13.0-alpha08

PR #4474, opened 2024-12-16
[NavigationView] Fix behavior of itemShapeInset* attributes

pubiqq:
Fixes https://github.com/material-components/material-components-android/issues/4473


pubiqq:
Ping

PR #4483, opened 2024-12-18
[Internal] Deprecate compat ViewOverlay

pubiqq:


leticiarossi:
ah sorry, I missed this PR and only saw https://github.com/material-components/material-components-android/pull/4736 which is being synced internally currently

pubiqq:
> ah sorry, I missed this PR and only saw https://github.com/material-components/material-components-android/pull/4736 which is being synced internally currently

It's okay, it looks like https://github.com/material-components/material-components-android/pull/4736 breaks backward compatibility, so my PR may still be useful.

pubiqq:
> since that is a restricted API it should be ok to removed it ...

Also, this API is public and I don't think it's ok to remove it without following a deprecation strategy. There are over 600 places in the `material-components` library that use androidx restricted API and I think it would be annoying if someone removed it in the same manner.

PR #4547, opened 2025-01-20
[TabLayoutMediator] Don't try to detach an already detached TabLayout

pubiqq:


leticiarossi (approved):
I don't think the warning is necessary, otherwise LGTM! I can remove the warning internally

pubiqq:
It seems to me that if the `attach()` function throws an error when trying to attach an attached component, then the `detach` function should at least log an attempt to detach a detached component. Anyway, I'll leave it up to you.

PR #4599, opened 2025-02-12
Add marquee (scrolling text) support for MaterialToolbar

Ishan09811:
This PR adds marquee (scrolling text) support for long titles in MaterialToolbar allowing better UX when displaying lengthy text in the app bar.

NOTE: it will not work with collapsingToolbarLayout as it handles text differently


Ishan09811:
@google-cla bot recheck

PR #4615, opened 2025-02-20
[SideSheet] Prevent NullPointerException when viewRef was nullified and only then accessed by runAfterLayout lambda

eugenegff:
Prevent NullPointerException when viewRef was nullified as part of SideSheetBehavior.onDetachedFromLayoutParams() and only then accessed by runAfterLayout lambda. Similar nearby checks outside of lambda are not enough, as all of them are executed immediately, and access in lambda is done later, after layout.


eugenegff:
signed cla

eugenegff:
This issue is still actual. Unfortunately we can not workaround it on our side in derived class, as viewRef is private field.

eugenegff:
@drchen FYI - fix for NPE in SideSheetBehavior.setState()

Issue #4639, opened 2025-03-04
[TopAppBar] Subtitle letters move sideways when fading


pubiqq:
https://github.com/user-attachments/assets/f5bd73eb-fb35-414b-9ab7-0488061dd3dd

**Material Library version:** 1.13.0-alpha11

PR #4640, opened 2025-03-04
[Internal] Make letter spacing dependent on text size in CollapsingTextHelper

pubiqq:
Also fixes https://github.com/material-components/material-components-android/issues/4639.

### DIFF
--- a/lib/java/com/google/android/material/internal/CollapsingTextHelper.java
+++ b/lib/java/com/google/android/material/internal/CollapsingTextHelper.java
@@ -696,17 +696,6 @@ public final class CollapsingTextHelper {
       textPaint.setColor(getCurrentCollapsedTextColor());
     }
 
-    if (collapsedLetterSpacing != expandedLetterSpacing) {
-      textPaint.setLetterSpacing(
-          lerp(
-              expandedLetterSpacing,
-              collapsedLetterSpacing,
-              fraction,
-              AnimationUtils.FAST_OUT_SLOW_IN_INTERPOLATOR));
-    } else {
-      textPaint.setLetterSpacing(collapsedLetterSpacing);
-    }
-
     // Calculates paint parameters for shadow layer.
     currentShadowRadius = lerp(expandedShadowRadius, collapsedShadowRadius, fraction, null);
     currentShadowDx = lerp(expandedShadowDx, collapsedShadowDx, fraction, null);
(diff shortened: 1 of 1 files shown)


drchen:
Hey Hunter, will you able to review this PR?

pubiqq:
Ping.

lib/java/com/google/android/material/internal/CollapsingTextHelper.java
```diff
@@ -712,17 +712,6 @@ private void calculateOffsets(final float fraction) {
...
-      textPaint.setLetterSpacing(
-          lerp(
-              expandedLetterSpacing,
-              collapsedLetterSpacing,
-              fraction,
-              AnimationUtils.FAST_OUT_SLOW_IN_INTERPOLATOR));
-    } else {
-      textPaint.setLetterSpacing(collapsedLetterSpacing);
-    }
-
```
hunterstich: Instead of removing this block, could it be moved into the `else` statement of `if (fadeModeEnabled)`?

```
if (fadeModeEnabled) {
   ...
} else {
   lerp letter spacing
}
```

hunterstich (changes requested):

hunterstich:
Hey @pubiqq!

I checked this PR internally and it is causes some breakages with TextInputLayout's hint. With this change, when the hint text is collapsed on focus, the hint's letter spacing is smaller than it was previously.

I added some comments suggesting an alternative approach. Let me know what you think.

hunterstich:
Hey @pubiqq. I went ahead and made the changes suggested in tis PR's comments internally. The change should be synced out soon and will close this PR.

pubiqq:
> I checked this PR internally and it is causes some breakages with TextInputLayout's hint. With this change, when the hint text is collapsed on focus, the hint's letter spacing is smaller than it was previously.

For intermediate values of `fraction`, the letter spacing is indeed slightly different from the previous implementation because in the PR it depends on the [text scale](https://github.com/material-components/material-components-android/blob/6b05f8a5e9e470fdd17108cb56e256fd4ea042f1/lib/java/com/google/android/material/internal/CollapsingTextHelper.java#L1102-L1105), which is interpolated using `textSizeInterpolator`, whereas the original implementation uses hardcoded [`AnimationUtils.FAST_OUT_SLOW_IN_INTERPOLATOR`](https://github.com/material-components/material-components-android/blob/00e5edac88e7f24086f95b866efccce4632b8630/lib/java/com/google/android/material/internal/CollapsingTextHelper.java#L716-L721). This seems like the more expected behavior, and that was part of the reason for creating this PR ("Make letter spacing dependent on text size").

PR #4643, opened 2025-03-07
Fix deprecation warning.

JonEckenrode:
Fix link in deprecation warning.


hunterstich:
Thanks, Jon

hunterstich:
This link actually looks correct to me as is. Closing but if feel free to re-open if I've missed something.

PR #4669, opened 2025-03-14
[LinearProgressIndicator] Fix stop indicator size when changing track thickness

pubiqq:


paulfthomas:
I can't reproduce the issue, can you provide more details? It may have been fixed already

pubiqq:
https://github.com/user-attachments/assets/fd8d538d-57ec-495f-81e2-63fd1f47aaec

Issue #4630, opened 2025-02-27
[FloatingActionButton] Animation jerk when switching from the intermediate FAB state to the expanded one


pubiqq:
https://github.com/user-attachments/assets/167ba024-61fe-44f7-b0e4-4bad983c3107

**Material Library version:** 1.13.0-alpha11

PR #4596, opened 2025-02-11
[FloatingActionButton] Fix the width calculation in the wrap_content extend strategy

pubiqq:

PR #4695, opened 2025-03-19
[FloatingActionButton] Fix the width calculation in the wrap_content extend strategy

pubiqq:
Fixes https://github.com/material-components/material-components-android/issues/4630.

--

_P.S. Originally posted by @pubiqq in https://github.com/material-components/material-components-android/issues/4596#issuecomment-2688859561_.


leticiarossi:
forgot to review here with a LGTM before syncing internally!

PR #4736, opened 2025-04-10
[Internal] Remove the internal `ViewOverlay` compat layer

MGaetan89:
Since the min SDK version of this library is 21, and `View.getOverlay()`/`ViewGroup.getOverlay()` are available since API 18, it is no longer necessary to use these compatibility classes.

This probably means that [b/144937975](https://issuetracker.google.com/issues/144937975) can be closed.


amir0001b:
Ꭿℳℐℛ

leticiarossi (approved):
Thank you!

PR #4745, opened 2025-04-13
feat(appbar): Allow multiline titles in MaterialToolbar (Fixes #4725)

### LINKED ISSUE #4725: [MaterialToolbar] Multiline title for accessibility purposes
**Missing feature**
The title of the toolbar is truncated with ellipsize when the user increases the font size from the system settings for accessibility reasons.
Truncated texts do not comply with accessibility guidelines.

**Desiderata behavior**
The title of the toolbar should wrap to the next line when it takes up more lines.

demoncoder-crypto:
[x] Identify the component the PR relates to in brackets in the title.
Suggested title: [AppBar] Allow multiline titles in MaterialToolbar

### DIFF
--- a/lib/java/com/google/android/material/appbar/res/values/integers.xml
+++ b/lib/java/com/google/android/material/appbar/res/values/integers.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!--
-  ~ Copyright (C) 2015 The Android Open Source Project
+  ~ Copyright (C) 2024 The Android Open Source Project
   ~
   ~ Licensed under the Apache License, Version 2.0 (the "License");
   ~ you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


hunterstich:
Hi @demoncoder-crypto

Thanks so much for the PR! I'm not sure if this is something we're likely to merge. If I understand the changes correctly, this is updating the default behavior of some of our themes and styles for M3 and MaterialComponents which would likely break existing usages. Additionally, our design guidelines call out [using a collapsing toolbar for multi-line titles and suggests avoiding wrapping title text in a standard Toolbar](https://m3.material.io/components/app-bars/guidelines#42b3110b-2f47-4640-8fd5-4506156e29f8).

Have you considered using a CollapsingToolbarLayout inside an AppBarLayout when titles might be multiline (which we support).  You can see an example of how to set this up in our [TopAppBarCollapsingMultilineDemoFragment](https://github.com/material-components/material-components-android/blob/master/catalog/java/io/material/catalog/topappbar/TopAppBarCollapsingMultilineDemoFragment.java) example. Here's what that looks like:


https://github.com/user-attachments/assets/84c2787a-7b86-4305-b831-a1f9923f39fa


Closing this for now but feel free to re-open if you'd like to continue discussing the changes.

PR #4766, opened 2025-05-05
[TimePicker] Fix isInTapRegion check

pubiqq:


paulfthomas:
Hey @pubiqq, thanks for posting these PRs, can you explain what the change does? Maybe provide a screenshot/video?

pubiqq:
The PR corrects the calculation of the distance between two points:

bruh

paulfthomas:
I understand this, could you describe the change in behavior with a screenshot or video?

pubiqq:
<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr> 
  <tr>
    <td><video src="https://github.com/user-attachments/assets/49071768-cf35-440d-8990-2e0f01cddb57"></td>
    <td><video src="https://github.com/user-attachments/assets/90328a99-ca2c-4ada-8e87-9520b71dc046"></td>
  </tr>
</table>

hunterstich:
Hey @pubiqq 

We we're looking at this and are still not sure of the change's intent. From the video it looks like the hand is snapping back to 5 when it looks like it should be 7. Could you elaborate on what this fixes?

pubiqq:
The hand is snapping back due to the touch slop. It's [implemented](https://github.com/material-components/material-components-android/blob/b85dd1a4fa42d7f0073e6de985b644b9c80dc0c2/lib/java/com/google/android/material/timepicker/ClockHandView.java#L322) in the code, but incorrectly, and this PR fixes it.

hunterstich:
I'm running the catalog at head vs with this change. With the change it seems more difficult to select numbers between 5 and 10. 

Here's a comparison of trying to drag to 6-7-8-9-10-8-6. The video of the catalog that includes the PR snaps back to 5 when I (as a user) don't think it should. Could it be that I release my pointer when I see the minute text field change to the value I'm targeting? I expect that value to stick even if that value isn't technically the closest to where my touch x/y UP event occurred.

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr> 
  <tr>
    <td><video src="https://github.com/user-attachments/assets/0eab97dc-de36-47c1-8be9-b9fcbce6f0ea"></td>
    <td><video src="https://github.com/user-attachments/assets/5a5c64ef-330b-4aad-86c3-4bb99ffeceb4"></td>
  </tr>
</table>

pubiqq:
> With the change it seems more difficult to select numbers between 5 and 10.

Yes, it does, but that's a completely different issue that needs to be addressed separately (including for Compose btw). This PR only fixes the incorrect touch distance check.

dsn5ft:
Based on the Before and After videos from https://github.com/material-components/material-components-android/pull/4766#issuecomment-3058603637, I don't think we can accept this change as-is, since it would be considered a behavioral regression by many app teams and users.

So I think we would need to combine the fix to the `isInTapRegion` math with whatever other fix / refactor is needed to make selecting numbers between 5 and 10 behave as expected.

pubiqq:
Unfortunately, in this case, I don't know what behavior you expect, given that it currently doesn't work quite as I expect and not quite as you describe. 

To achieve the "I see the minute text field has changed, and I expect that when I release the pointer, its value will remain the same" result, it's probably best not to take the touch slop into account at all.

dsn5ft:
Right. My main point here is that the current code, even with the bug, seems to achieve closer to the desired behavior:

"Hours and minutes can be selected by tapping a number or dragging the dial selector track."

https://m3.material.io/components/time-pickers/guidelines#ae285c54-a994-42cc-88f7-5bacea49e71b

So I agree we may need to remove or refactor the way the touch slop is handled altogether.

dsn5ft:
Closing this PR for clarity and we can continue the discussion on another Issue / PR.

Issue #1171, opened 2020-04-01
[TextInputLayout] Show/Hide Password Icons are Mixed Up


zkovar:
**Description:** 

I am using:
```
<com.google.android.material.textfield.TextInputLayout>
<com.google.android.material.textfield.TextInputEditText />
</com.google.android.material.textfield.TextInputLayout>
```
with
`<item name="passwordToggleEnabled">true</item>`

To create the UI for a Password entry with a show/hide password toggle. The behavior matches the specs from the Material Guidelines shown [here](https://material.io/components/text-fields/#outlined-text-field). However, these feel entirely counter-intuitive **AND** Google's own behavior for other apps does not match this.

**Current Behavior**
![component_show_password](https://user-images.githubusercontent.com/2594178/78181656-c4f81180-7432-11ea-9f69-1a160453fd71.png)
![component_hide_password](https://user-images.githubusercontent.com/2594178/78181646-c32e4e00-7432-11ea-9b3f-146b4b95c800.png)

**Expected Behavior (and Google's own behavior)**
![hide_password](https://user-images.githubusercontent.com/2594178/78181658-c590a800-7432-11ea-859d-677275941b1a.jpg)
![show_password](https://user-images.githubusercontent.com/2594178/78181661-c6c1d500-7432-11ea-9cef-f1fb02c1cb8c.jpg)

**Android API version:**
All versions

**Material Library version:** 
1.2.0-alpha05

**Device:** All devices

hunterstich:
Thanks for filing this @zkovar. I'm bringing this up with design to see what they have to say.

leticiarossi:
designed agreed the current behavior should be flipped, so that the icon shows the action (what will happen after the click), instead of how now it shows the current status

zkovar:
This has been fixed by 99c6a28.

sinnavj:
`    <com.google.android.material.textfield.TextInputLayout
        android:id="@+id/layout_password_callback"
        style="@style/Widget.MaterialComponents.TextInputLayout.OutlinedBox"
        android:layout_width="match_parent"
        android:layout_height="wrap_content"
        android:layout_marginStart="@dimen/padding_medium"
        android:layout_marginTop="@dimen/padding_xlarge"
        android:layout_marginEnd="@dimen/padding_medium"
        android:layout_marginBottom="@dimen/padding_medium"
        app:passwordToggleEnabled="true"   // It works in all api versions 1.2.0 to 1.3.0-beta01
[...]

PR #4827, opened 2025-06-22
Update design_password_eye.xml

PowerPizza:
Fixed reversed behavior of this drawable with passwordToggleEnabled attribute of TextInputLayout of material design.
previously happens,
Eye close = password shows as text
Eye open = password shows as dot dot
NOW (fixed),
Eye close = password shows as dot dot
Eye open = password shows as text

### DIFF
--- a/lib/java/com/google/android/material/textfield/res/drawable/design_password_eye.xml
+++ b/lib/java/com/google/android/material/textfield/res/drawable/design_password_eye.xml
@@ -20,13 +20,13 @@
     tools:ignore="NewApi">
 
     <item
-        android:id="@+id/visible"
-        android:drawable="@drawable/design_ic_visibility_off"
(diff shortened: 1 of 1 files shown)


hunterstich:
Hi @PowerPizza. Thanks for the PR.

We've had this issue in the past - https://github.com/material-components/material-components-android/issues/1171 - where we discussed this with our design team and intentionally landed on using:

Eye open = passwords are obscured
Eye with slash = passwords show as clear text

Looking at the TextFieldOutlinedIconsDemoFragment.java in our catalog, I believe the current behavior is correct. My understanding is the design intent is the icon signals the action to invoke on click and not the current state of the text's visibility.

PowerPizza:
Ohh okay... It makes sense,
The icon should represent what will happen if the user clicks it
not what is currently happening.

PR #4884, opened 2025-07-28
[Catalog] Use app:overflowMode="menu"

manabu-nakamura:
```
<com.google.android.material.button.MaterialButtonToggleGroup
    app:selectionRequired="true"
    app:singleSelection="true"/>
```

⬇️
```
<com.google.android.material.button.MaterialButtonGroup
    style="@style/Widget.Material3Expressive.MaterialButtonGroup.Connected"
    app:overflowMode="menu"/>
```

### DIFF
--- a/catalog/java/io/material/catalog/preferences/CatalogPreferencesDialogFragment.java
+++ b/catalog/java/io/material/catalog/preferences/CatalogPreferencesDialogFragment.java
@@ -31,7 +31,7 @@ import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import com.google.android.material.bottomsheet.BottomSheetDialogFragment;
 import com.google.android.material.button.MaterialButton;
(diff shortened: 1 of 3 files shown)


dsn5ft:
Hi! According to our designers / spec, when using toggle buttons it's best to wrap them to the next line as opposed to putting toggle buttons into an overflow. This is because toggle buttons in a group are all very related to each other, and it's important to show them all together on screen.

We're currently considering providing a wrap option for the button group component, and we'll update the catalog to use that if it works out.

manabu-nakamura:
https://github.com/material-components/material-components-android/blob/master/docs/components/ToggleButtonGroup.md:
> Toggle button group (Segmented buttons) ...

https://m3.material.io/components/segmented-buttons/overview:
> Segmented buttons are being deprecated in the Material 3 expressive update.

MaterialButtonToggleGroup isn't deprecated, right?

PR #4885, opened 2025-07-29
[MaterialContainerTransform] Add copyViewImage to decrease cost

ZL114514:
#### Don't forget:

In developing, I found that the MaterialContainerTransform costs more than MaterialSharedAxis, which is has set low-bitmap enough to decrease cost.

Digging deeper, the MaterialSharedAxis is extended from androidx.transition, which has set copyViewToImage to set low-bitmap.

So, I added it into TransitionUtils and MaterialContainerTransform to increase performance.

After 
Screenshot_20250501-145511_Hydrogen

Before 
Screenshot_20250419-192242_Hydrogen

### DIFF
--- a/lib/java/com/google/android/material/transition/MaterialContainerTransform.java
+++ b/lib/java/com/google/android/material/transition/MaterialContainerTransform.java
@@ -22,6 +22,7 @@ import static androidx.annotation.RestrictTo.Scope.LIBRARY_GROUP;
 import static androidx.core.util.Preconditions.checkNotNull;
 import static com.google.android.material.transition.TransitionUtils.calculateArea;
 import static com.google.android.material.transition.TransitionUtils.convertToRelativeCornerSizes;
+import static com.google.android.material.transition.TransitionUtils.copyViewImage;
 import static com.google.android.material.transition.TransitionUtils.createColorShader;
 import static com.google.android.material.transition.TransitionUtils.defaultIfNull;
 import static com.google.android.material.transition.TransitionUtils.findAncestorById;
@@ -935,11 +936,11 @@ public final class MaterialContainerTransform extends Transition {
     final TransitionDrawable transitionDrawable =
         new TransitionDrawable(
             getPathMotion(),
-            startView,
+            copyViewImage(sceneRoot, startView, startView.getParent() != null ? startView.getParent() : startView),
             startBounds,
             startShapeAppearanceModel,
             getElevationOrDefault(startElevation, startView),
-            endView,
+            copyViewImage(sceneRoot, endView, endView.getParent() != null ? endView.getParent() : endView),
             endBounds,
             endShapeAppearanceModel,
             getElevationOrDefault(endElevation, endView),
--- a/lib/java/com/google/android/material/transition/TransitionUtils.java
+++ b/lib/java/com/google/android/material/transition/TransitionUtils.java
@@ -18,7 +18,9 @@ package com.google.android.material.transition;
 
 import android.animation.TimeInterpolator;
 import android.content.Context;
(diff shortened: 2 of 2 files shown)


dsn5ft:
@hunterstich can you take a look? also, if we want to make this change I think we'll need to keep the default behavior the same (live view rendering during the transition) and add a flag / option for this new bitmap version.

lib/java/com/google/android/material/transition/MaterialContainerTransform.java
```diff
@@ -935,11 +936,11 @@ public Animator createAnimator(
     final TransitionDrawable transitionDrawable =
         new TransitionDrawable(
             getPathMotion(),
-            startView,
+            copyViewImage(sceneRoot, startView, startView.getParent() instanceof View ? (View) startView.getParent() : startView),
```
Pittvandewitt, line 939: I think this should become `startView.getParent() instanceof View ? copyViewImage(sceneRoot, startView, (View) startView.getParent()) : startView`

Maybe better to extract this to a function.
```
  private static View maybeGetViewImageCopy(ViewGroup sceneRoot, View view) {
    if (view.getParent() instanceof View) {
      return TransitionUtils.copyViewImage(sceneRoot, view, (View) view.getParent());
    }
    return view;
  }
  ```

File: lib/java/com/google/android/material/transition/TransitionUtils.java [hunk 2]
```diff
@@ -49,12 +54,96 @@ class TransitionUtils {
...
+     * the copy will use a scaled bitmap of the given view.
+     *
+     * @param sceneRoot The ViewGroup in which the view copy will be displayed.
+     * @param view The view to create a copy of.
+     * @param parent The parent of view.
+     */
+    public static View copyViewImage(ViewGroup sceneRoot, View view, View parent) {
+        Matrix matrix = new Matrix();
+        matrix.setTranslate(-parent.getScrollX(), -parent.getScrollY());
+        view.transformMatrixToGlobal(matrix);
```
Pittvandewitt, line 76: Method not available on lower APIs. Should become `ViewCompat.transformMatrixToGlobal(view, matrix);`.

lib/java/com/google/android/material/transition/TransitionUtils.java
```diff
@@ -49,12 +54,96 @@ class TransitionUtils {
...
+        sceneRoot.transformMatrixToLocal(matrix);
```
Pittvandewitt, line 77: Also not available on lower APIs. I suggest adding transformMatrixToLocal to ViewUtils, similar to the Transition package.

lib/java/com/google/android/material/transition/TransitionUtils.java
```diff
@@ -49,12 +54,96 @@ class TransitionUtils {
...
+            bitmapWidth *= scale;
+            bitmapHeight *= scale;
+            matrix.postTranslate(-bounds.left, -bounds.top);
+            matrix.postScale(scale, scale);
+            final Picture picture = new Picture();
+            final Canvas canvas = picture.beginRecording(bitmapWidth, bitmapHeight);
+            canvas.concat(matrix);
+            view.draw(canvas);
+            picture.endRecording();
+            bitmap = Bitmap.createBitmap(picture);
```
Pittvandewitt, line 138: createBitmap is not available on lower APIs. The Transition package contains an example on how to fix this.

ZL114514:
Now I find out why it's so slow.

The WebView does lots of things on the UI thread.

The true solution is to add LAYER_TYPE on the onPause function.

Just decreasing the pixel numbers didn't solve.

PR #4893, opened 2025-08-06
[ButtonGroup] Implement height change animation in vertical MaterialButtonGroup

patzly:
Hi all, let me know if I should open an issue first for a pull request like this: I wrote on Mastodon with Liam Spradlin from the design team who helped designing the button groups behavior, my question was whether a vertical button group also should animate the height of the button neighbors like it animates the width in horizontal orientation:

> [@iamliam](https://androiddev.social/@iamliam@mastodon.design) Hi Liam, one specific question regarding button groups: in the next expressive update of my metronome app I would like to use vertical button groups of MDC Android with large icon buttons. The view library doesn't animate the height of the buttons on tap as it would animate the width in a horizontal button group and I think there aren't any specs for vertical button groups in the docs. Should the height also be animated or are vertical button groups generally not intended?

> [@patzly](https://androiddev.social/@patzly) the height should be animated in both orientations, because the design intent is to affect the surrounding buttons on interaction. If it's not supported, I can pass that along to the team working on the library!

I found no open or closed issue for that so I implemented it on my own, I hope it's all correct! Tests are green and the catalog button group animations are working now in vertical orientation like Liam said.

### DIFF
--- a/lib/java/com/google/android/material/button/MaterialButton.java
+++ b/lib/java/com/google/android/material/button/MaterialButton.java
@@ -246,8 +246,11 @@ public class MaterialButton extends AppCompatButton implements Checkable, Shapea
 
   private int orientation = UNSET;
   private float originalWidth = UNSET;
+  private float originalHeight = UNSET;
   @Px private int originalPaddingStart = UNSET;
+  @Px private int originalPaddingTop = UNSET;
   @Px private int originalPaddingEnd = UNSET;
+  @Px private int originalPaddingBottom = UNSET;
 
   @Nullable private LayoutParams originalLayoutParams;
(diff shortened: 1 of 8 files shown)


patzly:
@pekingme Just wanted to kindly ask if this might make it into the next alpha? :)

patzly:
I have now updated my pull request with the latest changes of this project as there was a merge conflict in MaterialButtonGroup.

patzly:
@pekingme Can you maybe review my changes if you have time for that? :) All I did was copy the existing width-changing logic and adapt it to create an equivalent height-changing logic, if that helps.

PR #4906, opened 2025-08-27
[Gradle] Include `:tests` in `settings.gradle`

MGaetan89:
This commit makes the `:tests` module available in a Gradle build. This fixes false-positives in the version catalog (`dexmaker` not being used for example).


tests/build.gradle
```diff
@@ -13,6 +13,8 @@ dependencies {
 }
 
 android {
+  namespace "com.google.android.material"
```
MGaetan89, line 16: I've used the package name defined here:
https://github.com/material-components/material-components-android/blob/c30c2d8056bc9b12bd118f0fe55744198ac3738c/tests/javatests/com/google/android/material/AndroidManifest.xml#L19

MGaetan89:
@hunterstich any interest in this PR?

hunterstich:
Hey @MGaetan89 

By false-positive, you mean dependencies like `dexmaker` being flagged as unused?

MGaetan89:
> Hey @MGaetan89
> 
> By false-positive, you mean dependencies like `dexmaker` being flagged as unused?

Yes, exactly

hunterstich:
Okay gotcha. I see that that would be a nice fix.

I was making sure everything still builds as expected and was getting some errors running `./gradlew assembleDebug` - maybe related to [non-transitive R classes](https://developer.android.com/build/optimize-your-build#use-non-transitive-r-classes)? Are you seeing anything similar?

PR #4910, opened 2025-08-29
[TimePicker] Make the dial selector discrete

pubiqq:
The PR modifies the behavior of the dial selector so that it better matches the m3 guideline and the expected behavior specified in [this](https://github.com/material-components/material-components-android/pull/4766) thread.

In particular, this PR makes it so that the time [never changes when the selector is released](https://github.com/material-components/material-components-android/pull/4766#issuecomment-3058603637), and also fixes click detection for the dial selector (supercedes https://github.com/material-components/material-components-android/pull/4766).


pubiqq:
Ping

PR #4965, opened 2025-10-29
Android Mat_DEVLOPINPatch 5

LisMarEsc89:
- [ ]


dsn5ft:
Closing this since it looks like a strange combination of previous commits and the only new commit is related to BadgeState but the changes there are also strange.

PR #4995, opened 2026-01-14
[Shape] Fix corner animation interruption during state changes while maintaining performance

### LINKED ISSUE #4990: [Button] Corner morph animation broken in ButtonGroup
**Description:** Hi, version 1.14.0-alpha06 was the last version where the corner morph animation plays as expected when buttons are used in a MaterialButtonGroup. Since 1.14.0-alpha07, the corner morph animation of buttons in a MaterialButtonGroup plays only two frames, which leads to a very abrupt experience. This was very like caused by ea9d25050cc15e5f201f4622c179754013e4c7c2 or 9a8272d9c136787e7395d60374dd7f7559201b6f, I think.
In the screen recording of the latest Catalog app (1.14.0-alpha08) you can first see that the corner morph animation of standalone buttons is very smooth, like expected. Then when I switch to the Buttons Group page and tap on the buttons there (icon buttons are also affected), only two frames with either rounded corners or semi-rounded corners are played:

https://github.com/user-attachments/assets/97ef49d5-9c20-43b4-9ee0-519099032833

**Expected behavior:** The expected behavior is the pre-1.14.0-alpha07 behavior with smooth corner morph animations when buttons are placed inside of a MaterialButtonGroup.

**Android API version:** Android 16, API version 36

**Material Library version:** 1.14.0-alpha08

**Device:** Every emulator and every physical device

Thank you in advance for fixing this bug and for your great work!

jayyaj12:
## Summary
Fix corner morph animation interruption in MaterialButtonGroup while maintaining the performance optimization from ea9d25050.

## Problem
After ea9d25050, corner animations are interrupted during state changes because `onBoundsChange()` always skips animations.

## Solution
Instead of simply reverting ea9d25050, this PR uses a time-based approach:

- Track timestamp when state changes occur
- Allow animation within 500ms after state change (for smooth UX)
- Skip animation outside this window (for performance)

This achieves both smooth state transition animations AND fast layout performance, as suggested by reviewers.

## Implementation
- Added `stateChangeTimestamp` to track state change timing
- Modified `onBoundsChange()` to check time window
- Maintains backward compatibility and all existing optimizations

## Test
- Verified smooth corner animations in MaterialButtonGroup during state changes
- Confirmed no performance regression on screen rotation
- Tested initial layout performance remains optimal

Fixes #4990

### DIFF
--- a/lib/java/com/google/android/material/shape/MaterialShapeDrawable.java
+++ b/lib/java/com/google/android/material/shape/MaterialShapeDrawable.java
@@ -47,6 +47,7 @@ import android.graphics.drawable.Drawable;
 import android.os.Build.VERSION;
 import android.os.Build.VERSION_CODES;
 import android.os.Looper;
+import android.os.SystemClock;
 import android.util.AttributeSet;
 import android.util.Log;
 import androidx.annotation.AttrRes;
@@ -187,6 +188,10 @@ public class MaterialShapeDrawable extends Drawable implements TintAwareDrawable
   // Variables for corner morph.
   // Tracks if this is the first non-empty bounds (for skipping initial animation).
   private boolean boundsIsEmpty = true;
+  // Timestamp when state change occurred, used to allow animation during state transitions.
+  private long stateChangeTimestamp = 0;
+  // Duration (ms) to allow animation after state change, accounting for animation frames.
+  private static final long STATE_CHANGE_ANIMATION_WINDOW_MS = 500;
   private boolean isRoundRectCornerMorph = true;
   @NonNull private ShapeAppearanceModel strokeShapeAppearanceModel;
   @Nullable private SpringForce cornerSpringForce;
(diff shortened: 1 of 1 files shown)


jayyaj12:
Hi @dsn5ft, I'm checking in on this PR. I've already verified the fix through local testing and confirmed it restores the corner animation behavior as expected.

Since there hasn't been any feedback from @pekingme regarding a 'better way' for a while, I suggest moving forward with this approach to address the regression in the current alpha version.

Could you please review this or let me know if we should wait further?

dsn5ft:
The issue is that this PR as is reintroduces the issue fixed by https://github.com/material-components/material-components-android/commit/ea9d25050cc15e5f201f4622c179754013e4c7c2 right?

I believe @pekingme has been thinking about an update to fix both issues.

dsn5ft:
The solution being a flag in MaterialShapeDrawable that helps differentiate corner animations triggered by state change vs bound change.

jayyaj12:
Updated the implementation based on feedback!

Instead of simply reverting to `boundsIsEmpty`, I've implemented a time-based approach to distinguish state changes from layout changes.

**I've also updated the PR description** to reflect this new approach. Please take a look at the updated implementation. Thanks!

@dsn5ft @pekingme

dsn5ft:
thanks for updating the PR! but hmm checking elapsed time against a hardcoded value like 500 ms feels a little hacky. I believe we were thinking of introducing a more specific / deterministic based solution. @pekingme thoughts?

PR #5018, opened 2026-03-12
[Internal][CollapsingTextHelper] Fix letter spacing for collapsed text

pubiqq:
Collapsed text must have a letter spacing value of `collapsedLetterSpacing`, regardless of whether the text is scaled or not.


drchen:
Dan can you help take a look?

dsn5ft:
Seems reasonable to me. Do we know if there's a real reason why it swaps the letter spacing like this currently? Maybe @imhappi knows.

@drchen do you want to try importing and running tests?

PR #5033, opened 2026-03-23
[TimePicker] Fix dialog buttons being clipped at the bottom

### LINKED ISSUE #3584: [MaterialTimePicker] Buttons at the bottom are getting cut off

pubiqq:
Fixes https://github.com/material-components/material-components-android/issues/3584.

This PR is a better alternative to https://github.com/material-components/material-components-android/pull/3588, as it not only fixes the bug mentioned above, but also properly aligns the time selector switch button with the confirmation buttons.

**Screenshots:**

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
   </tr> 
   <tr>
    <td></td>
    <td></td>
    </tr>
   <tr>
    <td></td>
    <td></td>
    </tr>
</table>

### DIFF
--- a/lib/java/com/google/android/material/timepicker/res/layout/material_timepicker_dialog.xml
+++ b/lib/java/com/google/android/material/timepicker/res/layout/material_timepicker_dialog.xml
@@ -68,33 +68,41 @@
     android:id="@+id/material_timepicker_mode_button"
     style="?attr/imageButtonStyle"
     android:layout_width="wrap_content"
-    android:layout_height="wrap_content"
-    android:layout_marginStart="12dp"
-    app:layout_constraintStart_toStartOf="parent"
-    app:layout_constraintBottom_toBottomOf="parent" />
+    android:layout_height="wrap_content" />
+
+  <Space
+    android:id="@+id/material_timepicker_mode_cancel_space"
+    android:layout_width="0dp"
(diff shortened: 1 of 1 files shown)


hunterstich:
Hey @pubiqq. I pulled this in internally and it breaks screenshot tests since the buttons Y positions end up slightly higher than they previously were. From briefly looking at the results, the new positioning looks maybe more correct so I'm going to look into how difficult this would be to land and will get back to you.

